Issue #112, opened 2023-06-04
Failing to pull k0smotron image


iamunnip:
k0smotron controller manager pod is failing to pull image `quay.io/k0sproject/k0smotron:latest`

```
Normal   BackOff    17s               kubelet            Back-off pulling image "quay.io/k0sproject/k0smotron:latest"
  Warning  Failed     17s               kubelet            Error: ImagePullBackOff
  Normal   Pulling    5s (x2 over 21s)  kubelet            Pulling image "quay.io/k0sproject/k0smotron:latest"
  Warning  Failed     2s (x2 over 18s)  kubelet            Failed to pull image "quay.io/k0sproject/k0smotron:latest": rpc error: code = NotFound desc = failed to pull and unpack image "quay.io/k0sproject/k0smotron:latest": failed to resolve reference "quay.io/k0sproject/k0smotron:latest": quay.io/k0sproject/k0smotron:latest: not found
  Warning  Failed     2s (x2 over 18s)  kubelet            Error: ErrImagePull
```

makhov:
Hello! 

Please, use the `install.yaml` from the [docs](https://docs.k0smotron.io/stable/install/) website. It contains proper image versions.

```
kubectl apply -f https://docs.k0smotron.io/stable/install.yaml
```

iamunnip:
Hello,
Thanks for the update.
It's okay now.

```
kubectl -n k0smotron get pods
NAME                                            READY   STATUS    RESTARTS   AGE
k0smotron-controller-manager-7748bd58d8-kpls6   2/2     Running   0          61s
```

PR #117, opened 2023-06-05
Remove install.yaml from repo to avoid confusion

makhov:
Currently we keep it in the repo and uses the `latest` tag, which we don't publish. It creates a confusion like in the issue #112


jnummelin (commented):
LGTM, assuming we still generate install.yaml during release process for docs?

Since we create install.yaml for local testing, should we add it to `.gitignore`?

jnummelin (approved):

PR #119, opened 2023-06-05
Cluster API provider  - initial skeleton for Bootstrap & ControlPlane providers

jnummelin:
Very rough skeleton for control plane provider using k0smotron

This implements following features:
- ControlPlane provider using k0smotron controller
- Bootstrap provider for worker nodes

There's quite a few things to still "fix":

- [ ] Proper status handling, now the bootstrap&controlplane controllers output too many errors as they cannot do _stuff_. They should be able to better rely on status fields of depending things.
- [ ] Version is not properly used in Bootstrap provider
- [ ] Docs

I've currently been testing with Hetzner provider only but other infra providers _should_ work too as they all adhere to Cluster API contracts.


juanluisvaladas (commented):
Just a couple questions, obviously there are a lot of things to address but given the early stage I'm not making a full code review.

Makefile
```diff
@@ -69,13 +73,15 @@ api/k0smotron.io/v1beta1/zz_generated.deepcopy.go: $(CONTROLLER_GEN)
 
 generate: $(generate_targets) ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 
+
+GO_PKGS=$(shell go list ./... | grep -v /inttest/)
```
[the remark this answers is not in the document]
jnummelin: oh, that was removed "temporarily" only, need to add it back.

The problem was that with all the needed changes, including the need to drop back to kube 1.26 deps, rendered the inttest as non-compiling and thus the fmt and vet failed.

Makefile
```diff
@@ -56,6 +56,10 @@ help: ## Display this help.
 ##@ Development
 
 ### manifests
+manifests_targets += config/crd/bases/bootstrap.cluster.x-k8s.io_kzerosconfigs.yaml
```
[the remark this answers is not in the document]
jnummelin: This seemed to be a common convention in cluster API providers. I don't have strong opinion though so I can change this (and all related things). @makhov WDYT?

go.mod
```diff
@@ -6,10 +6,13 @@ require (
 	github.com/k0sproject/k0s v1.27.2-0.20230504131248-94378e521a29
 	github.com/onsi/ginkgo/v2 v2.9.5
 	github.com/onsi/gomega v1.27.7
+	github.com/pkg/errors v0.9.1
```
kke, line 9: github.com/pkg/errors is legacy https://github.com/pkg/errors/issues/245

 `apierrors.IsNotFound(errors.Cause(err))` should work as `apierrors.IsNotFound(err)`.
`errors.Wrap(..)` => `errors.Join(...)`
`errors.Cause(err) == x`  => `errors.Is(err, x)`

internal/controller/controlplane/controlplane_controller.go
```diff
@@ -0,0 +1,259 @@
...
+			if err := c.Client.Get(ctx, types.NamespacedName{Namespace: cluster.Namespace, Name: cluster.Spec.InfrastructureRef.Name}, infraCluster); err != nil {
+				log.Error(err, "Failed to get infrastructure cluster")
+				return false, err
+			}
+			log.Info("Found infrastructure cluster")
+			newEndpoint := map[string]interface{}{
+				"host": host,
+				"port": int64(port),
+			}
+			capiutil.IsExternalManagedControlPlane(infraCluster)
```
makhov, line 176: Looks like it's some leftover

internal/controller/k0smotron.io/k0smotroncluster_configmap.go
```diff
@@ -36,6 +36,12 @@ func init() {
 }
 
 func (r *ClusterReconciler) generateCM(kmc *km.Cluster) (v1.ConfigMap, error) {
+	// If LB type svc, force ports to 6443 and 8132
```
makhov, line 39: Why do we force it? I was under the impression, that we can still use custom ports, can't we?

internal/controller/k0smotron.io/k0smotroncluster_service.go
```diff
@@ -80,51 +94,60 @@ func (r *ClusterReconciler) reconcileServices(ctx context.Context, kmc km.Cluste
...
+		nodes, err := r.ClientSet.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+		if len(nodes.Items) == 0 {
+			return fmt.Errorf("no nodes found")
+		}
+		// Get random node from list
+		node := nodes.Items[rand.Intn(len(nodes.Items))]
+		for _, addr := range node.Status.Addresses {
```
makhov, line 140: Here we rely on the order of the addresses. Is it correct?

config/crd/bases/bootstrap.cluster.x-k8s.io_kzerosworkerconfigs.yaml
```diff
@@ -0,0 +1,68 @@
+---
```
juanluisvaladas, line 1: We need to regenerate install.yaml to include the new CRDs and the changes in `config/rbac/role.yaml`.

I believe `make release` will take care of everything without further changes to the Makefile or `config/default/kustomization.yaml`

makhov (approved):
LGTM in general. Left some minor comments

juanluisvaladas (changes requested):
It's mostly fine, other than the install.yaml which needs to be regenerated the rest are just nits. Also we need github actions to detect this in the future.

config/crd/bases/bootstrap.cluster.x-k8s.io_kzerosworkerconfigs.yaml
jnummelin: In #117 we removed it from the repo to avoid confusion. It is still generated as part of the release and the docs refer to it as:
```
kubectl apply -f https://docs.k0smotron.io/stable/install.yaml
```

docs/cluster-api.md
```diff
@@ -0,0 +1,11 @@
+# k0smotron as Cluster API provider
+
+k0smotron can act as a [Cluster API](https://cluster-api.sigs.k8s.io/) provider for both control planes and for `Machine` bootstrapping.
```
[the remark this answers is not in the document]
jnummelin: Oh yes, very experimental still

go.mod
jnummelin: Right. I took some inspiration for these from the upstream cluster api providers which seem to use these heavily still. And in the core of cluster api there's lot of usage of `pkg/errors`. I wonder if the changes would be fully compatible 🤔

internal/controller/controlplane/controlplane_controller.go
jnummelin: Oh yes 👍

internal/controller/k0smotron.io/k0smotroncluster_configmap.go
jnummelin: The idea was that when LB svc is in front, we can allow k8s to pick random node ports instead of users having to worry about which port is free etc.

internal/controller/k0smotron.io/k0smotroncluster_service.go
jnummelin: Well, sort of. It _should_ prefer `NodeExternalIP` but I see what you mean. In some cases if it sees the internal one on the list first it'll pick it up. I can work it bit more to pick up both types of addresses and return one preferring external. I'll move this to a separate function too.

docs/cluster-api.md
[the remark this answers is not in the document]
jnummelin: done

config/crd/bases/bootstrap.cluster.x-k8s.io_kzerosworkerconfigs.yaml
juanluisvaladas: Ah that's smart.

Makefile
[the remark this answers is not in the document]
jnummelin: I renamed this now to `K0sWorkerConfig` in the last commit

internal/controller/k0smotron.io/k0smotroncluster_configmap.go
jnummelin: Removed the forcing of the ports now.

jnummelin:
@makhov @juanluisvaladas I fixed all the commented bits now. PTAL again

config/samples/capi-hetzner.yaml
```diff
@@ -0,0 +1,75 @@
...
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Machine
+metadata:
+  name: capi-test-0
+spec:
+  clusterName: capi-test
+  bootstrap:
+    configRef: # This triggers our controller to create cloud-init secret
+      apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+      kind: KZerosWorkerConfig
```
juanluisvaladas, line 48: Type is KZeros

api/bootstrap/v1beta1/kzeros_types.go
```diff
@@ -0,0 +1,73 @@
+/*
```
juanluisvaladas, line 1: Probably this file should be renamed to k0s_types.go as well. I imagine renaming the file won't have any actual impact.
﻿

juanluisvaladas (changes requested):
As a side note, I noticed the new API is not defined in the PROJECT file. I imagine this happened because you added this manually without using the kubebuilder CLI. If this is the case we should either remove the PROJECT file or make sure it's consistent. 

The PROJECT file isn't adding a lot of value and I noticed it's already inconsistent in the main branch (became inconssistent after renaming the Cluster type) so I'm in favour of removing it entirely.

Anyway this doesn't have to be done in this PR, I'm requesting changes for the hertzner sample, the rest isn't important enough to block. The other changes look good to me.

api/bootstrap/v1beta1/kzeros_types.go
jnummelin: done

juanluisvaladas (approved):

PR #120, opened 2023-06-06
Allow to configure manifests via configmap

juanluisvaladas:


api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -65,6 +65,12 @@ type ClusterSpec struct {
 	//+kubebuilder:validation:Enum:=kuberouter;calico;custom
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="cniPlugin is immutable"
 	CNIPlugin string `json:"cniPlugin,omitempty"`
+	// Manifests allows to specify a confiMmap with manifests to be
```
jnummelin, line 68: hmm, could it be a secret instead? I mean in some cases there might be some API tokens or similar in these manifests.

internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
```diff
@@ -146,6 +147,19 @@ func (r *ClusterReconciler) generateStatefulSet(kmc *km.Cluster) (apps.StatefulS
...
+		statefulSet.Spec.Template.Spec.Volumes = append(statefulSet.Spec.Template.Spec.Volumes, v1.Volume{
+			Name: "manifests",
+			VolumeSource: v1.VolumeSource{
+				ConfigMap: &v1.ConfigMapVolumeSource{
+					LocalObjectReference: v1.LocalObjectReference{Name: kmc.Spec.Manifests},
+				}}})
+
+		statefulSet.Spec.Template.Spec.Containers[0].VolumeMounts = append(statefulSet.Spec.Template.Spec.Containers[0].VolumeMounts, v1.VolumeMount{
+			Name:      "manifests",
+			MountPath: "/var/lib/k0s/manifests/custom",
```
jnummelin, line 160: What if instead adding a single `custom` "stack" (in k0s implementation terms) we iterate over all the keys and add each key as a separate stack?

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
juanluisvaladas: That makes sense, I'll change that

internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
juanluisvaladas: There is certainly a benefit of having more than one stack for better error handling, I hadn't thought of that...  However, doing that would require the operator to watch for changes and redeploy the stateful set in the event of changes of the configmap.

What do you think of allowing multiple secrets instead of just one? I mean changing the type of `Manifests` from `string` to `[]string`. It would allow multiple stacks, and if there are changes in the `Cluster`object we already monitor for that.

jnummelin: > I mean changing the type of Manifests from string to []string. It would allow multiple stacks, and if there are changes in the Clusterobject we already monitor for that.

That makes sense.

Although I started to wonder if we can achieve this with plain secrets. I know you can mount different values to different files pretty automatically, but can the key contain a path? Probably worth of some trialing. I mean if that would work, then we wouldn't have to even re-create the statefullset pods as k8s will sync the secret mount automatically, IIRC.

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -65,6 +65,12 @@ type ClusterSpec struct {
...
+	// applied to the cluster. The configmap will be mounted as a volime
+	// in /var/lib/k0s/manifests. For more information check
+	// https://docs.k0sproject.io/latest/manifests/
+	//+kubebuilder:validation:Optional
+	Manifests string `json:"manifests,omitempty"`
```
makhov, line 73: Could we use a more verbose name like `ManifestSecrets`?

internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
makhov: > I mean changing the type of Manifests from string to []string

Makes a lot of sense. In theory, different manifests may have different sources.

juanluisvaladas: > Although I started to wonder if we can achieve this with plain secrets. I know you can mount different values to different files pretty automatically, but can the key contain a path?

As per kubernetes documentation no:
The keys of data and stringData must consist of alphanumeric characters, `-`, `_` or `.`.

I don't think doing this in a single secret is feasible without using subPath volume mounts but that would require us to monitor the secrets and I don't think this is something we want...

jnummelin: 👌

config/crd/bases/k0smotron.io_clusters.yaml
```diff
@@ -70,6 +70,11 @@ spec:
                 description: KineDataSourceURL defines the kine datasource URL. Required
                   for HA controlplane setup. Must be set if replicas > 1.
                 type: string
+              manifests:
+                description: Manifests allows to specify a confiMmap with manifests
```
kke, line 74:
```suggestion
                description: Manifests allows to specify a configMap with manifests
```

config/crd/bases/k0smotron.io_clusters.yaml
```diff
@@ -70,6 +70,11 @@ spec:
...
+                  to be applied to the cluster. The configmap will be mounted as a
+                  volime in /var/lib/k0s/manifests. For more information check https://docs.k0sproject.io/latest/manifests/
```
kke, line 76:
```suggestion
                  volume in /var/lib/k0s/manifests. For more information check https://docs.k0sproject.io/latest/manifests/
```

install.yaml
```diff
@@ -82,6 +82,11 @@ spec:
                 description: KineDataSourceURL defines the kine datasource URL. Required
                   for HA controlplane setup. Must be set if replicas > 1.
                 type: string
+              manifests:
+                description: Manifests allows to specify a confiMmap with manifests
```
kke, line 86:
```suggestion
                description: Manifests allows to specify a configMap with manifests
```

install.yaml
```diff
@@ -82,6 +82,11 @@ spec:
...
+                  to be applied to the cluster. The configmap will be mounted as a
+                  volime in /var/lib/k0s/manifests. For more information check https://docs.k0sproject.io/latest/manifests/
```
kke, line 88:
```suggestion
                  volume in /var/lib/k0s/manifests. For more information check https://docs.k0sproject.io/latest/manifests/
```

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
kke, line 68:
```suggestion
	// Manifests allows to specify a configMap with manifests to be
```

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
juanluisvaladas: Instead I think the best would be to use a `[]v1.Volume`. I think this only has benefits:
- The name must be a dns label, which works for the filesystem and we don't need to define a name for the stack, just pick the volume name
- We allow the the user to define either a secret or a configmap (or something else, up to them)
- Pretty much every kubernetes intermediate user is already familiar with it.
- Is part of the stable API so it's not going anywhere any time soon.

I updated the PR with this. What do you think?

jnummelin: Makes sense to me 👍

juanluisvaladas:
I just did a manual rebase to solve a very simple conflict. There are no significant changes.

makhov (approved):

PR #140, opened 2023-06-09
Full k0s config support

### LINKED ISSUE #121: Allow full customization of k0s controller config
We need to allow users to fully customize k0s controller config.

I see two possible ways

**ConfigMap**

We can allow user to specify a configmap as the reference for the config. k0smotron will need to "overwrite" some of the fields such as `spec.api.externalAddress` so we might need to copy the user given CM to something else, or at least do an in-place update of it.

**CRD Fields**

We can copy the relevant fields to k0smotron CRD. This would be best from UX point of view as it will expose all the typed fields in the CRDs.  That does have some drawbacks:
- We need to also copy the defaults
- If we copy those as code dependency we will create an unnecessary dependency on k0s version

makhov:
Fixes #121


api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -58,13 +59,10 @@ type ClusterSpec struct {
 	// Required for HA controlplane setup. Must be set if replicas > 1.
 	//+kubebuilder:validation:Optional
 	KineDataSourceURL string `json:"kineDataSourceURL,omitempty"`
-	// CNIPlugin defines the CNI plugin to be used.
-	// Possible values are KubeRouter and Calico. Uses KubeRouter by default.
-	// Cannot be modified after deploying the cluster.
-	//+kubebuilder:default=kuberouter
-	//+kubebuilder:validation:Enum:=kuberouter;calico;custom
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="cniPlugin is immutable"
-	CNIPlugin string `json:"cniPlugin,omitempty"`
```
makhov, line 67: We don't need it as a separate field anymore since we can do it now via `k0sConfig`

docs/configuration.md
```diff
@@ -52,4 +51,17 @@ spec:
...
 
 
+## K0s configuration
+
+K0smotron allows you to configure k0s via `spec.k0sConfig` field. If empty, will be used default configuration. 
+More info about the k0s config: https://docs.k0sproject.io/stable/configuration/ 
+
+**Note**: some fields will be overwritten by k0smotron. K0smotron will set the following fields:
+- `spec.k0sConfig.spec.api.externalAddress` will be set to the value of `spec.externalAddress` if `spec.externalAddress` is set. 
+   If not, k0smotron will try to detect `externalAddress` out of nodes IP addresses.
```
jnummelin, line 61: nit: Or from LB address

jnummelin (approved):

juanluisvaladas (approved):

docs/configuration.md
makhov: Added info about LB address

PR #150, opened 2023-06-12
Copy labels and annotations from the Cluster to all child resources

### LINKED ISSUE #124: Copy labels from k0smotron object to all other objects
We probably want to copy annotations too

makhov:
Fixes #124


juanluisvaladas (commented):
Shouldn't we also add the  annotations for the secret generated by the JoinTokenRequestReconciler? Other than that, LGTM

makhov:
Added annotations to the JoinTokenRequestReconciler

juanluisvaladas (approved):

PR #155, opened 2023-06-14
[CAPI] cloud init customization

### LINKED ISSUE #126: cloud-init customization
We need to also support some generic cloud-init “additions”
- Create supporting files
- Run command as pre and post k0s setup “hooks” to setup e.g. kernel tunings etc.

jnummelin:
### Add more configurable items in `K0sWorkerConfig`
Files: Extra files create via cloud-init
Pre/Post commands: Executed pre and post k0s install and start
Args: Extra args passed to `k0s worker` command
DownloadURL: If k0s needs to be downloaded from custom URL instead of via get.k0s.sh
PreInstalledK0s: If k0s does not need to be downloaded at all

fixes #126

### DIFF
--- a/api/bootstrap/v1beta1/k0s_types.go
+++ b/api/bootstrap/v1beta1/k0s_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1beta1
 
 import (
+	"github.com/k0sproject/k0smotron/internal/cloudinit"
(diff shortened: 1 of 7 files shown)


jnummelin:
hmm, maybe I should break up the docs part into it's own PR? @makhov @juanluisvaladas WDYT?

PR #160, opened 2023-06-16
Add Docker CAPI example

stevejr:
Add the Docker CAPI example which will allow the users to deploy a k0smotron workplane with the Docker CAPI Provider.


juanluisvaladas:
It looks good so far, but I'm not sure if we should merge it without a worker. WDYT @jnummelin @makhov

jnummelin:
> I'm not sure if we should merge it without a worker.

@juanluisvaladas not sure I get what you mean. the example has a worker node and is in `Provisioned` state.

jnummelin:
@stevejr Thanks for the contribution!

PR #171, opened 2023-06-21
Monitoring

### LINKED ISSUE #125: Monitoring
Enable monitorability on k0smotron controlplanes

makhov:
Exposing control-plane metrics for Prometheus-like scrappers.
The feature adds two sidecars to the k0s pod: Prometheus to scrape metrics from the k0s system components and nginx proxy to make them accessible for the mothership's scraper.

Disabled by default.

Fixes #125

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_controller.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_controller.go
@@ -100,6 +100,13 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{Requeue: true, RequeueAfter: time.Minute}, err
 	}
 
+	if kmc.Spec.EnableMonitoring {
+		if err := r.reconcileMonitoringCM(ctx, kmc); err != nil {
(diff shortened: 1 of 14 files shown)


internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -100,6 +100,11 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{Requeue: true, RequeueAfter: time.Minute}, err
 	}
 
+	if err := r.reconcileMonitoringCM(ctx, kmc); err != nil {
```
jnummelin, line 103: Shouldn't we only create this if monitoring is enabled?

jnummelin (commented):
Some minor questions only

internal/controller/k0smotron.io/k0smotroncluster_controller.go
makhov: Right, thanks! Fixed

jnummelin (approved):

PR #197, opened 2023-07-20
Copy both Cluster and Cluster.spec.service.annotations to created svc

jnummelin:


internal/controller/k0smotron.io/k0smotroncluster_service.go
```diff
@@ -69,6 +69,15 @@ func (r *ClusterReconciler) generateService(kmc *km.Cluster) v1.Service {
 
 	labels := labelsForCluster(kmc)
 
+	// Copu both Cluster level annotations and Service annotations
```
juanluisvaladas, line 72: There is a typo here, Copu instead of Copy

internal/controller/k0smotron.io/k0smotroncluster_service_test.go
```diff
@@ -127,3 +129,67 @@ func TestClusterReconciler_findNodeAddress(t *testing.T) {
...
+				"test": "test",
+			},
+		},
+		{
+			name: "when annotations are set on both Cluster and svc",
+			kmc: &km.Cluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+					Annotations: map[string]string{
+						"test": "test",
```
juanluisvaladas, line 170: I think it would make sense to test if an annotation is defined on both it gets the service annotation instead.
This small change should be enough:
```suggestion
						"test": "test",
						"foo":  "test",
```

I don't feel very strongly about this, so if you think it doesn't make sense then don't add it.

juanluisvaladas (changes requested):

internal/controller/k0smotron.io/k0smotroncluster_service_test.go
jnummelin: TEsting this makes sense. I'll add a separate case for this

PR #207, opened 2023-07-26
Revert "Bump markdown from 3.3.7 to 3.4.4 in /docs"

juanluisvaladas:
Reverts https://github.com/k0sproject/k0smotron/pull/204

mkdocs 1.4.3 requires Markdown <3.4. This is fixed in upstream but until the new release is made we can't merge this.

I'm not quite sure why the github actions didn't catch this before merging but I'll check that up.


juanluisvaladas:
Picked the wrong source branch

PR #213, opened 2023-07-27
Cluster API Docs rework

jlnhnng:
Added learnings from k0smotron with CAPA & aligned examples & naming

### DIFF
--- a/docs/capi-aws.md
+++ b/docs/capi-aws.md
@@ -1,22 +1,20 @@
 # Cluster API - AWS
 
-In this guide, we show you how to use AWS for the worker plane with the k0smotron control plane.
+This example demonstrates how k0smotron can be used with CAPA (Cluster API Provider Amazon Web Services).
 
 ## Prerequisites
 
-See the common [prerequisites](capi-examples.md#prerequisites) for k0smotron.
-
-Also, you need to have an AWS account and your AWS CLI configured with your credentials.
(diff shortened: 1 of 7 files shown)


docs/capi-aws.md
```diff
@@ -67,79 +65,141 @@ metadata:
...
+    apiPort: 6443
+    konnectivityPort: 8132
+    annotations:
+      service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
+```
+
+For scenarios involving Classic ELBs or NLBs without special options, the AWS CCM can be utilized.
+
+#### Internal NLB
+If you aim to use the NLB and set the schema to `internal`, the target group attribute `preserve_client_ip.enabled=false` is required due to "hairpinning" or "NAT loopback". In such cases, the AWS CCM cannot be used because it doesn't support setting Target Group Attributes. Therefore, the AWS Load Balancer Controller, which has the ability to set Target Group Attributes, becomes necessary.
```
jnummelin, line 183: Add a link to the AWS LB controller setup instructions.

jnummelin (changes requested):
Couple minor additions needed

docs/capi-aws.md
jlnhnng: Yes, makes sense.

jnummelin (approved):

PR #230, opened 2023-08-11
K0sControlPlane controller improvements

makhov:
- Better cluster creation
- Better kubeconfig creation
- Better cluster name propagation


jnummelin (approved):

internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -19,40 +19,34 @@ package bootstrap
...
 	"context"
 	"fmt"
-	kutil "github.com/k0sproject/k0smotron/internal/util"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"sigs.k8s.io/cluster-api/controllers/remote"
-	"sigs.k8s.io/cluster-api/util/kubeconfig"
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
```
kke, line 25: pkg/errors is deprecated/archived. Use `fmt.Errorf("foo: %w", err)` instead of `errors.Wrap(err, "foo")`.

kke: Was late

PR #289, opened 2023-10-04
Ensure compatible versions are used between CP and workers

jnummelin:
Currently in some tests the CP gets created with "latest" (e.g. 1.28.2) and workers as 1.27.1, the workers will never get up since they cannot find the proper kubelet configs etc. as k0s assumes same minor version in the initial node bootstrap.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -84,7 +84,7 @@ jobs:
   smoketest:
     name: Smoke test
     needs: build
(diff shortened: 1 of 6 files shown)


juanluisvaladas:
Looks good, but the last PR should probably be a fixup of the first one.

jnummelin:
> the last PR should probably be a fixup of the first one

yup, fixed now

makhov (approved):

PR #291, opened 2023-10-04
Remove redundant k0s version field

makhov:
subj


inttest/capi-controlplane-docker-worker/capi_controlplane_docker_worker_test.go
```diff
@@ -195,7 +195,7 @@ metadata:
 spec:
   replicas: 1
   k0sConfigSpec:
-    version: v1.28.2+k0s.0
+    version:  v1.27.3+k0s.0
```
makhov, line 198: Kind doesn't have v1.28.2 image (only 1.28.0). The closest version which is supported both by kind and k0s is 1.27.3

jnummelin (approved):

PR #302, opened 2023-10-13
Get go version from the go.mod on pipelines

makhov:


.github/workflows/publish-docs.yml
```diff
@@ -6,6 +6,8 @@ on:
   release:
     types:
       - published
+env:
+  GO_VERSION: "1.21.1"
```
kke, line 10: There's:

```yaml
    - name: Set up Go
      uses: actions/setup-go@v4
      with:
        go-version-file: go.mod
```

to get the version from go.mod

makhov: Thanks! I've updated the PR

kke (approved):

PR #319, opened 2023-10-27
RemoteMachine reset

### LINKED ISSUE #310: [CAPI] Should `RemoteMachine` ctrl stop and reset k0s on machine delete?
When `Machine` get deleted which is "implemented" by a `RemoteMachine`, should the remote machine controller do the following:
- SSH into the machine
- stop k0s
- run reset
- remove finalizer

When the `Machine` itself is finally removed, the CAPI machine controller will remove the respective node from the child cluster API. 

@makhov @juanluisvaladas WDYT?

makhov:
Fixes #310

### DIFF
--- a/internal/controller/infrastructure/provisioner.go
+++ b/internal/controller/infrastructure/provisioner.go
@@ -38,6 +38,13 @@ type Provisioner struct {
 	log           logr.Logger
 }
 
+const stopCommandTemplate = `(command -v systemctl > /dev/null 2>&1 && systemctl stop %s) || (command -v rc-service > /dev/null 2>&1 && rc-service %s stop) || (echo "Not a supported init system"; false)`
+
+const (
+	ctrlService   = "k0scontroller"
+	workerService = "k0sworker"
+)
+
 // Provision provisions a new machine
 // The provisioning process is as follows:
 // 1. Open SSH connection to the machine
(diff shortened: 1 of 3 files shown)


internal/controller/infrastructure/provisioner.go
```diff
@@ -98,6 +104,56 @@ func (p *Provisioner) Provision(_ context.Context) error {
...
+// 1. Open SSH connection to the machine
+// 2. Stops k0s
+// 3. Removes node from etcd
+// 4. Runs k0s reset
+func (p *Provisioner) Cleanup(_ context.Context, mode RemoteMachineMode) error {
+	if mode == ModeNonK0s {
+		return nil
+	}
+
+	authM, err := rig.ParseSSHPrivateKey(p.sshKey, rig.DefaultPasswordCallback)
```
kke, line 118: I don't know if it's a problem here but the password callback will prompt for input if the key is passphrase protected. If you want to avoid having it waiting for input forever, just pass nil.

makhov: Thanks!

internal/controller/infrastructure/provisioner.go
```diff
@@ -38,6 +38,13 @@ type Provisioner struct {
 	log           logr.Logger
 }
 
+const startCommandTemplate = `(command -v systemctl > /dev/null 2>&1 && systemctl stop %s) || (command -v rc-service > /dev/null 2>&1 && rc-service %s stop) || (echo "Not a supported init system"; false)`
```
jnummelin, line 41: Why do we need this, can't we rely on `k0s start` subcommand to handle this?

makhov: It's basically a copy of [the command that we use to start k0s](https://github.com/k0sproject/k0smotron/blob/main/internal/controller/bootstrap/bootstrap_controller.go#L228). And of course, the name of the constant should be `stopCommandTemplate`. I've fixed that.

jnummelin (approved):

PR #325, opened 2023-11-02
ClusterClass support

makhov:
`ClusterClass` support for k0smotron. 

The PR adds:
- K0sControlPlaneTemplate object
- K0smotronControlPlaneTemplate object
- `spec.version` field to the `K0sControlPlane` and `K0smotronControlPlane` objects object, since the CAPI controller manager fills it on creating
- Integration test

⚠️ Breaking changes:
- `spec.k0sVersion` field was removed to avoid having two version fields both for `K0sControlPlane` and `K0smotronControlPlane` objects
- `spec.k0sImage` field was renamed to `spec.image` for `K0smotronControlPlane` object


jnummelin:
Should we also add some docs around how to use `ClusterClass` and related stuff?

config/crd/bases/controlplane.cluster.x-k8s.io_k0scontrolplanetemplates.yaml
```diff
@@ -0,0 +1,225 @@
...
+  scope: Namespaced
+  versions:
+  - name: v1beta1
+    schema:
+      openAPIV3Schema:
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```
kke, line 26: I don't know if it matters, but this syntax will cause the string to become:

```console
APIVersion defines the versioned schema of this representation
              of an object. Servers should convert recognized schemas to the latest
              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
```

The `|` syntax would preserve the linefeeds, but would remove the extra indentation:

```suggestion
            description: | 
              APIVersion defines the versioned schema of this representation
              of an object. Servers should convert recognized schemas to the latest
              internal value, and may reject unrecognized values. More info: 
              https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
```

And the result becomes:

```console
APIVersion defines the versioned schema of this representation
of an object. Servers should convert recognized schemas to the latest
internal value, and may reject unrecognized values. More info: 
https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
```

The `>` syntax will remove all the linefeeds and indentation and you end up with just a long one-liner:

```suggestion
            description: > 
              APIVersion defines the versioned schema of this representation
              of an object. Servers should convert recognized schemas to the latest
              internal value, and may reject unrecognized values. More info:
              https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
```

Results in:

```console
APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
```

config/crd/bases/controlplane.cluster.x-k8s.io_k0scontrolplanetemplates.yaml
```diff
@@ -0,0 +1,225 @@
...
+                                  such as desiredState.manifest.containers[2]. For
+                                  example, if the object reference is to a container
+                                  within a pod, this would take on a value like: "spec.containers{name}"
+                                  (where "name" refers to the name of the container
+                                  that triggered the event) or if no container name
+                                  is specified "spec.containers[2]" (container with
+                                  index 2 in this pod). This syntax is chosen only
+                                  to have some well-defined way of referencing a part
+                                  of an object. TODO: this design is not final and
+                                  this field is subject to change in the future.'
```
kke, line 170: This string becomes:

```console
If referring to a piece of an object
                                  instead of an entire object, this string should
                                  contain a valid JSON/Go field access statement,
                                  such as desiredState.manifest.containers[2]. For
                                  example, if the object reference is to a container
                                  within a pod, this would take on a value like: "spec.containers{name}"
                                  (where "name" refers to the name of the container
                                  that triggered the event) or if no container name
                                  is specified "spec.containers[2]" (container with
                                  index 2 in this pod). This syntax is chosen only
                                  to have some well-defined way of referencing a part
                                  of an object. TODO: this design is not final and
                                  this field is subject to change in the future.
```

config/crd/bases/controlplane.cluster.x-k8s.io_k0scontrolplanetemplates.yaml
makhov: It's an autogenerated code, so it will be changed back to the current state if we re-generate it again.

api/controlplane/v1beta1/k0s_types.go
```diff
@@ -45,7 +45,8 @@ type K0sControlPlaneSpec struct {
 	MachineTemplate *K0sControlPlaneMachineTemplate `json:"machineTemplate"`
 	//+kubebuilder:validation:Optional
 	//+kubebuilder:default=1
-	Replicas int32 `json:"replicas,omitempty"`
+	Replicas int32  `json:"replicas,omitempty"`
+	Version  string `json:"version,omitempty"`
```
makhov, line 49: That feels weird to have both `version` and `k0sVersion` on the same object. Can we just remove `k0sVersion`?
Basically, we can support the k0s and k8s versions here, adding `+k0s.0` if it's missing. @jnummelin wdyt?

makhov: I removed `K0sVersion` field

api/controlplane/v1beta1/k0s_types.go
```diff
@@ -47,9 +47,10 @@ type K0sControlPlaneSpec struct {
 	//+kubebuilder:validation:Optional
 	//+kubebuilder:default=1
 	Replicas int32 `json:"replicas,omitempty"`
-	// K0sVersion defines the k0s version to be deployed. If empty k0smotron
-	// will pick it automatically.
-	K0sVersion string `json:"k0sVersion"`
+	// Version defines the k0s version to be deployed. Can be used direct k0s version (e.g. v1.27.1+k0s.0) or
+	// just the Kubernetes version (e.g. v1.27.1). If empty k0smotron will pick it automatically.
```
kke, line 51:
```suggestion
	// Version defines the k0s version to be deployed. You can use a specific k0s version (e.g. v1.27.1+k0s.0) or
	// just the Kubernetes version (e.g. v1.27.1). If left empty, k0smotron will select one automatically.
```

config/clusterapi/controlplane/bases/controlplane.cluster.x-k8s.io_k0scontrolplanetemplates.yaml
```diff
@@ -0,0 +1,225 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.11.4
```
kke, line 6: Btw, isn't this a tad bit old? Maybe it's on purpose.

config/samples/capi/capi-controlplane-hetzner.yaml
```diff
@@ -24,7 +24,7 @@ kind: K0smotronControlPlane # This would somehow map to k0smotron
 metadata:
   name: cp-test
 spec:
-  k0sVersion: v1.27.2-k0s.0
+  version: v1.27.2-k0s.0
```
kke, line 27: Should this be `v1.27.2+k0s.0`?

kke: (it's also like this in the couple of examples below this)

docs/capi-clusterclass.md
```diff
@@ -0,0 +1,216 @@
+# ClusterClass
+
+K0smotron supports ClusterClass, a simple way to create many Cluster of a similar shape.
```
kke, line 3:
```suggestion
K0smotron supports ClusterClass, a simple way to create many clusters of a similar shape. 
```

docs/capi-aws.md
```diff
@@ -203,4 +203,4 @@ To access the child cluster we can get the kubeconfig for it with `clusterctl ge
 
 ## Deleting the cluster
 
-For cluster deletion, do **NOT** use `kubectl delete -f my-aws-cluster.yaml` as that will result into orphan AWS resources. Instead, delete the top level `Cluster` object. This approach ensures the proper sequence in deleting all child resources, effectively avoid orphan resources.
\ No newline at end of file
+For cluster deletion, do **NOT** use `kubectl delete -f my-aws-cluster.yaml` as that will result into orphan AWS resources. Instead, delete the top level `Cluster` object. This approach ensures the proper sequence in deleting all child resources, effectively avoid orphan resources.
```
kke, line 206:
```suggestion
For cluster deletion, do **NOT** use `kubectl delete -f my-aws-cluster.yaml` as that will result in orphan AWS resources. Instead, delete the top level `Cluster` object. This approach ensures the proper sequence in deleting all child resources, effectively avoiding orphan resources.
```

config/samples/capi/capi-controlplane-hetzner.yaml
makhov: No, here it is the docker image tag, and we use `-` there.

config/clusterapi/controlplane/bases/controlplane.cluster.x-k8s.io_k0scontrolplanetemplates.yaml
makhov: Yeah, maybe. But it's the version, that we use: https://github.com/k0sproject/k0smotron/blob/main/Makefile.variables#L2

I think we can update it in a separate PR

jnummelin (approved):
LGTM

PR #339, opened 2023-11-13
Using k8s versions everywhere

### LINKED ISSUE #333: Unify and simplify version field
Now we have a kind of mess of different types of version fields and version values. The current plan is:
- Unify version fields and start using plain Kubernetes versions (without `+k0s.0`)
- Introduce optional annotations (default value: `k0s.0`) for the CAPI and K0smotron CRDs, which k0smotron will use to generate k0s version.
- The steps above allow us to more actively use version fields of the CAPI objects, probably we can remove some version fields from our CRDs

makhov:
Currently, we are using different ways to set up a version, which is confusing. In the PR we are switching to using a Kubernetes version (eg, v1.27.2) instead of k0s version (eg, v1.27.2+k0s.0)

Fixes #333

### DIFF
--- a/api/bootstrap/v1beta1/k0s_types.go
+++ b/api/bootstrap/v1beta1/k0s_types.go
@@ -25,6 +25,10 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+const (
+	K0sSuffixAnnotation = "k0smotron.io/k0s-suffix"
+)
+
 func init() {
 	SchemeBuilder.Register(&K0sWorkerConfig{}, &K0sWorkerConfigList{})
 	SchemeBuilder.Register(&K0sControllerConfig{}, &K0sControllerConfigList{})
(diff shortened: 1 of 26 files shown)


juanluisvaladas:
I'm not quite sure I want this...

I think it's a bit more clear but also makes it harder to change the k0s suffix in case you need to set it to `+k0s.1` (I really don't like using an annotation for this). Also it's not very intuitive if someone wants to use a custom image...

I would prefer not to merge this... @jnummelin what do you think about this?

jnummelin:
The challenge here is that some controllers rely that the Machine (and some other "abstract" level of objects) have Kubernetes version, i.e. not with any k0s specific suffix.

So what we discussed with @makhov is the following approach:
- k0s controllers should check the Machine etc. objects for a version field
- if the "upper" level object has a version, use it and add the default `k0s.0` suffix
- if the upper level has no version specified, use "latest" k0s; pretty much like we do currently

I think this gives us the "best of both worlds"; i.e. we can use the Machine etc. version fields but user still has a way to override the k0s specific suffix using k0s objects.

makhov:
We decided to go with #400, so I'm closing this one

PR #400, opened 2024-01-05
Get version from the Machine object

makhov:
Getting k0smotron version from the Machine object as we discussed [here](https://github.com/k0sproject/k0smotron/pull/339#issuecomment-1876831716).


jnummelin (approved):
LGTM

PR #433, opened 2024-01-29
Add yaml inline separator for proper installation of the K0smotronControlPlane

infinitydon:
Hi,

PR is to fix the yaml format of _Cluster API - Remote machine_ provider first example manifest.


makhov (approved):

makhov:
Thanks!

PR #444, opened 2024-02-06
Initial CAPI vSphere documentation

nekwar:
Documentation on how to use k0smotron in conjunction with CAPI for vSphere.

### DIFF
--- a/docs/capi-vsphere.md
+++ b/docs/capi-vsphere.md
@@ -1,0 +1,211 @@
+# Cluster API - VMware
+
+This example demonstrates how k0smotron can be used with CAPV (Cluster API Provider vSphere).
+
+**Table of Contents**
+
+- [Cluster API - VMware](#cluster-api---vmware)
+  * [Setting the scene](#setting-the-scene)
+  * [Preparations](#preparations)
+    + [Configure clusterctl on the local machine](#configure-clusterctl-on-the-local-machine)
+    + [Deploy Cluster API in the management cluster](#deploy-cluster-api-in-the-management-cluster)
+    + [(Optional) IPAM IP pool creation](#optional-ipam-ip-pool-creation)
+  * [(Optional) MetalLB as Load Balancer solution in the management cluster](#optional-metallb-as-load-balancer-solution-in-the-management-cluster)
+  * [Operating child clusters](#operating-child-clusters)
+    + [Generate a child cluster definition using the template](#generate-a-child-cluster-definition-using-the-template)
+      - [Control plane in Pods](#control-plane-in-pods)
+      - [Control plane in VMs](#control-plane-in-vms)
+    + [Deploy the child clusters](#deploy-the-child-clusters)
+    + [Observe the child cluster objects](#observe-the-child-cluster-objects)
+    + [Deleting the child cluster](#deleting-the-child-cluster)
+
+## Setting the scene
+
+To use Cluster API we will bring three environments:
+1. Local machine -- the machine from where some of the commands will be executed
+2. Management cluster -- Kubernetes cluster that will be used to control and manage child clusters
+3. Child cluster -- k0s cluster that will be spinned up with the help of k0smotron
+
+## Preparations
+
+Before starting this example, ensure that you have met the [general prerequisites](capi-examples.md#prerequisites).
+
+### Configure clusterctl on the local machine
+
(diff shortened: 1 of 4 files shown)


nekwar:
I know that +2k lines is quite big for one commit, but the thing is that two yaml templates that I've added are very similar, but they have some important differences

juanluisvaladas:
> I know that +2k lines is quite big for one PR, but the thing is that two yaml templates that I've added are very similar, but they have some important differences

That's actually very reasonable for for this kind of PR...

juanluisvaladas:
@dvirassamy you have some experience with CAPV, would you mind having a look?

docs/capi-vsphere.md
```diff
@@ -0,0 +1,178 @@
+# Cluster API - VMware
+
+This example demonstrates how k0smotron can be used with CAPV (Cluster API Provider vSphere).
+
+## Preparations
+
+Before starting this example, ensure that you have met the [general prerequisites](capi-examples.md#prerequisites).
+
+### Prepare the development/deployment server for ClusterAPI
```
jlnhnng, line 9: @nekwar What is the development/deployment server for CAPI?
The docs do look different to the once we have for e.g. CAPO or CAPA. Can you have a look into them and try to stay in the same schema?

docs/capi-vsphere.md
```diff
@@ -0,0 +1,178 @@
...
+  start: 192.168.117.152
+  end: 192.168.117.180
+```
+
+2. Deploy the IPAM configuration: 
+```
+kubectl apply -f capi-ipam.yaml
+```
+
+## MetalLB as Load Balancer solution
```
jlnhnng, line 84: This only applies to management clusters that are running in vSphere. Please add this as _optional_ for the ones with mgmt cluster on vSphere.

docs/capi-vsphere.md
```diff
@@ -0,0 +1,178 @@
...
+### Generate a child cluster definition using the template:
+
+#### For control plane in Pods:
+
+If you want to have child cluster control plane in a Pods, use [this template](capi-vsphere-tmpl-cp-in-pods.yaml): 
+
+```sh
+clusterctl generate cluster my-cluster --control-plane-machine-count 1 --worker-machine-count 1 --from k0scluster-tmpl-cp-in-pods.yaml > my-cluster.yaml
+```
+
```
jlnhnng, line 129: Add _or_ to clearly show up that these are two options.

docs/capi-vsphere.md
```diff
@@ -0,0 +1,178 @@
...
+If you want to have child cluster control plane in separate VMs, use [that template](capi-vsphere-tmpl-cp-in-pods.yaml) (CP will consist of 3 controllers):
+
+```sh
+clusterctl generate cluster my-cluster --worker-machine-count 1 --from k0scluster-tmpl-cp-in-vms.yaml > my-cluster.yaml 
+```
+
+### Deploy the child clusters
+
+1. Deploy the child cluster: `kubectl apply -f my-cluster.yaml`. Check the deployment status in the Mothership.
```
jlnhnng, line 140: I would suggest to say: "You can check the deployment status in the management cluster."
And not "Check the deployment status in the Mothership."

jlnhnng (changes requested):
I have added comments for suggested changes. I think we should clearly define where the management cluster is running in the docs. This is not clear from my point of view and can lead to confusion. One does not need to run the management cluster on in the infrastructure provider that is documented here.

docs/capi-vsphere.md
nekwar: CAPO and CAPA docs does not say anything about this "development/deployment" server from where `clusterctl` and `kubectl` commands should be executed. IMO, CAPO and CAPA docs should have this info, rather than removing it from CAPV docs

But I agree that the naming ("development/deployment") is a bit confusing, I will change it to "Environment configuration for clusterctl"

Additionally, CAPV docs could not look similar to CAPO and CAPA docs, because there are a lot of differences between platforms.

nekwar:
@jlnhnng then we need to define it in all the other docs (CAPO and CAPA) as well

dvirassamy:
Reviewed the reworked version. Good for me.

jlnhnng (approved):
lgtm

PR #449, opened 2024-02-08
Fix Machine owner reference created by K0sControlPlane

nekwar:
That change sets `Machine`s owner reference, created by a `K0sControlPlane`, to the `K0sControlPlane` (before owner was `Cluster` resource, which breaks the logic)

Additionally, it was not working good with CAPV -- deletion of `Machine`s with `Cluster` set as owner was not successful


makhov (approved):

makhov:
Thanks!

PR #453, opened 2024-02-14
K0sControlPlane upgrade

makhov:


makhov:
Well, the idea with annotation was good until I realized, that we can't map Machine names to ControlNode names.

Here is the rough implementation, which I reverted:
https://github.com/k0sproject/k0smotron/pull/453/commits/ea1030a3204b53e8d22330d4b1f4467765f74976

PR #494, opened 2024-03-12
[DOCS] Compatibility matrix

MagdaDziadosz:


jnummelin (commented):
We do not do any actual compatibility testing so I'm not sure we can state actually much on the topic.

What is relevant is that k0smotron implements specific version of Cluster API [contract](https://cluster-api.sigs.k8s.io/developer/providers/contracts#provider-contract). Currently we implement the `v1beta1` contract for all providers.

PR #495, opened 2024-03-12
[DOCS] Edits to FAQ and Troubleshooting in response to comments.

KoryKessel-Mirantis:


docs/troubleshooting.md
```diff
@@ -1,29 +1,25 @@
 
 # Troubleshooting
 
-## I cannot join a worker node
+## Worker nodes cannot be joined
 
-In [standalone](usage-overview.md#standalone) mode, check whether the join
+If you are using k0smotron in standalone mode, check whether the join
```
jnummelin, line 6: We need to provide some step how to inspect the token.

1. Decode the token:
```
echo "<token>" | base64 -d | gunzip
```
2. Take note of the field `users.user.token`. I looks something like `gb823t.b8ftcytc4ktmvkjz`
3. Using the KUBECONFIG of the child cluster:
```
kubectl -n kube-system get secret bootstrap-token-gb823t --template='{{.data.expiration}}' | base64 -d
2024-03-14T11:08:13Z
```

docs/troubleshooting.md
```diff
@@ -1,29 +1,25 @@
...
-provider controller.
+If you are using k0smotron as a Cluster API provider, check the logs of your
+infrastructure provider controller.
 
-Also, check to see whether different versions were used in the initial cluster
+Furthermore, check whether different versions were used in the initial cluster
 creation for the control plane and the worker nodes, as Kubernetes is
 compatible with a limited number of different versions. For more information,
 refer to the Kubernetes [Version Skew
 Policy](https://kubernetes.io/releases/version-skew-policy/).
```
jnummelin, line 17: This is actually not 100% correct. Kubernetes version skew generically does support running e.g. controller on 1.28.x and worker on 1.27.x. But in case of k0s, and probably most other distros, the initial minor versions need to match at cluster bootstrap time. This is due to the fact that k0s 1.28 controller will NOT setup needed configurationa for older minor versions. In case the controller is started initially with 1.27.x it will of course have those. Then when upgrading, the new version will also of course setup all needed configs for 1.28 too.

jnummelin (changes requested):

docs/troubleshooting.md
KoryKessel-Mirantis: @jnummelin, for Step 2, what will the user do with the `users.user.token` that they have taken note of? For now -- or, at least, for me -- this step just kind of hangs there.

docs/troubleshooting.md
KoryKessel-Mirantis: While I am sure this is essential and fascinating, @jnummelin, it does not actually provide me with any direction (important to remember that I am a technical writer who knows nothing about anything). Should I remove the paragraph entirely, or do we need to rework it? And if we do need to rework it, please provide the rough text or key edits needed to get the pararaph in the correct shape.

docs/troubleshooting.md
jnummelin: They need to take the first part, before the `.`, and use it as part of the secret name. In the above example it's the `gb823t `. So that maps the token to a secret in which the value is the actual full token information.

docs/troubleshooting.md
jnummelin: Maybe something like this:

Furthermore, check whether different Kubernetes minor versions were used in the initial cluster
creation for the control plane and the worker nodes, as k0s mandates that both controllers and workers are using the same minor version during the initial cluster creation.

jnummelin (approved):

ogusarenko (approved):

PR #499, opened 2024-03-13
[DOCS] Move common Monitoring section to main TOC

ologvinova:
Reason: the section also applies to the Cluster API in-cluster use case.


ologvinova:
Abandoning this commit for the sake of https://github.com/k0sproject/k0smotron/pull/500

PR #507, opened 2024-03-18
Check for deletion timestamp on controller

joshperry:
This gets set not only with finalizers, but also when a user does a foreground delete operation (which is default for some UX like argo).

This keeps the controller from desparately attempting to recreate its resources during foreground deletes.


makhov:
Hi! Thanks for the contribution! 
Could you please sign off the commit to make the DCO check path? Here are the [instructions](https://github.com/k0sproject/k0smotron/pull/507/checks?check_run_id=22805656235), on how to do it.

joshperry:
No problem! Sorry, didn't see the DCO failure.

makhov (approved):

PR #570, opened 2024-05-13
Skipping etcd certs checks when Kine is used

makhov:
The regression in the v0.9.5 is fixed.

We don't check for etcd certs if `KineDataSourceURL` is not an empty string. But currently it doesn't work with `KineDataSourceSecretName`. We do the following to make it work, but by mistake, we pass `kmc` not by reference and this doesn't change the KineDataSourceURL and the check works wrongly
```
if kmc.Spec.KineDataSourceSecretName != "" {
	kmc.Spec.KineDataSourceURL = kineDataSourceURLPlaceholder
}
```


jnummelin:
I don't get how this fixes the issue, maybe bit more explanation on the commit message might help

makhov:
Added a comment

jnummelin (approved):

PR #571, opened 2024-05-14
K0sControlPlane update strategies

makhov:


Schnitzel:
I think this is superseeded by https://github.com/k0sproject/k0smotron/pull/667 ?

PR #574, opened 2024-05-15
fix: add missing fsgroup option otherwise etcd couldn't read the dataDir

cwrau:


jnummelin:
@cwrau The linter is not happy, could you have a look?

Schnitzel:
btw, I'm running into exact same issue, just for me it works with `fsGroup: 0` but `fsGroup: 1001` should also work.

I'm confused though why tests previously passed without this, as I can't get it running unless this is set.

makhov (approved):

makhov:
Thanks for the contribution!

makhov:
@Schnitzel We use [rancher/local-path-provisioner](https://github.com/rancher/local-path-provisioner) in our tests and it runs ` mkdir -m 0777 -p "$VOL_DIR"` on volume provisioning. 
I'm thinking about switching to [openebs/dynamic-localpv-provisioner](https://github.com/openebs/dynamic-localpv-provisioner) for our tests

cwrau:
@makhov sorry, I was on holiday, thanks for taking care of this 😅

Great that this got addressed so quickly!

PR #578, opened 2024-05-17
updateStrategy field 

makhov:
The field is added to K0sControlPlane object, but changes nothing atm. 
We will add other update strategies later.

### DIFF
--- a/api/controlplane/v1beta1/k0s_types.go
+++ b/api/controlplane/v1beta1/k0s_types.go
@@ -28,6 +28,12 @@ func init() {
 	SchemeBuilder.Register(&K0sControlPlane{}, &K0sControlPlaneList{})
 }
 
+type UpdateStrategy string
+
+const (
+	UpdateInPlace UpdateStrategy = "InPlace"
+)
+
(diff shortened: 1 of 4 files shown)


api/controlplane/v1beta1/k0s_types.go
```diff
@@ -28,6 +28,13 @@ func init() {
 	SchemeBuilder.Register(&K0sControlPlane{}, &K0sControlPlaneList{})
 }
 
+type UpdateStrategy string
+
+const (
+	UpdateStrategyRolling  UpdateStrategy = "Rolling"
+	UpdateStrategyRecreate UpdateStrategy = "Recreate"
```
jnummelin, line 35: IMO neither of these describe very well the autopilot update way. I mean it's more like `InPlace`. And if we only support one strategy, IMO we should only define one strategy as option

makhov: Changed it to `InPlace` and remove the second one.

jnummelin (approved):

PR #602, opened 2024-05-31
Use server-side apply for installation 

makhov:
Due to the size of CRDs, client-side apply doesn't work with "metadata.annotations: Too long" error. It tries to add it to the `last-applied-configuration` annotation, but it's too long

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -7,7 +7,7 @@ control plane.
 !!! note "TL;DR"
 
     ```bash
-    kubectl apply -f https://docs.k0smotron.io/{{{ extra.k0smotron_version }}}/install.yaml
+    kubectl apply --server-side=true -f https://docs.k0smotron.io/{{{ extra.k0smotron_version }}}/install.yaml
     ```
 
 ## Known limitations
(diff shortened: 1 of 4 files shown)


docs/install.md
```diff
@@ -7,7 +7,7 @@ control plane.
 !!! note "TL;DR"
 
     ```bash
-    kubectl apply -f https://docs.k0smotron.io/{{{ extra.k0smotron_version }}}/install.yaml
+    kubectl create -f https://docs.k0smotron.io/{{{ extra.k0smotron_version }}}/install.yaml
```
juanluisvaladas, line 10: This is fine for the makefile and fine for new installations, but raises the question on how to update.

I guess for now we can merge this but we should add a note after 1.0 saying that in order to upgrade do kubectl patch instead. Anyway I don't think this is a high priority right now... so LGTM

juanluisvaladas (approved):

docs/install.md
makhov: Yes… `kubectl apply --server-side=true` does the trick as well, should we use it instead?

juanluisvaladas: I think I'd prefer using `--server-side=true`.

juanluisvaladas (approved):

PR #613, opened 2024-06-10
Finetune docs/commercial-support description

jnummelin:


docs/commercial-support.md
```diff
@@ -1,11 +1,13 @@
...
-Mirantis can provide various different levels of support starting from DevCare (9-to-5) all the way to OpsCare+ with fully managed service.
+Mirantis supports k0smotron three ways:
 
-Commercial support also includes support for k0smotron related tooling such as [k0s](https://github.com/k0sproject/k0s).
+* LabCare: 8 x 5 (M-F) business hours support. For non-production workloads
+* Opscare: 24 x 7 support with 30-minute response-time SLA for critical issues.
+* Opscare Plus: Fully-managed proactive support, with 15-minute response-time SLA for critical issues.
 
-If you are interested in commercial support for k0smotron check out our [support description](https://www.mirantis.com/support/enterprise-support-options/) and please [contact us](https://www.mirantis.com/contact/) for further details.
+k0smotron support includes support for [k0s](https://github.com/k0sproject/k0s), making use of k0s in k0smotron contexts effectively free.
```
juanluisvaladas, line 11: This line is a bit confusing. After reading it I understand I can get free support for both the mothership and the child clusters, however it's not clear to me if a third cluster which isn't the mothership or a child cluster would be included.

If I read only "k0smotron support includes support for [k0s](https://github.com/k0sproject/k0s)." I understand I can use k0s without k0smotron and still get support. But from this sentence I understand "making use of k0s in k0smotron contexts effectively free." it's only supported when used with k0smotron.

I think it should be either: 

"K0smotron support includes support for [k0s](https://github.com/k0sproject/k0s). Making use of k0s effectively free."

Or 

"K0smotron support includes support for [k0s](https://github.com/k0sproject/k0s) in k0smotron clusters.

Also the first K should be upper case.

juanluisvaladas (approved):

PR #616, opened 2024-06-12
PVC resizing support

### LINKED ISSUE #610: Persistent volume resize
If I recall correctly, the volumeClaimTemplate is immutable in StatefulSets, so we probably need to remove the sts with --cascade=orphan first and create a new one. And maybe also rollout restart after that.

makhov:
Fixes #610 

k0smotron removes and recreates the Statefulset with an updated size. If the StorageClass allows expansion, k0smotron resizes PVCs and deletes pods to apply the changed PVC. 

Also includes a minor etcd improvement, that makes check-etcd-scaling test pass

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -413,3 +413,13 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -236,6 +236,7 @@ rules:
   resources:
   - pods
   verbs:
+  - delete
   - get
   - list
 - apiGroups:
@@ -420,6 +421,4 @@ rules:
   verbs:
   - get
   - list
-  - patch
-  - update
   - watch
--- a/internal/controller/k0smotron.io/k0smotroncluster_etcd.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_etcd.go
@@ -130,7 +130,7 @@ func (r *ClusterReconciler) generateEtcdStatefulSet(kmc *km.Cluster, replicas in
 
 	size := kmc.Spec.Etcd.Persistence.Size
 
-	if n, _ := size.AsInt64(); n == 0 {
+	if size.IsZero() {
 		size = resource.MustParse("1Gi")
 	}
 	pvc := v1.PersistentVolumeClaim{
@@ -299,6 +299,18 @@ func (r *ClusterReconciler) initialCluster(kmc *km.Cluster, replicas int32) stri
 
 func (r *ClusterReconciler) generateEtcdInitContainers(kmc *km.Cluster) []v1.Container {
 	return []v1.Container{
+		{
+			// Wait for the pods dns name is resolvable, since it takes some time after the pod is created
+			// and etcd tries to connect to the other members using the dns names
+			Name:            "dns-check",
+			Image:           kmc.Spec.GetImage(),
+			ImagePullPolicy: v1.PullIfNotPresent,
+			Command:         []string{"/bin/bash", "-c"},
+			Args:            []string{"getent ahostsv4 ${HOSTNAME}.${SVC_NAME}." + kmc.Namespace + ".svc"},
+			Env: []v1.EnvVar{
+				{Name: "SVC_NAME", Value: kmc.GetEtcdServiceName()},
+			},
+		},
 		{
 			Name:            "init",
 			Image:           kmc.Spec.Etcd.Image,
--- a/internal/controller/k0smotron.io/k0smotroncluster_pvc.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_pvc.go
@@ -1,0 +1,171 @@
+package k0smotronio
+
+import (
+	"context"
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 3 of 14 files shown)


internal/controller/k0smotron.io/k0smotroncluster_pvc.go
```diff
@@ -0,0 +1,183 @@
...
+		} else {
+			// Do not check other PVCs if expansion is not allowed
+			break
+		}
+	}
+
+	return r.Delete(ctx, &sts, &client.DeleteOptions{PropagationPolicy: ptr.To(metav1.DeletePropagationOrphan)})
+}
+
+func (r *ClusterReconciler) reconcileEtcdPVC(ctx context.Context, kmc km.Cluster) error {
```
jnummelin, line 116: This implementation looks identical ti the `reconcilePVC` func, would it be possible to have single resize implementation for both cases? Normally some duplication is ok, but in this case the logic in not really trivial and hence having single implemntation for it would be easier to maintain.

makhov: Yeah, thanks! Done

config/rbac/role.yaml
```diff
@@ -413,3 +413,13 @@ rules:
...
   - patch
   - update
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - get
+  - list
+  - patch
```
jnummelin, line 423: I don't think we need to patch `storageclasses` objects 😄

internal/controller/k0smotron.io/k0smotroncluster_etcd.go
```diff
@@ -299,6 +299,16 @@ func (r *ClusterReconciler) initialCluster(kmc *km.Cluster, replicas int32) stri
...
+		{
+			Name:            "dns-check",
+			Image:           kmc.Spec.GetImage(),
+			ImagePullPolicy: v1.PullIfNotPresent,
+			Command:         []string{"/bin/bash", "-c"},
+			Args:            []string{"getent ahostsv4 ${HOSTNAME}.${SVC_NAME}." + kmc.Namespace + ".svc"},
+			Env: []v1.EnvVar{
+				{Name: "SVC_NAME", Value: kmc.GetEtcdServiceName()},
+			},
+		},
```
jnummelin, line 311: What's the purpose of this? I get that it checks the etcd svc can resolve, but why do we need it?

internal/controller/k0smotron.io/k0smotroncluster_pvc.go
```diff
@@ -0,0 +1,122 @@
...
+					Name:      fmt.Sprintf("%s-%d", stsName, i),
+					Namespace: kmc.Namespace,
+				},
+			}, &client.DeleteOptions{})
+
+			if err != nil {
+				return fmt.Errorf("failed to delete pod '%s' for resizing: %w", fmt.Sprintf("%s-%d", stsName, i), err)
+			}
+		} else {
+			// Do not check other PVCs if expansion is not allowed
```
jnummelin, line 116: hmm, maybe we should somehow tell user this? I see two options, update the reconcile state and/or create an event

inttest/pvc/pvc_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	err = wait.PollUntilContextCancel(s.Context(), time.Second, true, func(_ context.Context) (done bool, err error) {
+		sts, err := kc.AppsV1().StatefulSets("kmc-test").Get(s.Context(), "kmc-kmc-test", metav1.GetOptions{})
+		if err != nil {
+			return false, nil
+		}
+
+		return "200Mi" == sts.Spec.VolumeClaimTemplates[0].Spec.Resources.Requests.Storage().String(), nil
+	})
+	s.Require().NoError(err)
```
jnummelin, line 86: I guess in the current test setup we cannot check much more as the local provisioners do not really support resizing?

In optimal case I think we'd need to verify couple things:
- resize actually happened
- no data has been lost
- child cluster still behaves properly

internal/controller/k0smotron.io/k0smotroncluster_etcd.go
makhov: The check ensures the node is accessible from the other nodes during the join process.

inttest/pvc/pvc_test.go
```diff
@@ -0,0 +1,219 @@
...
+	s.T().Log("updating k0smotron cluster")
+	s.updateK0smotronCluster(s.Context(), rc)
+
+	err = wait.PollUntilContextCancel(s.Context(), time.Second, true, func(_ context.Context) (done bool, err error) {
+		sts, err := kc.AppsV1().StatefulSets("kmc-test").Get(s.Context(), "kmc-kmc-test", metav1.GetOptions{})
+		if err != nil {
+			return false, nil
+		}
+
+		return "250Mi" == sts.Spec.VolumeClaimTemplates[0].Spec.Resources.Requests.Storage().String(), nil
```
juanluisvaladas, line 113: I think we should also verify that the etcd volume has been resized to 70Mi

internal/controller/k0smotron.io/k0smotroncluster_etcd.go
```diff
@@ -299,6 +299,18 @@ func (r *ClusterReconciler) initialCluster(kmc *km.Cluster, replicas int32) stri
 
 func (r *ClusterReconciler) generateEtcdInitContainers(kmc *km.Cluster) []v1.Container {
 	return []v1.Container{
+		{
+			// Wait for the pods dns name is resolvable, since it takes some tima after the pod is created
```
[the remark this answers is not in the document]
makhov: Thanks, fixed!

inttest/pvc/pvc_test.go
makhov: Yeah, makes a lot of sense, thanks! Added the check for etcd statefulset

juanluisvaladas (approved):

config/rbac/role.yaml
makhov: fixed

internal/controller/k0smotron.io/k0smotroncluster_pvc.go
makhov: Done

inttest/pvc/pvc_test.go
makhov: Thanks! Updated the test.

PR #631, opened 2024-06-27
fix: set AutomountServiceAccountToken to false

cwrau:
this is not needed and it's best-practice to disable it
otherwise fails in clusters with policies forbidding this.


juanluisvaladas (approved):
Thanks for the contribution! LGTM

PR #641, opened 2024-07-03
Allow mounting any files to the HCP pods

### LINKED ISSUE #639: Config files for apiserver
In kubeadm you can additionally create files to be used by the kube-apiserver, e.g. https://github.com/teutonet/teutonet-helm-charts/blob/main/charts/t8s-cluster/files/admission-control-config.yaml, is there a way to do this in k0smotron?

I looked at the code and didn't see any configmap volumes or something like that, but maybe via the k0s config?

makhov:
Fixes #639

The potential use case might be to use apiserver admission configuration
https://docs.k0sproject.io/stable/podsecurity/


api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -84,6 +84,13 @@ type ClusterSpec struct {
 	// https://kubernetes.io/docs/concepts/storage/volumes
 	//+kubebuilder:validation:Optional
 	Manifests []v1.Volume `json:"manifests,omitempty"`
+	// Files allows to specify list of volumes with any files to be
+	// mounted in the controlplane pod. K0smotron allows any kind of volume, but the
+	// recommendation is to use secrets and configmaps.
+	// For more information check:
+	// https://kubernetes.io/docs/concepts/storage/volumes
+	//+kubebuilder:validation:Optional
+	Files []Files `json:"files,omitempty"`
```
jnummelin, line 93: Could this be called something like `Volumes` or `Mounts`. IMO files sounds bit wrong as we're really not doing files per se directly

makhov: Yeah, makes sense. Renamed it to `Mounts`, thanks.

PR #667, opened 2024-08-07
K0sControlPlane update strategy Recreate

makhov:
Added support for `K0sControlPlane.spec.updateStrategy=Recreate`, with which k0smotron now supports the usual CAPI way of updating controllers.


jnummelin (changes requested):
We need to handle the special case of running controller with `--single` _somehow_. We can't recreate that `Machine` as that would end up in loss of ALL data. I guess we could error out early, with some proper status saying "we can't recreate a controller with --single". Also we need to document that one should never even try to do that 😄 

Once we have the proper admission controller, we can then prevent users creating this corner case.

jnummelin (approved):

PR #685, opened 2024-08-16
Merge user-provided and k0smotron-defined k0s args

makhov:
Parses and merges `--kubelet-extra-args` to add `--hostname-override` if needed.


jnummelin (changes requested):
I really think we should have some testing for this args handling

makhov:
Added a test, thanks!

PR #687, opened 2024-08-19
Use Machine's name as a k8s node name by default

makhov:
Adds `--hostname-override` to the worker nodes to match the kubernetes node name and machine name. Can be disabled by setting the `UseSystemHostname` to true.

Also `mergeExtraArgs` func improved to work both with controller and worker args.

### DIFF
--- a/api/bootstrap/v1beta1/k0s_types.go
+++ b/api/bootstrap/v1beta1/k0s_types.go
@@ -64,6 +64,12 @@ type K0sWorkerConfigSpec struct {
 	// +kubebuilder:validation:Optional
 	Version string `json:"version,omitempty"`
 
+	// UseSystemHostname specifies whether to use the system hostname for the kubernetes node name.
+	// By default, k0smotron will use Machine name as a node name. If true, it will pick it from `hostname` command output.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	UseSystemHostname bool `json:"useSystemHostname,omitempty"`
+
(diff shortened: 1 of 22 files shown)


api/bootstrap/v1beta1/k0s_types.go
```diff
@@ -64,6 +64,11 @@ type K0sWorkerConfigSpec struct {
 	// +kubebuilder:validation:Optional
 	Version string `json:"version,omitempty"`
 
+	// UseSystemHostname specifies whether to use the system hostname for the kubernetes node name.
+	// By default, k0s will uses Machine name as a node name. If true, it will pick it from `hostname` command output.
```
jnummelin, line 68:
```suggestion
	// By default, k0smotron will use Machine name as a node name. If true, it will pick it from `hostname` command output.
```

api/bootstrap/v1beta1/k0s_types.go
```diff
@@ -64,6 +64,11 @@ type K0sWorkerConfigSpec struct {
...
+	// +kubebuilder:validation:Optional
```
jnummelin, line 69: Just to be explicit, should we define `false` as the default value?

makhov: Sorry, forgot about this. Fixed!

api/bootstrap/v1beta1/k0s_types.go
makhov: Thanks, fixed!

jnummelin (approved):

PR #696, opened 2024-08-28
ContentFrom field in bootstrap config

### LINKED ISSUE #692: Add ContentFrom option for Files configuration in K0sConfigSpec and K0sWorkerConfigSpec
## Request 

`ContentFrom` option should be added to bootstrap API to give an ability to reference secret content as a file for cloud-init configuration.

The same functionality is present in [KubeadmConfig](https://github.com/kubernetes-sigs/cluster-api/blob/88c757fa5e2308eedacb100213b00b805bc732a3/bootstrap/kubeadm/api/v1beta1/kubeadmconfig_types.go#L525) for CAPI kubeadm bootstrap provider.

## Rationale 

Some infrastructure providers (like Azure CAPI provider) are using this functionality to pass infrastructure specific configuration to all nodes. In case of CAPZ the dynamically generated `azure.json` from the `Secret` object to all nodes. This file is required for CCM to function properly.

Example gist of `contentFrom` usage for kubeadm+azure:

```yaml
    files:
    - contentFrom:
        secret:
          key: control-plane-azure.json
          name: az-test-control-plane-azure-json
      owner: root:root
      path: /etc/kubernetes/azure.json
      permissions: "0644"
```

Documentation reference for Azure provider is [here](https://capz.sigs.k8s.io/topics/cloud-provider-config)

makhov:
Fixed #692

Secret and ConfigMap support as a source of files in bootstrap config


api/bootstrap/v1beta1/k0s_types.go
```diff
@@ -66,7 +66,7 @@ type K0sWorkerConfigSpec struct {
 
 	// Files specifies extra files to be passed to user_data upon creation.
 	// +kubebuilder:validation:Optional
-	Files []cloudinit.File `json:"files,omitempty"`
+	Files []File `json:"files,omitempty"`
```
jnummelin, line 69: Could we not enhance the `cloudinit.File` struct directly instead of creating new one? I see that we "inherit" from it but somehow feels bit wonky 😄

makhov: Well, technically we can, but we will have to move ContentSource and ContentSourceRef structs to the cloudinit package. Is it ok?

jnummelin: IMO it makes sense to have the cloudinit related things in same package

makhov: Well, not sure it's directly related to cloud-init, it's more a type that will be converted to cloudinit.File. 

I tried to simply move it to the cloudinit package, but it breaks `zz_generated.deepcopy.go` generation, because it requires `DeepCopyInto` method of the complex struct (with pointers to other structs):

```
if in.Files != nil {
	in, out := &in.Files, &out.Files
	*out = make([]cloudinit.File, len(*in))
	for i := range *in {
		(*in)[i].DeepCopyInto(&(*out)[i])
	}	
}
```

jnummelin: > I tried to simply move it to the cloudinit package, but it breaks zz_generated.deepcopy.go generation, because it requires DeepCopyInto method of the complex struct (with pointers to other structs)

ok, that's enough proof that it's more effort to move it rather than have it as is. so IMO ok to go ahead as-is!

jnummelin (approved):
LGTM

PR #703, opened 2024-09-02
Update k0smotron dependencies

juanluisvaladas:
K0smotron currently imports k0s 1.27 which is nearing the end of its life and has multiple vulnerable imports (which I believe aren’t affecting us).

In order to upgrade this we need to upgrade k8s.io/api, apimachinery, client-go, Kubernetes and controller-runtime as well which eventually makes us need to upgrade cluster API

There are 3 big changes:
1- Upgraded k8s version
2- Upgraded cluster API version to v0.18.
3- Moved from footloose to bootloose.


juanluisvaladas:
Closed in favor of https://github.com/k0sproject/k0smotron/pull/712

PR #725, opened 2024-09-12
Fix k0s.yaml config generation

makhov:
Currently, dynamic config reconciliation mutates given k0s config, which causes some fields are missing in the static config


jnummelin (approved):
LGTM

PR #747, opened 2024-09-26
Ensure, that object names are shorter than the kubernetes limit

### LINKED ISSUE #739: Names of resources created by resources are too long Cluster.k0smotron.io/v1beta1 are too long
I've hit an issue with the names of resources created by `"apiVersion":"k0smotron.io/v1beta1","kind":"Cluster"`. If you create this resource with name that exceeds a certain length, the StatefulSet created by Cluster cannot create underlying pods, because one of the pods has the volume with the name that exceeds 63 chars (which is the limit in k8s). See this error msg that I found by kubectl describing sts:
`Warning  FailedCreate      9m3s (x18 over 19m)  statefulset-controller  create Pod kmc-dk-enc-nodes-basic-sep17-v5-lg5hz-njttw-0 in StatefulSet kmc-dk-enc-nodes-basic-sep17-v5-lg5hz-njttw failed error: Pod "kmc-dk-enc-nodes-basic-sep17-v5-lg5hz-njttw-0" is invalid: [spec.volumes[3].name: Invalid value: "kmc-prometheus-dk-enc-nodes-basic-sep17-v5-lg5hz-njttw-config-nginx": must be no more than 63 characters, spec.containers[2].volumeMounts[0].name: Not found: "kmc-prometheus-dk-enc-nodes-basic-sep17-v5-lg5hz-njttw-config-nginx"]`

The volume name seems to be created by concatenating:
1. `kmc-prometheus-`
2. Cluster name
3. `-config-nginx`

1 and 3 eat up together **28** characters, leaving the users with 63-28=35 characters (only 55%), which might be a problem if the Cluster is created by some other resource, like for example Crossplane Composition in my case. 

The name of the volume could just be `config-nginx`, without using the parent resource name. This issue might be also triggered by https://github.com/k0sproject/k0smotron/blob/v1.1.0/api/k0smotron.io/v1beta1/k0smotroncluster_types.go#L323-L325

makhov:
Fixes #739

### DIFF
--- a/api/k0smotron.io/v1beta1/k0smotroncluster_types.go
+++ b/api/k0smotron.io/v1beta1/k0smotroncluster_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1beta1
 
 import (
+	"crypto/md5"
 	"fmt"
 	"strings"
 
@@ -304,7 +305,7 @@ func init() {
 }
 
 func GetStatefulSetName(clusterName string) string {
-	return fmt.Sprintf("kmc-%s", clusterName)
+	return shortName(fmt.Sprintf("kmc-%s", clusterName))
 }
 
 func (kmc *Cluster) GetStatefulSetName() string {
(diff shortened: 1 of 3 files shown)


api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -350,21 +356,34 @@ func (kmc *Cluster) GetServiceName() string {
...
-	return fmt.Sprintf("kmc-%s-nodeport", kmc.Name)
+	return kmc.getObjectName("kmc-%s-nodeport")
 }
 
 func (kmc *Cluster) GetVolumeName() string {
-	return fmt.Sprintf("kmc-%s", kmc.Name)
+	return kmc.getObjectName("kmc-%s")
+}
+
+const limit = 63
```
apedriza, line 378: maybe adding a reason of the value as a const comment can add more context

apedriza (approved):
lgtm

PR #754, opened 2024-09-30
Use PVC retention policy to allow pvc deletion with statefulset

### LINKED ISSUE #675: Add an ability to delete hosted cluster state
_This is more like an ask and if such feature is already present please point me to the corresponding docs, because I couldn't find it._

## Description

Right now if I delete k0smotron cluster (hosted) all state (PVCs) doesn't get deleted, so I should clean everything up by hands and then recreate the cluster.

This is ok, until I need to conduct some tests or have some disposable clusters. In this case 2 things happen, which aren't convenient:

1. PVCs linger in the cluster and consuming resources in the infrastructure. So this should be handled separately by some additional controller or manually.
2. If same name is used for the cluster old objects (like `Nodes`) are lingering. This makes cluster less stable and more confusing, since old resources (which don't exist) should be cleaned up by hands.

## Proposal

If would be much more convenient to have a flag that should be explicitly set by user to delete all cluster state in case of cluster deletion.

This will make process more declarative avoiding lingering resources or confusing state in case of cluster re-deployment.

### LINKED ISSUE #750: Cluster.k0smotron.io/v1beta1 does not offer an ability to delete PVC upon its deletion
Basically this is feature request: make it possible to override StatefulSet's `persistentVolumeClaimRetentionPolicy`. We've been using Clusters a lot lately and after a while we've got ~200 stale, old unneeded PVCs

Thank you!

makhov:
Fixes #750 
Fixes #675

### DIFF
--- a/api/k0smotron.io/v1beta1/k0smotroncluster_types.go
+++ b/api/k0smotron.io/v1beta1/k0smotroncluster_types.go
@@ -196,6 +196,10 @@ type PersistenceSpec struct {
 	// PersistentVolumeClaim defines the PVC configuration. Will be used as is in case of .spec.persistence.type is pvc.
 	//+kubebuilder:validation:Optional
 	PersistentVolumeClaim *PersistentVolumeClaim `json:"persistentVolumeClaim,omitempty"`
+	// AutoDeletePVCs defines whether the PVC should be deleted when the cluster is deleted.
+	//+kubebuilder:default=false
+	//+kubebuilder:validation:Optional
+	AutoDeletePVCs bool `json:"autoDeletePVCs,omitempty"`
 	// HostPath defines the host path configuration. Will be used as is in case of .spec.persistence.type is hostPath.
 	//+kubebuilder:validation:Optional
 	HostPath string `json:"hostPath,omitempty"`
@@ -261,6 +265,10 @@ type EtcdSpec struct {
 	// Persistence defines the persistence configuration.
 	//+kubebuilder:validation:Optional
 	Persistence EtcdPersistenceSpec `json:"persistence"`
+	// AutoDeletePVCs defines whether the PVC should be deleted when the etcd cluster is deleted.
+	//+kubebuilder:default=false
+	//+kubebuilder:validation:Optional
+	AutoDeletePVCs bool `json:"autoDeletePVCs,omitempty"`
 	// DefragJob defines the etcd defragmentation job configuration.
 	//+kubebuilder:validation:Optional
 	DefragJob DefragJob `json:"defragJob"`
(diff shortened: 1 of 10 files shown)


api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -196,6 +196,10 @@ type PersistenceSpec struct {
 	// PersistentVolumeClaim defines the PVC configuration. Will be used as is in case of .spec.persistence.type is pvc.
 	//+kubebuilder:validation:Optional
 	PersistentVolumeClaim *PersistentVolumeClaim `json:"persistentVolumeClaim,omitempty"`
+	// PVCDeletionAllowed defines whether the PVC should be deleted when the cluster is deleted.
+	//+kubebuilder:default=false
+	//+kubebuilder:validation:Optional
+	PVCDeletionAllowed bool `json:"pvcDeletionAllowed,omitempty"`
```
jnummelin, line 202: IMO the name does not fully match the behaviour. Some alternative suggestions:
- `DeleteWithCluster`
- `DeletePVCsWithCluster`
- `AutoDeletePVCs`

internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
```diff
@@ -200,6 +200,12 @@ func (r *ClusterReconciler) generateStatefulSet(kmc *km.Cluster) (apps.StatefulS
 		if kmc.Spec.Persistence.PersistentVolumeClaim.Name == "" {
 			kmc.Spec.Persistence.PersistentVolumeClaim.Name = kmc.GetVolumeName()
 		}
+
+		if kmc.Spec.Persistence.AutoDeletePVCs {
+			statefulSet.Spec.PersistentVolumeClaimRetentionPolicy = &apps.StatefulSetPersistentVolumeClaimRetentionPolicy{
+				WhenDeleted: apps.RetainPersistentVolumeClaimRetentionPolicyType,
```
apedriza, line 206:
```suggestion
				WhenDeleted: apps.DeletePersistentVolumeClaimRetentionPolicyType,
```
As I have understood the problem, shouldn't this type of removal policy be followed if the flag is true?

apedriza (commented):
ETCD PVCs are not removed, should we add this policy to the ETCD StatefulSet as well?

internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
makhov: Heh, great catch, thanks! Fixed!

apedriza (approved):

Issue #650, opened 2024-07-14
Document k0s persistence


kbumsik:
Hi, I am experimenting k0smotron to implement multi-tenant k8s with remote workers.

While playing around k0smotron, one thing I noticed is that documentation on [Cluster.spec.persistence](https://docs.k0smotron.io/stable/resource-reference/#clusterspecpersistence) is lacking, while it seems to be important for production deployment.

For example:

- What it actually does - the doc doesn't mention what it persists at all. Just saying "persistence configuration" is very confusing. It seems to mount /var/lib/k0s though.
-  Recommended size and performance of the volume
    - dose it need to be high-performance? Suggested IOPS? What is the expected size of the volume? Does it keep growing? 
    - In case of etcd, [it is easier to figure out the optimal pvc settings.](https://etcd.io/docs/v3.2/platforms/aws/#performancethroughput) However, it is hard to figure out the hardware requirements of k0s control plane instance.
- Maybe also add why it needs a persistent volume. I thought kube-apiserver is stateless.

Since v1 is out, some insights regarding production deployments like this would really helps.

nealf:
I'm in the same situation of not knowing whether it is needed or not for non-ephemeral clusters. If it is necessary, we found that it does not currently work in Standalone mode (at least when replicas >1) because it doesn't use the StatefuleSets persistent volume template correctly. It creates the volume template, but the VolumeMount doesn't use the template name and pods are never created because of that config error.

apedriza:
> I'm in the same situation of not knowing whether it is needed or not for non-ephemeral clusters. If it is necessary, we found that it does not currently work in Standalone mode (at least when replicas >1) because it doesn't use the StatefuleSets persistent volume template correctly. It creates the volume template, but the VolumeMount doesn't use the template name and pods are never created because of that config error.

@nealf thanks for reporting this issue 👍  Fix for it was released in `v.1.12`

apedriza:
> Hi, I am experimenting k0smotron to implement multi-tenant k8s with remote workers.
> 
> While playing around k0smotron, one thing I noticed is that documentation on [Cluster.spec.persistence](https://docs.k0smotron.io/stable/resource-reference/#clusterspecpersistence) is lacking, while it seems to be important for production deployment.
> 
> For example:
> 
> * What it actually does - the doc doesn't mention what it persists at all. Just saying "persistence configuration" is very confusing. It seems to mount /var/lib/k0s though.
> * Recommended size and performance of the volume
>   
[...]

PR #756, opened 2024-10-01
Fix PVC binding when Cluster PVC name is given

apedriza:
fix https://github.com/k0sproject/k0smotron/issues/650#issuecomment-2346707289

When the cluster persistence was of type`pvc` and the cluster name was given by the user, the statefulset related to the k0smotron Cluster could not create the pods because the init container was using the wrong mount name. It was using the name `kmc-<clustername>` instead of the one configured by the user.

### HOW TO REPRO ISSUE 
1. Create a cluster with a PVC name:
```yml
apiVersion: k0smotron.io/v1beta1
kind: Cluster
metadata:
  name: k0smotron-test
spec:
  replicas: 1
  image: k0sproject/k0s
  version: v1.27.1-k0s.0
  service:
    type: NodePort
    apiPort: 30443
    konnectivityPort: 30132
  persistence:
    type: pvc
    persistentVolumeClaim:
      metadata:
        name: my-pvc-test
      spec:
        accessModes:
        - ReadWriteOnce
        storageClassName: standard
        resources: 
          requests: 
            storage: 200Mi
```
2. Check StatefulSet pods are not created because the PVC used by the init container mount is not found.

_After this fix, mount name used in the init container is the one given by the user in the Cluster declaration._

### DIFF
--- a/inttest/pvc/pvc_test.go
+++ b/inttest/pvc/pvc_test.go
@@ -19,7 +19,6 @@ package pvc
 import (
 	"context"
 	"fmt"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


inttest/pvc/pvc_test.go
```diff
@@ -172,6 +172,9 @@ func (s *PVCSuite) createK0smotronCluster(ctx context.Context, kc *kubernetes.Cl
 			"persistence": {
 				"type": "pvc",
 				"persistentVolumeClaim": {
+					"metadata": {
+						"name": "kmc-volume-test"
+					},
```
apedriza, line 177: this triggers the issue

PR #760, opened 2024-10-03
Rely on machines instead of control plane status for machine update

makhov:
Check the actual machine state instead of relying on the status field alone.

### DIFF
--- a/internal/controller/controlplane/helper.go
+++ b/internal/controller/controlplane/helper.go
@@ -53,12 +53,19 @@ func (c *K0sController) deleteMachine(ctx context.Context, name string, kcp *cpv
 	return nil
 }
 
-func (c *K0sController) generateMachine(_ context.Context, name string, cluster *clusterv1.Cluster, kcp *cpv1beta1.K0sControlPlane, infraRef corev1.ObjectReference) (*clusterv1.Machine, error) {
+func machineVersionFromControlPlaneVersion(kcp *cpv1beta1.K0sControlPlane) (string, error) {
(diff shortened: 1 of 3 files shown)


internal/controller/controlplane/helper.go
```diff
@@ -53,12 +53,19 @@ func (c *K0sController) deleteMachine(ctx context.Context, name string, kcp *cpv
...
 }
 
-func (c *K0sController) generateMachine(_ context.Context, name string, cluster *clusterv1.Cluster, kcp *cpv1beta1.K0sControlPlane, infraRef corev1.ObjectReference) (*clusterv1.Machine, error) {
+func machineVersionFromControlPlaneVersion(kcp *cpv1beta1.K0sControlPlane) (string, error) {
 	ver, err := semver.NewVersion(kcp.Spec.Version)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing version %q: %w", kcp.Spec.Version, err)
+		return "", fmt.Errorf("error parsing version %q: %w", kcp.Spec.Version, err)
+	}
+	return fmt.Sprintf("v%d.%d.%d", ver.Major(), ver.Minor(), ver.Patch()), nil
```
apedriza, line 61: what are the implications of changing version format `x.x.x` to `vx.x.x`?

makhov: This is the fun part, because v* is actually the only correct notation. I guess, I did it by accident and the CAPI was fixing it for us

PR #764, opened 2024-10-07
Add test for k0smotron upgrade

apedriza:
This test recreates an upgrade of the k0smotron operator and checks that the behavior remains as expected after that.

### DIFF
--- a/inttest/upgrade/upgrade_test.go
+++ b/inttest/upgrade/upgrade_test.go
@@ -1,0 +1,266 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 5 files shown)


inttest/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// If the status of the Owner is correct, the status of all resources
+	// that have OwnerRef to this resource should also be correct.
+	var kmc km.Cluster
+	err = yaml.Unmarshal(result, &kmc)
+	s.Require().NoError(err)
+	s.Require().True(kmc.Status.Ready)
+}
+
+func (s *UpgradeSuite) checkStatePersists(ctx context.Context, mountedPath string, kc *kubernetes.Clientset, rc *rest.Config) {
```
makhov, line 124: We also need to check the cluster state in etcd, eg create some object using API before the upgrade and check if it still exists.

apedriza: done 👍

PR #771, opened 2024-10-10
Check if controlPlaneEndpoint can accept join requests

makhov:
Adding a probe on k0s api port to check, if the control plane LB has this port open. Some LBs open only kubernetes API port, so they can't be used for joining controllers. On the other hand, we can't always use the machine's address, since some providers report addresses incorrectly

### DIFF
--- a/internal/controller/bootstrap/controlplane_bootstrap_controller.go
+++ b/internal/controller/bootstrap/controlplane_bootstrap_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"net/netip"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -595,6 +594,29 @@ func mergeControllerExtraArgs(scope *ControllerScope) []string {
...
 }
 
+func (c *ControlPlaneController) detectJoinHost(ctx context.Context, scope *ControllerScope, firstControllerMachine *clusterv1.Machine) (string, error) {
+	httpClient := &http.Client{
+		Transport: c.RESTConfig.Transport,
+		Timeout:   time.Second,
+	}
+
+	// TODO: fix hardcoded port
+	port := "9443"
```
apedriza, line 604: nit: worth use a `const` for k0s api port?

apedriza (approved):

PR #774, opened 2024-10-14
Add missing K0sControlPlane status fields and refactor status calc

jnummelin:
`K0sControlPlane` was missing some required status fields relating to replicas. https://cluster-api.sigs.k8s.io/developer/providers/control-plane?search=pause#required-status-fields-for-implementations-using-replicas

While this doesn't really break anything when using this in "standalone" fashion, e.g. without ClusterClass, things go south when using with ClusterClass. What happens is that when the CAPI topology controller does not see the expected fields, it'll determine that the CP is still scaling up and hence prevents users from upgrading CP via CC.

Previously we dropped the k0s version suffix from the Machines, in this PR we change it to always include the suffix if given in the CP version. This is to allow e.g. `v1.30.0+k0s.0` --> `v1.30.0+k0s.1` type upgrades.


internal/controller/controlplane/status.go
```diff
@@ -0,0 +1,195 @@
...
+
+	machineVersion := *machine.Spec.Version
+	kcpVersion := ver
+
+	// If either of the versions is missing the suffix, we need to add it
+	// But take the suffix from kcp version if present
+	kcpSuffix := getVersionSuffix(kcpVersion)
+	if kcpSuffix == "" {
+		kcpVersion = kcpVersion + "+k0s.0"
+		kcpSuffix = "k0s.0"
```
apedriza, line 59:
```suggestion
                kcpSuffix = "k0s.0"
		kcpVersion = kcpVersion + kcpSuffix
```
nit: I think less error prone

jnummelin:
Thanks @apedriza , pushed the suggested changes

internal/controller/controlplane/status.go
```diff
@@ -0,0 +1,185 @@
...
+	kcp.Status.Replicas = int32(len(machines))
+	readyReplicas := 0
+	updatedReplicas := 0
+	unavailableReplicas := 0
+	// Count the machines in different states
+	for _, machine := range machines {
+		// TODO Do we need to distinguish between Running and Provisioned
+		// when we run ctrl node without --enable-worker?
+		switch machine.Status.Phase {
+		case string(clusterv1.MachinePhaseRunning):
```
makhov, line 90: We need to add `string(clusterv1.MachinePhaseProvisioned)` as well here. 
Otherwise, the status looks like this:
```
    ready: true
    readyReplicas: 0
    replicas: 3
    selector: cluster.x-k8s.io/cluster-name=aws-test-cluster,cluster.x-k8s.io/control-plane
    unavailableReplicas: 3
    updatedReplicas: 3
    version: v1.30.1+k0s.0
```

jnummelin: Pushed handling for this, PTAL

internal/controller/controlplane/status.go
```diff
@@ -0,0 +1,188 @@
...
+	unavailableReplicas := 0
+	// Count the machines in different states
+	for _, machine := range machines {
+		switch machine.Status.Phase {
+		case string(clusterv1.MachinePhaseRunning):
+			readyReplicas++
+		case string(clusterv1.MachinePhaseProvisioned):
+			// If we're running without --enable-worker, the machine will never transition
+			// to running state, so we need to count it as ready when it's provisioned
+			if !kcp.WorkerEnabled() {
```
makhov, line 92: The only thing, that if the worker is enabled, it'll still hit this case and won't be counted for unavailableReplicas

jnummelin: Good catch. Pushed a commit fixing this

makhov (approved):

PR #778, opened 2024-10-15
Add machine names as etcd name

### LINKED ISSUE #768: clean up controlnodes and etcdmembers after Machine removal
if a Machine is deleted, the reference to its controlnode persists as does its etcdmember.

### How to repro:
1. Deploy https://github.com/k0sproject/k0smotron/blob/main/inttest/capi-controlplane-docker/capi_controlplane_docker_test.go#L212
2. Once environment is ready,  delete a Machine related to a controller
3. Check controlnode and etcmember persists by executing:
`k0s kubectl get controlnodes` 
`k0s etcd member-list` 

ref: https://docs.k0sproject.io/v1.28.4+k0s.0/remove_controller/

apedriza:
Partially fix #768

### DIFF
--- a/internal/controller/bootstrap/controlplane_bootstrap_controller.go
+++ b/internal/controller/bootstrap/controlplane_bootstrap_controller.go
@@ -52,6 +52,7 @@ import (
 	bootstrapv1 "github.com/k0sproject/k0smotron/api/bootstrap/v1beta1"
 	"github.com/k0sproject/k0smotron/internal/cloudinit"
 	kutil "github.com/k0sproject/k0smotron/internal/util"
+	"github.com/k0sproject/version"
 )
 
 type ControlPlaneController struct {
@@ -62,6 +63,8 @@ type ControlPlaneController struct {
 }
 
 const joinTokenFilePath = "/etc/k0s.token"
+
+var minVersionForETCDName = version.MustParse("v1.31.1+k0s.0")
 
 // +kubebuilder:rbac:groups=bootstrap.cluster.x-k8s.io,resources=k0scontrollerconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=bootstrap.cluster.x-k8s.io,resources=k0scontrollerconfigs/status,verbs=get;list;watch;create;update;patch;delete
@@ -155,6 +158,7 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
 	)
 
 	if config.Spec.K0s != nil {
+
 		nllbEnabled, found, err := unstructured.NestedBool(config.Spec.K0s.Object, "spec", "network", "nodeLocalLoadBalancing", "enabled")
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("error getting nodeLocalLoadBalancing: %v", err)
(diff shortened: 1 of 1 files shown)


internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -154,6 +154,18 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
 	)
 
 	if config.Spec.K0s != nil {
+		storage, found, err := unstructured.NestedMap(config.Spec.K0s.Object, "spec", "storage", "etcd")
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("error retrieving storage.etcd: %w", err)
+
+		}
+		if found && storage != nil {
```
makhov, line 162: We probably need to set it even if it's not found in the original config

apedriza: yes, I added a check to set it always if kind storage is not set

makhov (approved):

apedriza:
`capi-controlplane-docker` test failing. Converting PR to draft until fix it

internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -155,6 +156,25 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
 	)
 
 	if config.Spec.K0s != nil {
+
+		minVersionForETCDName := version.MustParse("v1.31.1+k0s.0")
```
makhov, line 160: let's move it to the beginning of the file to some var to avoid calling `MustParse` on each reconciliation.

internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -155,6 +156,25 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
...
+		currentKCPVersion := version.MustParse(config.Spec.Version)
```
makhov, line 161: It's shouldn't happen, but just to be safe, let's return an error instead of panic here

internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -155,6 +156,25 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
 	)
 
 	if config.Spec.K0s != nil {
```
makhov, line 158: A few thoughts:
- Should we set the field even if user didn't provide us any custom k0s config?
- At the end of this if statement, we call `kutil.ReconcileDynamicConfig`, but we probably need to remove the etcd node name from the DynamicConfig. 
- Should the most of the code of this statement be in the control plane controller instead of bootstrap controller?

apedriza: > * Should we set the field even if user didn't provide us any custom k0s config?

Agree. It is already used etcd node name as default value if kine is not set so I think is safe assume that. Added
> * At the end of this if statement, we call `kutil.ReconcileDynamicConfig`, but we probably need to remove the etcd node name from the DynamicConfig.

Im not sure if I follow your point. It is not removed `spec.storage` in `kutil.ReconcileDynamicConfig` from the DynamicConfig?
> * Should the most of the code of this statement be in the control plane controller instead of bootstrap controller?

If you mean to move most of the logic to the place when we create `bootstrapv1.K0sControllerConfig` Im ok with that. But I think that change should be done carefully, Im not sure exactly what it might imply rn. Maybe addressing that refactoring in another PR makes more sense, wdyt?

makhov: > Im not sure if I follow your point. It is not removed spec.storage in kutil.ReconcileDynamicConfig from the DynamicConfig?

Sorry, missed it. Yeah, that's good then. 

> Maybe addressing that refactoring in another PR makes more sense, wdyt?

Sorry, I was not clear enough: of course, it's better to do it in a separate PR.

apedriza: https://github.com/k0sproject/k0smotron/issues/798

makhov (approved):

PR #788, opened 2024-10-21
Inttest/capi clusterclass recreate upgrade

jnummelin:
This PR adds a CAPI related test for ClusterClass and upgrading a cluster via `Cluster` topology

Builds currently on top of #774 , once that lands I'll rebase. For now look at the last commit only. 😄

### DIFF
--- a/inttest/capi-docker-clusterclass-recreate-upgrade/capi_docker_clusterclass_test.go
+++ b/inttest/capi-docker-clusterclass-recreate-upgrade/capi_docker_clusterclass_test.go
@@ -1,0 +1,420 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


apedriza:
I think we need to add this test to CI

jnummelin:
> I think we need to add this test to CI

Absolutely. For some reason my brain was telling me that if I add it to `inttest/Makefile.variables` it'll be picked up. 🤦

inttest/capi-docker-clusterclass-recreate-upgrade/capi_docker_clusterclass_test.go
```diff
@@ -0,0 +1,420 @@
...
+	// Update the cluster and wait for the reported version to change
+	s.T().Log("updating cluster")
+	cluster, err := util.GetCluster(s.ctx, s.client, "docker-test-cluster", "default")
+	s.Require().NoError(err)
+
+	cluster.Spec.Topology.Version = "v1.30.1"
+	err = util.UpdateCluster(s.ctx, s.client, cluster)
+	s.Require().NoError(err)
+
+	s.T().Log("waiting for control-plane nodes to be updated")
```
apedriza, line 210: Is it worth checking that the worker nodes are also updated with new version?

apedriza (approved):

PR #789, opened 2024-10-22
Fix machine implementation updating

makhov:
Merge and patch existing machine objects to avoid owner conflict and/or admission webhook issues

### DIFF
--- a/internal/controller/controlplane/helper.go
+++ b/internal/controller/controlplane/helper.go
@@ -134,10 +134,17 @@ func (c *K0sController) createMachineFromTemplate(ctx context.Context, name stri
 		return nil, err
 	}
 
-	resourceType := strings.ToLower(machineFromTemplate.GetKind()) + "s"
+	pluralName := ""
+	resList, _ := c.ClientSet.Discovery().ServerResourcesForGroupVersion(existingMachineFromTemplate.GetAPIVersion())
+	for _, apiRes := range resList.APIResources {
+		if apiRes.Kind == existingMachineFromTemplate.GetKind() && !strings.Contains(apiRes.Name, "/") {
+			pluralName = apiRes.Name
+			break
+		}
+	}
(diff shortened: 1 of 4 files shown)


internal/controller/controlplane/helper.go
```diff
@@ -102,12 +103,45 @@ func (c *K0sController) createMachineFromTemplate(ctx context.Context, name stri
...
+	spec, _, _ := unstructured.NestedMap(existingMachineFromTemplate.Object, "spec")
+	patch := unstructured.Unstructured{Object: map[string]interface{}{
+		"spec": spec,
+	}}
+	data, err := patch.MarshalJSON()
+	if err != nil {
+		return nil, err
+	}
+
+	resourceType := strings.ToLower(machineFromTemplate.GetKind()) + "s"
```
apedriza, line 137: wdyt about find the apiresource.Name using resource kind and rely on the crd plural?

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -61,9 +62,10 @@ var ErrNewMachinesNotReady = fmt.Errorf("waiting for new machines")
 
 type K0sController struct {
 	client.Client
-	Scheme     *runtime.Scheme
-	ClientSet  *kubernetes.Clientset
-	RESTConfig *rest.Config
+	Scheme        *runtime.Scheme
+	ClientSet     *kubernetes.Clientset
+	DynamicClient dynamic.Interface
```
apedriza, line 67: do we need `DynamicClient` ?

makhov: Not anymore. Removed it, thanks!

internal/controller/controlplane/helper.go
makhov: Changed it to more reliable way

apedriza (approved):

PR #790, opened 2024-10-22
Clean up controlnode if machine related to K0sControlPlane is deleted

### LINKED ISSUE #768: clean up controlnodes and etcdmembers after Machine removal

apedriza:
Fix #768 

Follow up PR to https://github.com/k0sproject/k0s/pull/5128. We need to use machine name as `ControlNode` name in order to be able to delete it when applies

### DIFF
--- a/internal/controller/controlplane/k0s_controlplane_controller.go
+++ b/internal/controller/controlplane/k0s_controlplane_controller.go
@@ -418,6 +418,10 @@ func (c *K0sController) reconcileMachines(ctx context.Context, cluster *clusterv
 			return replicasToReport, fmt.Errorf("error deleting machine from template: %w", err)
 		}
 
+		if err := c.deleteControlNode(ctx, name, kubeClient); err != nil {
+			return replicasToReport, fmt.Errorf("error deleting controlnode: %w", err)
+		}
+
 		logger.Info("Deleted machine", "machine", name, "replicasToReport", replicasToReport)
 		return replicasToReport, nil
 	}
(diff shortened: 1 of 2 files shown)


internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -406,6 +406,10 @@ func (c *K0sController) reconcileMachines(ctx context.Context, cluster *clusterv
 			return replicasToReport, fmt.Errorf("error marking controlnode to leave: %w", err)
 		}
 
+		if err := c.deleteControlNode(ctx, name, kubeClient); err != nil {
```
apedriza, line 409: @makhov I dont know which logic reacts to this https://github.com/k0sproject/k0smotron/blob/main/internal/controller/controlplane/helper.go#L182-L189 so I thought is safe to remove controlnode at this point. WDYT?

makhov: No, I think it's too early, let's do it at the very end, after the machine object is deleted

apedriza: done

makhov (approved):

PR #800, opened 2024-10-30
Fix cluster annotations propagation to statefulset and pods

### LINKED ISSUE #735: Annotations propagation from standalone cluster metadata
It seems that currently https://docs.k0smotron.io/stable/resource-reference/#cluster `metadata` only propagates user defined `labels` to cluster statefulset but not `annotations`. It built to only include a fixed hash annotation for statefulset.

Would be nice either defining annotation for statefulset or pods template through Cluster spec. One usecase being datadog log collection related config or any additional component to process statefulset resources data.

apedriza:
fix #735 

cluster annotations are propagated to statefulset and pods

statefulset.Annotations = cluster.Annotations + computed hash
pod.Annotations = cluster.Annotations

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
@@ -52,6 +52,7 @@ func (r *ClusterReconciler) findStatefulSetPod(ctx context.Context, statefulSet 
 func (r *ClusterReconciler) generateStatefulSet(kmc *km.Cluster) (apps.StatefulSet, error) {
 
 	labels := labelsForCluster(kmc)
+	annotations := annotationsForCluster(kmc)
 
 	statefulSet := apps.StatefulSet{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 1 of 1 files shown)


internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
```diff
@@ -52,17 +52,17 @@ func (r *ClusterReconciler) findStatefulSetPod(ctx context.Context, statefulSet
...
 			Kind:       "StatefulSet",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        kmc.GetStatefulSetName(),
-			Namespace:   kmc.Namespace,
-			Labels:      labels,
-			Annotations: annotationsForCluster(kmc),
+			Name:      kmc.GetStatefulSetName(),
+			Namespace: kmc.Namespace,
+			Labels:    labels,
```
makhov, line 65: Small nitpick: we can keep setting annotations here and change the hash adding piece to:
```
statefulSet.Annotations[statefulSetAnnotation] = controller.ComputeHash(&statefulSet.Spec.Template, statefulSet.Status.CollisionCount)
```

makhov (approved):

internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
apedriza: I thought about it but as I use the `annotations` variable also for the pod annotations, `statefulset.Annotations` and `pod.Annotations` point to the same reference so if I add a field in `statefulset.Annotations` then I also add it for the pod (I don't think the pod should contain that hash). That's why I opted for this approach. I also wanted to avoid calling `annotationsForCluster()` twice

PR #806, opened 2024-11-11
Add tests for k0scontrolplane controller

apedriza:
### Changes
-  add unit and integration tests on the k0scontrolplane controller. They use envtest to spin-up to `kube-apiserver`+`etcd` without using a cluster. In some of these tests it is enough to use `fakeclient` for the creation of objects and subsequent state check but since for others it is necessary to use envtest (an apiserver is needed) and in order to reduce the different ways of creation of the environment I have chosen to use envtest for any test. For the test implementation I have taken inspiration from how upstream projects do this job. 

    The idea is, after agreeing on how we want this type of testing for the project, to add the rest of the tests for the other controllers.
    
    *More detailed testing of the machine reconciliation logic can be added but since it is undergoing changes I have decided to add only basic testing for the time being.*

- updates docs about testing and moving some line to the proper place


internal/test/envtest/environment.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	options := manager.Options{
+		Scheme: scheme.Scheme,
+		Metrics: metricsserver.Options{
+			BindAddress: "0",
+		},
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
```
apedriza, line 134: If caching is not disabled in these cases, testing might encounter errors due to delays in cache synchronization, as the cache may not immediately reflect the current state in the API server

internal/test/builder/bootstrap.go
```diff
@@ -0,0 +1,104 @@
+/*
```
makhov, line 1: I'm not fully following, are all these files needed for envtest?

apedriza: you are absolutely right. Removed all that unnecessary code

apedriza:
@makhov sorry for the noise. I didnt realize that unit/integration testing was failing due to new updates in machines reconciliation. Ready for review :)

internal/controller/controlplane/helper.go
```diff
@@ -43,14 +43,15 @@ import (
...
+	generatedMachineRoleLabel     = "cluster.x-k8s.io/generateMachine-role"
 )
 
 func (c *K0sController) createMachine(ctx context.Context, name string, cluster *clusterv1.Cluster, kcp *cpv1beta1.K0sControlPlane, infraRef corev1.ObjectReference, failureDomain *string) (*clusterv1.Machine, error) {
 	machine, err := c.generateMachine(ctx, name, cluster, kcp, infraRef, failureDomain)
 	if err != nil {
 		return nil, fmt.Errorf("error generating machine: %w", err)
 	}
-	_ = ctrl.SetControllerReference(kcp, machine, c.Scheme)
+	_ = ctrl.SetControllerReference(kcp, machine, c.Client.Scheme())
```
makhov, line 54: Out of curiosity, why do we change it?

apedriza: writing tests for the controller I saw that we access the schema at this point and I thought about refactoring it using the client schema directly and remove `K0scontroller.Scheme` field. Both are obtained from `manager.Manager` so I thought it was safe to make this change

internal/controller/controlplane/helper.go
```diff
@@ -210,7 +211,11 @@ func (c *K0sController) generateMachineFromTemplate(ctx context.Context, name st
 		return nil, err
 	}
 
-	_ = ctrl.SetControllerReference(kcp, unstructuredMachineTemplate, c.Scheme)
+	_ = ctrl.SetControllerReference(cluster, unstructuredMachineTemplate, c.Client.Scheme())
+	err = c.Client.Patch(ctx, unstructuredMachineTemplate, client.Merge, &client.PatchOptions{FieldManager: "k0smotron"})
```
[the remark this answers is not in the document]
apedriza: currently we patch InfraMachine but not InfraMachineTemplate. The current SetControllerReference in (Infra)MachineTemplate is lost. If I dont patch this ownership, test will fail when checking ownership:

```
g.Expect(gmt.GetOwnerReferences()).To(ContainElement(metav1.OwnerReference{
		APIVersion:         clusterv1.GroupVersion.String(),
		Kind:               "Cluster",
		Name:               cluster.Name,
		Controller:         ptr.To(true),
		BlockOwnerDeletion: ptr.To(true),
		UID:                cluster.UID,
	}))
```

apedriza:
Ready for review, also I added some docs changes about testing

docs/contributing/contribute-testing.md
```diff
@@ -1,70 +1,22 @@
...
+- **Unit Testing**: Focused on isolated components, unit tests validate the correctness of individual functions or methods in a controlled environment, ensuring they behave as expected.
 
+- **Integration Testing**: For integration testing, k0smotron uses [envtest](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest), which provides a lightweight Kubernetes control plane with the essential components needed to test interactions between the code and Kubernetes API. This approach allows us to validate the behavior of our controllers and other Kubernetes integrations in a reproducible and efficient manner without requiring a full cluster. 
 
-   ```shell
-   go fmt ./...
-   ```
+  Following the best practices suggested in the [Cluster API documentation](https://cluster-api.sigs.k8s.io/developer/core/testing), integration tests are written using [**generic infrastructure providers**](https://cluster-api.sigs.k8s.io/developer/core/testing#generic-providers) rather than a specific provider. This ensures that tests remain agnostic and reusable across different infrastructures, fostering better maintainability and adaptability.
 
-3. Check documentation.
```
apedriza, line 27: I think all this should be in contribute workflow instead testing so I moved it there

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -357,19 +358,27 @@ func (c *K0sController) reconcileMachines(ctx context.Context, cluster *clusterv
...
 		}
 	}
+	desiredMachineNames := make(map[string]bool)
+	for i := range desiredMachineNamesSlice {
+		desiredMachineNames[desiredMachineNamesSlice[i]] = true
+	}
+
+	// if it is necessary to reduce the number of replicas even counting the replicas to be eliminated
+	// because they are outdated, we choose the oldest among the valid ones.
+	if activeMachines.Len() > int(kcp.Spec.Replicas)+len(machineNamesToDelete) && len(desiredMachineNamesSlice) > 0 {
```
makhov, line 379: Could you elaborate a bit on why the logic was changed? The result looks more or less the same

makhov (approved):

internal/controller/controlplane/k0s_controlplane_controller.go
makhov: Great catch!

PR #815, opened 2024-11-15
Remove obsolete statements regarding HA and kine configuration from docs

makhov:
Now HA control plane could run with etcd as well, not only with kine


franpog859 (approved):
Great news!

apedriza (approved):

PR #837, opened 2024-11-28
feat: allow setting annotations all the way down to pods

cwrau:
pull annotations from K0smotronControlPlane all the way through the
k0sCluster to the StatefulSet template

this allows the user to add e.g. `checksum/configmap` on the K0smotronControlPlane to auto-roll the StatefulSet pods

Signed-off-by: Chris Werner Rau <cwrau@cwrau.info>

---

I'm not 100% sure if this is the preferred way, as this uses the "global" `annotationsForCluster` function. But this way there is at least no API change 😅

### DIFF
--- a/api/k0smotron.io/v1beta1/k0smotroncluster_types.go
+++ b/api/k0smotron.io/v1beta1/k0smotroncluster_types.go
@@ -354,6 +354,10 @@ func (kmc *Cluster) GetConfigMapName() string {
 	return kmc.getObjectName("kmc-%s-config")
 }
 
+func (kmc *Cluster) GetConfigMapChecksumAnnotationName() string {
+	return "checksum/configmap"
+}
+
 func (kmc *Cluster) GetServiceName() string {
 	switch kmc.Spec.Service.Type {
 	case v1.ServiceTypeNodePort:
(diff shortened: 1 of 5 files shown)


makhov:
I'm sorry, forgot about this one. @cwrau could you resolve conflicts, pls?

cwrau:
Sorry, I missed the failed build 😅

apedriza:
Annotations propagated to statefulset seems to be _duplicated_. 

```yaml
Pod Template:
  Labels:       app=k0smotron
                cluster=docker-test
                cluster.x-k8s.io/cluster-name=docker-test
                cluster.x-k8s.io/control-plane=true
                component=cluster
  Annotations:  checksum: 82bc21341125c90173dd9fe332ef810708f119f8f55078dee23d10be2502d109
                checksum/configmap: 82bc21341125c90173dd9fe332ef810708f119f8f55078dee23d10be2502d109
```

This is because I think because in some points we are modifying the returned value of `AnnotationsForK0smotronCluster` func, which is a reference to `cluster.Annotation` not a copy. To avoid unexpected behaviors like these duplicate annotations (they don't have exactly the same key but express the same thing) I would say to always use the same key for both configmap and sts annotations even if it is a bit redundant the name `checksum/configmap` as annotation in the configmap. This way we will not duplicate the propagated values and seems simpler

apedriza:
@cwrau main had a test bug that got carried over to your changes after the rebase. The fix is already in main, so you just need to rebase again. Sorry for the inconvenience! 😄

cwrau:
We've since switched to rolling our own control-plane controller(https://github.com/teutonet/cluster-api-provider-hosted-control-plane), therefore this is not needed for us anymore.

If someone else wants this change, I can still rebase it or whatever is needed. Or you can just take my code, no problem.

apedriza:
this should be doable by using [patches feature](https://github.com/k0sproject/k0smotron/pull/1341)

PR #840, opened 2024-12-02
K0s ClusterID added to annotations

### LINKED ISSUE #683: Expose k0s cluster ID in kosmotron API
K0s is configured to collect telemetry data from clusters. It collects `cluster_id` as the UID of the `kube-system` namespace in the format of `kube-system:5aed46fc-9567-4f77-b67b-d0139653cc48`. See: https://github.com/k0sproject/k0s/blob/761867a95cf089eda128f1b35a23a780a527f8f8/pkg/telemetry/telemetry.go#L70

Project 2A also needs to send telemetry and we need the k0s `cluster_id` to be exposed in some of the k0smotron objects to send project 2A-specific metrics and map them with k0s metrics by `cluster_id`.

makhov:
Fixes #683

### DIFF
--- a/api/controlplane/v1beta1/groupversion_info.go
+++ b/api/controlplane/v1beta1/groupversion_info.go
@@ -34,3 +34,5 @@ var (
 	// AddToScheme adds the types in this group-version to the given scheme.
 	AddToScheme = SchemeBuilder.AddToScheme
 )
+
+const K0sClusterIDAnnotation = "k0sproject.io/cluster-id"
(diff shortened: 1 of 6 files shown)


makhov:
Yeah, good point. 

@eromanova is it ok, if we add clusterID as annotation to `K0smotronControlPlane` and `K0sControlPlane` objects? 

@apedriza or should we add the annotation to the `Cluster` object itself?

eromanova:
@makhov  Yes, it’s fine if the `clusterID` is stored in the annotation. What matters is that we can retrieve it from the public API objects, regardless of where the field is located.

apedriza:
> @apedriza or should we add the annotation to the `Cluster` object itself?

Yes, good idea 👍

apedriza (approved):

PR #844, opened 2024-12-04
Remove etcdmember that left the cluster

makhov:
Delete old EtcdMember from the cluster after leaving


apedriza:
I'm a little confused here. Is this change because k0s does soft delete on etcdmembers + we reuse the node names or because of the https://github.com/k0sproject/k0s/issues/5161 bug? If it is the latter could we add some kind of TODO commenting the temporary fix?

makhov:
Yes, it's because k0s does soft delete and name reuse, but also I just think it's better to clean outdated objects from the cluster in general.

apedriza (approved):

PR #845, opened 2024-12-04
Add controlplane upgrade e2e test using capi framework

### LINKED ISSUE #775: Refactor smoke tests
We need to refactor some/most of the smokes.

**Duplication**

There's lot of code duplication for some test lifecycle stuff like applying the CAPI yamls etc. We should really refactor all these out to common place.

**Hard to debug on CI**

Currently, when a smoke fails on CI, it's pretty impossible to say what went sideways. Yes, we do collect k0smotron controller logs but often times that does not really tell much. What we would need is also some state of _thing_, like a full dump of all the related objects. And possibly logs of other controllers. What I think in the end we'd need is basically a [support dump](https://github.com/k0sproject/k0s/blob/main/inttest/common/bootloosesuite.go#L359) like we do on k0s side for the smokes.

**Move some of the standaloone k0smotron tests away from k0s bootloose suite**

We don't really need that functionality, we should be able to run all the tests in similar way on top of kind.

apedriza:
fix #775 

This PR makes use of the [e2e framework](https://pkg.go.dev/sigs.k8s.io/cluster-api/test/framework) built by CAPI and used by several providers. This is intended to address current problems described in this [issue](https://github.com/k0sproject/k0smotron/issues/775) such as code repetition or a proper collection of logs in case of failure. Also, this approach prepares the ground for testing using an agnostic infrastructure provider. By adopting this framework, much of the code required for each test (such as bootstrap management cluster setup) is imported from the CAPI framework. On the other hand, a log dump of all relevant test components logs + configuration used is implemented if the test fails, improving the debugging capability of the test in case of failure. 

For now, a basic configuration is sufficient. The intention is to migrate the current CAPI smoke tests little by little from the current approach to the proposed one.


makhov:
I'm personally not a big fan of Ginkgo (if not to say more), could we use the framework without it?

.github/workflows/go.yml
```diff
@@ -171,3 +171,31 @@ jobs:
     uses: ./.github/workflows/capi-smoke-tests.yml
     with:
       smoke-suite: ${{ matrix.smoke-suite }}
+  e2e-migration:
+    name: E2E migration (WIP)
```
jnummelin, line 175: WIP still?

e2e/util/cleanup.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return nil
+}
+
+// waitForClusterDeleted waits until the cluster object has been deleted.
+func waitForClusterDeleted(ctx context.Context, input capiframework.WaitForClusterDeletedInput, interval Interval) {
+	fmt.Printf("Waiting for cluster %s to be deleted", klog.KObj(input.Cluster))
+
+	clusterName := input.Cluster.GetName()
+	clusgerNamespace := input.Cluster.GetNamespace()
```
jnummelin, line 57:
```suggestion
	clusterNamespace := input.Cluster.GetNamespace()
```

e2e/workload_cluster_upgrade_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// Setup a Namespace where to host objects for this spec and create a watcher for the namespace events.
+	namespace, _ := util.SetupSpecNamespace(ctx, testName, managementClusterProxy, artifactFolder)
+
+	clusterName := fmt.Sprintf("%s-%s", testName, capiutil.RandomString(6))
+
+	workloadClusterTemplate := clusterctl.ConfigCluster(ctx, clusterctl.ConfigClusterInput{
+		ClusterctlConfigPath: clusterctlConfigPath,
+		KubeconfigPath:       managementClusterProxy.GetKubeconfigPath(),
+		// select cluster templates
+		Flavor: "ooc",
```
jnummelin, line 59: I guess this maps to the yaml filename, right? Bit more descriptive names would make life easier 😄

jnummelin (changes requested):
Added couple questions to clarify things

Couple of small fixes IMO needed, looks pretty good in general

.golangci.yml
```diff
@@ -17,4 +17,4 @@ linters:
     - gofmt        # Checks whether code was gofmt-ed
     - goheader     # Checks is file headers matche a given pattern
     - revive       # Stricter drop-in replacement for golint
-    - ginkgolinter # Enforces standards of using ginkgo and gomega
+    - ginkgolinter # Enforces standards of using ginkgo and gomega
```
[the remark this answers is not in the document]
apedriza: mm I don't know why this diff appears, at the beginning I added some linter-ignoring stuff when ginkgo was added but then I removed it after agree on remove ginkgo. Anyway `ginkgolinter` was added some time ago probably for this usage: https://github.com/k0sproject/k0smotron/blob/main/internal/controller/k0smotron.io/suite_test.go 👈  I think is scaffolded code from kubebuilder. Doesn't add any value so I think could be removed

Done in 016c67a2c3271953cdaa7ba6d0d0327f202ce868

.github/workflows/go.yml
apedriza: removed! 502a3701fdc83b2e67c61eac82ddc2ce3f32a67c

e2e/data/infrastructure-docker/cluster-template-ooc.yaml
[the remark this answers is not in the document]
apedriza: is a cluster predefined configuration template that can be used by more than one test scenario (using docker infra provider). It is what CAPI refers as "flavor". Another custer-template for some of our tests should be a definition using clusterclass for example. In this case this cluster-template flavor uses is called "OOC" because controlplane is Out-of-Cluster (k0scontrolplane). I can rename this template to `cluster-template-k0scontrolplane` maybe.

e2e/workload_cluster_upgrade_test.go
apedriza: wdyt about `out-of-cluster` meaning it uses k0scontrolplane in the cluster template? I could add some comments in the template header too.

e2e/util/controlplane.go
```diff
@@ -0,0 +1,260 @@
...
+	// TODO: avoid check node conditions because "NodeHealthy" is "False" due to
+	// NodeMemoryPressure and NodePIDPressure is False
+
+	// Logf("Waiting for control-plane machines to have the upgraded kubernetes version")
+	//capiframework.WaitForControlPlaneMachinesToBeUpgraded(ctx, capiframework.WaitForControlPlaneMachinesToBeUpgradedInput{
+	// 	Lister:                   mgmtClient,
+	// 	Cluster:                  input.Cluster,
+	// 	MachineCount:             int(input.ControlPlane.Spec.Replicas),
+	// 	KubernetesUpgradeVersion: input.KubernetesUpgradeVersion,
+	// }, "10m")
```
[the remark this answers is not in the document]
apedriza: done! 6e9350ce251da179469e0723d3193ee52a662069

apedriza:
@jnummelin thanks for the feedback! Comments addressed. Some of them accidentally lost but addressed too 9c9b3c0b201b7474177d1db4725c03f521d88bdc

e2e/data/infrastructure-docker/cluster-template-ooc.yaml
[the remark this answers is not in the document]
apedriza: 9c418b47a96a87da8ca576096faade88b560a324

PR #878, opened 2025-01-07
Remove outdated info about the persistence

makhov:


apedriza (approved):

tmsdce:
Hi @makhov 
Does it still hold true when using cluster-api in HCP mode ?
Is it really required as stated in https://github.com/k0smotron/k0smotron/blob/main/docs/update/update-cluster-pod.md?plain=1#L39 ?

makhov:
No, it's also outdated info and should be removed from the docs. Thanks for pointing out!

PR #886, opened 2025-01-15
Fix k0smotronControlplane status report for updatedReplicas field

apedriza:
a conversion is necessary to compare the versions since the format reported by the `k0s version` command and the one stored in the registry differ


apedriza:
@makhov `v1.4.0-rc.0` includes the new feature for k0smotroncontrolplane status reporting. Worth bump v1.4.0 RC with this fix?

PR #887, opened 2025-01-15
Fix update process for older k0s versions

makhov:
We remove controlnode from the cluster too early and it leads to the node not leaving the cluster.

So I removed this step completely from the deletion process and remove all controlnodes that don't have a corresponding machine

### DIFF
--- a/internal/controller/controlplane/k0s_controlplane_controller.go
+++ b/internal/controller/controlplane/k0s_controlplane_controller.go
@@ -373,6 +373,13 @@ func (c *K0sController) reconcileMachines(ctx context.Context, cluster *clusterv
 	}
 	log.Log.Info("Collected machines", "count", machines.Len(), "desired", kcp.Spec.Replicas, "updating", clusterIsUpdating, "deleting", len(machineNamesToDelete), "desiredMachines", desiredMachineNames)
 
+	go func() {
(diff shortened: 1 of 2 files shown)


internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -373,6 +373,13 @@ func (c *K0sController) reconcileMachines(ctx context.Context, cluster *clusterv
 	}
 	log.Log.Info("Collected machines", "count", machines.Len(), "desired", kcp.Spec.Replicas, "updating", clusterIsUpdating, "deleting", len(machineNamesToDelete), "desiredMachines", desiredMachineNames)
 
+	go func() {
```
apedriza, line 376: I think if a machine is deleted and this goroutine finish before machine deletion is run, controlnode related will no be removed until next reconciliation is needed which may leave us in a state where the machine does not exist but the controlnode does. Could we force a new reconciliation or similar to force run controlnode deletion logic if machine A is removed in order to avoid this race condition? maybe a go routine after machine deletion checking node leaves and if so removing the controlnode

makhov: We update the status which triggers the reconciliation, so I get the actual list of nodes in my tests

apedriza (approved):

PR #889, opened 2025-01-16
Treat missing machines as unavailable

makhov:
Currently, if there are only two machines in a cluster with 3 replicas, K0sControlPlane will report 2 nodes available and 0 unavailable 

With this change it will report the missing one as unavailable

### DIFF
--- a/internal/controller/controlplane/status.go
+++ b/internal/controller/controlplane/status.go
@@ -106,6 +106,11 @@ func computeStatus(machines collections.Machines, kcp *cpv1beta1.K0sControlPlane
 		}
 	}
 
+	// If some machines are missing, count them as unavailable
+	if int(kcp.Spec.Replicas) > machines.Len() {
+		unavailableReplicas += int(kcp.Spec.Replicas) - machines.Len()
+	}
+
 	kcp.Status.ReadyReplicas = int32(readyReplicas)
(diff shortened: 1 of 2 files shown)


makhov:
@jnummelin since you were diving into this, could you confirm, that my assumption is correct?

jnummelin:
@makhov I think your assumption is correct, from CAPI [docs](https://cluster-api.sigs.k8s.io/developer/providers/contracts/control-plane#controlplane-replicas):
```
    // unavailableReplicas is the total number of unavailable machines targeted by this control plane.
    // This is the total number of machines that are still required for the deployment to have 100% available capacity. 
    // They may either be machines that are running but not yet ready or machines
    // that still have not been created.
    // +optional
    UnavailableReplicas int32 `json:"unavailableReplicas"`
```

jnummelin (commented):
I think we should have a testcase for this scenario

makhov:
Yeah, added the test case

apedriza (approved):
LGTM

as a side note, I've been testing how kuebadm reports unavailable replicas and seems it only takes into account a replica as unavailable **if it is created** + other conditions which makes replicas unavailable, which doesnt fit with what docs states I think:
> They may either be machines that are running but not yet ready or machines that still have not been created.

PR #891, opened 2025-01-17
Improve unhealthy machine recreation process

makhov:
Currently, the remediation process will delete the machine, but in the same reconciliation cycle, we will recreate it again. It can lead to a infinite loop


internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -438,13 +437,19 @@ func (c *K0sController) reconcileMachines(ctx context.Context, cluster *clusterv
 		name := machineName(kcp, machineNamesToDelete, desiredMachineNames)
 		log.Log.Info("desire machine", "name", len(desiredMachineNames))
 
+		for _, mn := range deletedMachines.Names() {
```
makhov, line 440: This is the actual change

juanluisvaladas (approved):

PR #902, opened 2025-01-28
Experimental Helm chart for k0smotron

### LINKED ISSUE #122: Helm chart for installing k0smotron

makhov:
Fixes #122

Currently, we need to generate it manually for each release, going to improve it later


apedriza:
should controller be installed in k0smotron namespace by default as current installation does?

apedriza:
Probably Im doing something wrong but `make helm` + `helm install ..` fails because `no template "{{ .Release.Namespace }}.fullname" associated with template "gotpl"`. Should not be `k0smotron.fullname` instead? @makhov 

Also, do you think we should add new lines about this new installation method in the docs?

makhov:
Closing in favor of #1475

PR #904, opened 2025-01-29
Stale bot GHA

makhov:


.github/workflows/stale.yml
```diff
@@ -0,0 +1,19 @@
...
+      - uses: actions/stale@v9
+        with:
+          stale-issue-message: 'The issue is marked as stale since no activity has been recorded in 30 days'
+          stale-pr-message: 'The PR is marked as stale since no activity has been recorded in 30 days'
+          days-before-stale: 30
+          close-pr-label: "stale"
+          remove-issue-stale-when-updated: true
+          remove-pr-stale-when-updated: true
+          labels-to-remove-when-unstale: "stale"
+          exempt-issue-labels: "enhancement"
```
apedriza, line 19:
```suggestion
          exempt-issue-labels: "enhancement"
          exempt-pr-labels: "enhancement"
```
same for PRs?

apedriza: nit: maybe some label with a general name could fit better than `enhancement` in this case. For example `keep` if we just want to avoid stale workflow

makhov: Fixed, thanks!

apedriza (approved):

PR #908, opened 2025-01-30
Fix entrypoint script

makhov:
Noticed weird error running k0smotron cluster on newer k0s version. Haven't found the root cause yet, but this change fixes it
```
/entrypoint.sh: executing /k0smotron-entrypoint.sh
env: can't execute '/k0smotron-entrypoint.sh': Exec format error
```

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
@@ -118,7 +118,7 @@ func (r *ClusterReconciler) generateStatefulSet(kmc *km.Cluster) (apps.StatefulS
 						Name:            "controller",
 						Image:           kmc.Spec.GetImage(),
 						ImagePullPolicy: v1.PullIfNotPresent,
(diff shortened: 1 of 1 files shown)


apedriza (approved):

internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
```diff
@@ -118,7 +118,7 @@ func (r *ClusterReconciler) generateStatefulSet(kmc *km.Cluster) (apps.StatefulS
 						Name:            "controller",
 						Image:           kmc.Spec.GetImage(),
 						ImagePullPolicy: v1.PullIfNotPresent,
-						Args:            []string{"/k0smotron-entrypoint.sh"},
+						Args:            []string{"/bin/sh", "-c", "/k0smotron-entrypoint.sh"},
```
twz123, line 121: Would have `[]string{"/bin/sh", "/k0smotron-entrypoint.sh"}` (without the -c) fixed the problem as well? I really don't have a clue what the root cause may be :thinking:

PR #919, opened 2025-02-12
Remove unneeded status update since we now update the status via defe…

jnummelin:
…r func

This puts the controller into update failing "loop" since we now update the status twice from same reconciliation round.


apedriza:
I think the proposed change is correct since the status update is done in the defer but I have tested and still encounter the problem sometimes. I think this issue is similar to https://github.com/k0sproject/k0smotron/issues/921. I have tried using a `client.MergeFrom` in the patch operation and I am not encountering the problem again. 

As a side note, in other patching operations for other resources we are also encountering this problem. I have not seen it in depth but maybe we should take a look at using https://pkg.go.dev/sigs.k8s.io/cluster-api@v1.8.5/util/patch in the future to handle these updates.

makhov (approved):

makhov:
@apedriza yeah. I think we can merge this one and I'll check all the merges in my fix for #921

PR #925, opened 2025-02-18
Adding topologySpreadConstraints to HCP clusters

makhov:
Topology Spread Constraints adds more control how  pods are spread across your cluster

### DIFF
--- a/api/k0smotron.io/v1beta1/k0smotroncluster_types.go
+++ b/api/k0smotron.io/v1beta1/k0smotroncluster_types.go
@@ -101,6 +101,9 @@ type ClusterSpec struct {
 	//+kubebuilder:default={"image":"quay.io/k0sproject/etcd:v3.5.13","persistence":{}}
 	Etcd EtcdSpec `json:"etcd,omitempty"`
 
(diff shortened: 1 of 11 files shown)


api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -101,6 +101,8 @@ type ClusterSpec struct {
 	//+kubebuilder:default={"image":"quay.io/k0sproject/etcd:v3.5.13","persistence":{}}
 	Etcd EtcdSpec `json:"etcd,omitempty"`
 
+	// TopologySpreadConstraints will be passed directly to BOTH etcd and .
```
jnummelin, line 104: and what? looks like some words missing :D

jnummelin (changes requested):
The API comment seems to be missing something 😄

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
makhov: And it's classified :)  
Yeah, nice catch, fixed!

jnummelin (approved):

PR #932, opened 2025-02-20
Change all imports to the new repo url

makhov:
Since we moved the repo, let's rename all imports and links


apedriza:
does lines like https://github.com/k0smotron/k0smotron/blob/main/.github/workflows/publish-docs.yml#L69 in CI need to be changed?

makhov:
Yeah, great catch, fixed!

apedriza (approved):

PR #944, opened 2025-03-04
Fix conflicts updating resources

### LINKED ISSUE #937: conflicts when patching a resource
### What happened?

Occasionally, resources are tried to be updated but this is not possible because they have been modified by another entity (CAPI). This can be avoided by using MergeFrom in the patch to calculate the diff.

### Involved Areas

_No response_

No conflicts updating resources.

### Steps to reproduce

-

### k0smotron version

 1.4.0

### k0s version

any

releted to https://github.com/k0smotron/k0smotron/issues/921

apedriza:
fix #937 

Change patch followed strategy to using https://pkg.go.dev/sigs.k8s.io/cluster-api@v1.8.5/util/patch. This packages uses `client.MergeFrom` option to patch the resource using the diff between updated one and the original. This way we prevent conflicts if other controller are updating the resource concurrently. By adopting this new patch strategy, some bugs have appeared when updating resources. For example, not taking into account the previous status when filing the resource. This was not a problem because only the subresource `.status` was updated but at runtime the resource could have duplicate values (see diff comments).

New approach it is only change in those resources which can potentially face a conflict when they are updated by k0smotron and CAPI.

### DIFF
--- a/api/controlplane/v1beta1/k0s_types.go
+++ b/api/controlplane/v1beta1/k0s_types.go
@@ -127,15 +127,55 @@ type K0sControlPlaneList struct {
 
 type K0sControlPlaneStatus struct {
 	// Ready denotes that the control plane is ready
-	Ready                       bool   `json:"ready"`
-	Inititalized                bool   `json:"initialized"`
-	ExternalManagedControlPlane bool   `json:"externalManagedControlPlane"`
-	Replicas                    int32  `json:"replicas"`
-	Version                     string `json:"version"`
-	Selector                    string `json:"selector"`
-	UnavailableReplicas         int32  `json:"unavailableReplicas"`
-	ReadyReplicas               int32  `json:"readyReplicas"`
-	UpdatedReplicas             int32  `json:"updatedReplicas"`
+	// +optional
+	Ready bool `json:"ready"`
+
+	// initialized denotes that the KubeadmControlPlane API Server is initialized and thus
+	// it can accept requests.
+	// NOTE: this field is part of the Cluster API contract and it is used to orchestrate provisioning.
+	// The value of this field is never updated after provisioning is completed. Please use conditions
+	// to check the operational state of the control plane.
+	// +optional
+	Inititalized bool `json:"initialized"`
+
+	// externalManagedControlPlane is a bool that should be set to true if the Node objects do not exist in the cluster.
+	// +optional
+	ExternalManagedControlPlane bool `json:"externalManagedControlPlane"`
+
+	// replicas is the total number of non-terminated machines targeted by this control plane
+	// (their labels match the selector).
+	// +optional
+	Replicas int32 `json:"replicas"`
+
+	// version represents the minimum Kubernetes version for the control plane machines
+	// in the cluster.
+	// +optional
+	Version string `json:"version"`
+
+	// selector is the label selector in string format to avoid introspection
+	// by clients, and is used to provide the CRD-based integration for the
+	// scale subresource and additional integrations for things like kubectl
+	// describe.. The string will be in the same format as the query-param syntax.
+	// More info about label selectors: http://kubernetes.io/docs/user-guide/labels#label-selectors
+	// +optional
+	Selector string `json:"selector"`
+
+	// unavailableReplicas is the total number of unavailable machines targeted by this control plane.
+	// This is the total number of machines that are still required for
+	// the deployment to have 100% available capacity. They may either
+	// be machines that are running but not yet ready or machines
+	// that still have not been created.
+	// +optional
+	UnavailableReplicas int32 `json:"unavailableReplicas"`
+
+	// readyReplicas is the total number of fully running and ready control plane machines.
+	// +optional
+	ReadyReplicas int32 `json:"readyReplicas"`
+
+	// updatedReplicas is the total number of non-terminated machines targeted by this control plane
+	// that have the desired template spec.
+	// +optional
+	UpdatedReplicas int32 `json:"updatedReplicas"`
 
 	// Conditions defines current service state of the K0sControlPlane.
 	// +optional
--- a/internal/controller/controlplane/k0smotron_controlplane_controller.go
+++ b/internal/controller/controlplane/k0smotron_controlplane_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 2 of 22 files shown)


api/controlplane/v1beta1/k0s_types.go
```diff
@@ -127,15 +127,55 @@ type K0sControlPlaneList struct {
...
+	UnavailableReplicas int32 `json:"unavailableReplicas"`
+
+	// readyReplicas is the total number of fully running and ready control plane machines.
+	// +optional
+	ReadyReplicas int32 `json:"readyReplicas"`
+
+	// updatedReplicas is the total number of non-terminated machines targeted by this control plane
+	// that have the desired template spec.
+	// +optional
+	UpdatedReplicas int32 `json:"updatedReplicas"`
```
apedriza, line 178: required mark them as `optional` for a correct patch

internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -221,10 +232,7 @@ func (c *K0smotronController) waitExternalAddress(ctx context.Context, cluster *
 			log.Info("Updated infrastructure cluster", "host", host, "port", port)
 			cluster.Spec.ControlPlaneEndpoint.Host = host
 			cluster.Spec.ControlPlaneEndpoint.Port = int32(port)
-			if err := c.Client.Update(ctx, cluster); err != nil {
-				log.Error(err, "Failed to update Cluster endpoint")
-				return false, err
-			}
```
apedriza, line 227: `cluster` is updated in the _defer_ func in `Reconcile` loop. Removing resources updates within the recocile loop as much as possible.

internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -295,35 +304,27 @@ func (c *K0smotronController) reconcile(ctx context.Context, cluster *clusterv1.
...
-		if kcp.Spec.ExternalAddress == "" {
-			kcp.Spec.ExternalAddress = foundCluster.Spec.ExternalAddress
-		}
-		if !reflect.DeepEqual(foundCluster.Spec, kcp.Spec) {
-			err := c.Client.Patch(ctx, &kcluster, client.Apply, &client.PatchOptions{
-				FieldManager: "k0smotron",
-			})
 
-			return ctrl.Result{}, false, err
-		}
```
apedriza, line 313: I think is not needed because `kcluster` has no changes since is retrieved couple lines above.

internal/controller/k0smotron.io/k0smotroncluster_configmap.go
```diff
@@ -60,7 +60,7 @@ func (r *ClusterReconciler) generateConfig(kmc *km.Cluster, sans []string) (v1.C
 		case "k0s.k0sproject.io/v1beta1":
 			existingSANs, found, err := unstructured.NestedStringSlice(unstructuredConfig, "spec", "api", "sans")
 			if err == nil && found {
-				sans = append(sans, existingSANs...)
+				sans = kcontrollerutil.AddToExistingSans(existingSANs, sans)
```
apedriza, line 63: avoiding duplicates values here too.

internal/controller/k0smotron.io/k0smotroncluster_configmap.go
```diff
@@ -247,7 +251,7 @@ func getV1Beta1Spec(kmc *km.Cluster, sans []string) map[string]interface{} {
 			"type": "etcd",
 			"etcd": map[string]interface{}{
 				"externalCluster": map[string]interface{}{
-					"endpoints":      []string{fmt.Sprintf("https://%s:2379", kmc.GetEtcdServiceName())},
+					"endpoints":      []interface{}{fmt.Sprintf("https://%s:2379", kmc.GetEtcdServiceName())},
```
apedriza, line 254: `Unstructured.DeepCopy` neither support `int` not `[]string` so converting them to supported types. More info here: https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/converter.go#L636. This is hit by the new _patch_ package introduced.

internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -179,64 +192,52 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		if !isAPIServerEtcdClientCertRef {
+			logger.Info("Reconciling etcd certs")
+			err := r.ensureEtcdCertificates(ctx, kmc)
+			if err != nil {
+				return ctrl.Result{Requeue: true, RequeueAfter: time.Minute}, fmt.Errorf("error generating etcd certificates: %w", err)
+			}
+			kmc.Spec.CertificateRefs = append(kmc.Spec.CertificateRefs, km.CertificateRef{
+				Type: string(secret.APIServerEtcdClient),
+				Name: secret.Name(kmc.Name, secret.APIServerEtcdClient),
+			})
```
apedriza, line 211: this is to avoid values duplicated. In the same loop certificates were propagated to statufulset correctly and after that `.spec` was not updated (which I think i should be updated) only `.status` was updated

internal/controller/k0smotron.io/k0smotroncluster_service.go
```diff
@@ -178,9 +178,6 @@ func (r *ClusterReconciler) reconcileServices(ctx context.Context, kmc km.Cluste
 			return err
 		}
 		kmc.Spec.ExternalAddress = util.FindNodeAddress(nodes)
-		if err := r.Client.Update(ctx, &kmc); err != nil {
-			return err
-		}
```
apedriza, line 183: updated in main _defer_

makhov (approved):
Just a small comment

internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -137,8 +149,12 @@ func (c *K0smotronController) Reconcile(ctx context.Context, req ctrl.Request) (
 		log.Error(err, "Reconciliation failed")
 		return res, err
 	}
+	if !res.IsZero() {
```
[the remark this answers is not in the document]
apedriza: Great, added a comment about requeue when k0smotron cluster just been created. https://github.com/k0smotron/k0smotron/compare/28b437b059ba652eba3be9cf1f0bb0030d53e3b5..f308dbca07f8f4cedd819515b82a89d1d73bdf7d

PR #953, opened 2025-03-12
 HPA support for hosted control planes.

### LINKED ISSUE #588: HPA/VPA for Hosted Control Planes
It would be nice to have autoscaling capabilities for HCPs. The standard Kubernetes HPA/VPA might be overridden on the next reconciliation.

makhov:
Signed-off-by: Alexey Makhov <amakhov@mirantis.com>

Fixes #588 

HPA support for HCP. 

**Note**: it never scales etcd statefulset down, only up to avoid issues with etcd cluster

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_etcd.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_etcd.go
@@ -188,20 +188,23 @@ func (r *ClusterReconciler) reconcileEtcdDefragJob(ctx context.Context, kmc *km.
 }
 
 func (r *ClusterReconciler) reconcileEtcdStatefulSet(ctx context.Context, kmc *km.Cluster) error {
-	desiredReplicas := calculateDesiredReplicas(kmc)
-
 	foundStatefulSet, err := r.ClientSet.AppsV1().StatefulSets(kmc.Namespace).Get(ctx, kmc.GetEtcdStatefulSetName(), metav1.GetOptions{})
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
-	} else {
+	}
+	desiredReplicas := calculateDesiredReplicas(kmc, foundStatefulSet)
+	// We can't check for nil due to the client_go implementation
+	if foundStatefulSet.GetName() != "" {
+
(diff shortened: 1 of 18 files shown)


internal/controller/k0smotron.io/k0smotroncluster_etcd.go
```diff
@@ -188,20 +188,21 @@ func (r *ClusterReconciler) reconcileEtcdDefragJob(ctx context.Context, kmc *km.
...
-	desiredReplicas := calculateDesiredReplicas(kmc)
-
 	foundStatefulSet, err := r.ClientSet.AppsV1().StatefulSets(kmc.Namespace).Get(ctx, kmc.GetEtcdStatefulSetName(), metav1.GetOptions{})
+	desiredReplicas := calculateDesiredReplicas(kmc, foundStatefulSet)
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
 	} else {
+
```
apedriza, line 198: I don't know if I'm missing something but wdyt about this change for better readability?

```go
	foundStatefulSet, err := r.ClientSet.AppsV1().StatefulSets(kmc.Namespace).Get(ctx, kmc.GetEtcdStatefulSetName(), metav1.GetOptions{})
	if err != nil {
		if !apierrors.IsNotFound(err) {
			return err
		}
	}
	desiredReplicas := calculateDesiredReplicas(kmc, foundStatefulSet)
	
	if foundStatefulSet != nil {
		// If we want to scale up existing etcd statefulset, we always scale up by 1 replica at a time and wait for the previous member to be ready
		// This is to avoid the situation where the new member is not able to join the cluster because the previous member is not ready
		if desiredReplicas > *foundStatefulSet.Spec.Replicas {
			// Scale up by 1 replica
			desiredReplicas = int32(*foundStatefulSet.Spec.Replicas) + 1
		}

		if desiredReplicas > foundStatefulSet.Status.ReadyReplicas+1 {
			// Wait for the previous member to be ready. For example, if the desired replicas is 3, we need to wait for the 2nd member to be ready before adding the 3rd member
			return fmt.Errorf("waiting for previous etcd member to be ready")
		}
	}
```

makhov: Done

apedriza (approved):

PR #960, opened 2025-03-19
Disable cache for ConfigMap and Secret

### LINKED ISSUE #957: K0smotron controller is caching resources without filtering
### What happened?

K0smotron caches secrets, services, PVC, etc., without any filtering. This causes constant memory growth when you have a big cluster with lots of secrets. It is noticeable when you are using big Helm Charts, in which templating results are stored in secrets for every revision.

### Involved Areas

Standalone

K0smotron should cache only the resources it creates and manages, not all. It can be achieved with ByObject option from controller runtime cache implementation - https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/cache/cache.go#L244

### Steps to reproduce

Create many big secrets on the cluster and monitor K0smotron controller manager pod memory usage

### k0smotron version

1.4.2

### k0s version

v1.30.4+k0s.0

_No response_

apedriza:
fix #957 

this way we reduce memory footprint. Cluster could have several number of configmaps/secrets, e.g, secrets created for each helm revision as reported in the original issue.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os"
 
+	corev1 "k8s.io/api/core/v1"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/tools/clientcmd"
 
@@ -40,6 +41,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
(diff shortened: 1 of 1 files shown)


jnummelin (approved):

cmd/main.go
```diff
@@ -149,6 +151,15 @@ func main() {
 		// if you are doing or is intended to do any operation such as perform cleanups
 		// after the manager stops then its usage might be unsafe.
 		// LeaderElectionReleaseOnCancel: true,
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{
```
makhov, line 156: Have you checked the possibility to limit cache by the cluster label? Does it work better/worse?

hudymi: Disabling the cache will make a bigger load on the k8s API server. So, the preferred option should be limiting by labels what is cached.

apedriza: I agree with caching according to that tag but that leads us to the case of having to use an alternative client to retrieve resources that do not have that tag, I think the client does not fallback to the api directly from what I understand. That would imply maybe a major change that's why I opted for this option which on the other hand would increase the calls to the api (I don't think significantly).

@makhov wdyt about that change?

apedriza: yes they use this configuration but this implies created an specific client which uses cached secrets as comment states:
> The cached secrets will only be used by the secretCachingClient we create below.

using client directly from manager (as we do now) will do a live lookup as secrets and configmaps are disabled. If we want to just use the cache for secrets with cluster labels we shouldn't combine it with the `DisableFor` opt if there are controllers that need read secrets without that label, i.e when we read data from configmaps/secrets refs. I can propose changes using a client caching by label or disabling it by controller case  but the change will be something bigger than what is proposed I believe :sweat_smile:

hudymi:
Hi. Is there any update on this? Is there any ETA for when it will be included in the release? We would appreciate it.

apedriza:
> Hi. Is there any update on this? Is there any ETA for when it will be included in the release? We would appreciate it.

@makhov can you take a look at the last changes? It is approved but as I have subsequently added significant changes I would like a second look

makhov (approved):

PR #986, opened 2025-04-04
Stabilize SANS field order in ClusterConfig

### LINKED ISSUE #985: kube-router repeatedly restarts due to ClusterConfig sans order changes
### What happened?

k0smotron-controller-manager-control-plane frequently rewrites the order of `spec.api.sans` in the `ClusterConfig`, which causes kube-router pods to be continuously recreated (restarted). As a result, kube-router becomes unstable and keeps restarting.

### Involved Areas

Cluster API integration

kube-router should operate stably without constant restarts. I expected that the pods wouldn't be recreated just because the order of `spec.api.sans` changes slightly.

### Steps to reproduce

1. Refer to [k0smotron documentation](https://github.com/k0smotron/k0smotron/blob/main/docs/capi-aws.md) and set up CAPI (Cluster API) + k0smotron on AWS.
2. Configure to use kube-router as the CNI.
3. When managing the cluster configuration, the entry order of `spec.api.sans` in `ClusterConfig` is frequently changed, and each time this happens, the kube-router Pod is redeployed.
4. You can observe the phenomenon of kube-router continuously restarting.

### k0smotron version

v1.4.3

### k0s version

v1.31.2-k0s.0

- There are no differences other than the order of `sans`, and since the content itself is identical, it appears that the change in order is the cause of the issue.

kahirokunn:
## Overview
This PR stabilizes the order of the `api.sans` field in ClusterConfig to ensure consistent output across reconciliations.

fixes #985

## Problem
Currently, the `api.sans` field in ClusterConfig gets generated with an unstable order during each reconciliation. This causes unnecessary diffs to be detected, leading to components like kube-router being unnecessarily restarted, affecting overall system stability.

## Solution
- Sort the SANS list consistently after generation to ensure a stable output order
- Modified two key functions:
  - Updated `AddToExistingSans` to return a sorted slice
  - Updated `genSANs` to ensure the generated SANS are in a consistent order

## Testing
All tests pass with the updated implementation.

This change improves system stability by preventing unnecessary pod restarts caused by non-semantic configuration changes.

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_configmap.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_configmap.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sort"
 	"strings"
 
 	"github.com/imdario/mergo"
@@ -221,6 +222,9 @@ func (r *ClusterReconciler) genSANs(kmc *km.Cluster) ([]string, error) {
 	// Trim the trailing dot from the CNAME
 	trimmedCNAME, _ := strings.CutSuffix(cname, ".")
 	sans = append(sans, trimmedCNAME)
+
+	// Sort the sans to ensure stable output order
+	sort.Strings(sans)
 
 	return sans, nil
 }
--- a/internal/controller/util/util.go
+++ b/internal/controller/util/util.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"context"
+	"sort"
 
 	km "github.com/k0smotron/k0smotron/api/k0smotron.io/v1beta1"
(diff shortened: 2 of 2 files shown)


kahirokunn:
@cgroschupp Hi, You mentioned that you added `slices.Sort(sans)` after line 64, but I couldn’t locate this change in the diff. Could you please clarify where exactly this fix was applied? Thank you very much for your work and for any clarification you can provide. Thank you.

makhov:
Thanks for the PR. Could you resolve conflicts, pls?

kahirokunn:
@makhov It has been fixed.

makhov (approved):

PR #989, opened 2025-04-07
Add clusterprofileref in order to target a cluster for HCP deployment

apedriza:


internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -97,8 +120,48 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		}
+	}()
+
 	if !kmc.ObjectMeta.DeletionTimestamp.IsZero() {
 		logger.Info("Cluster is being deleted")
+		// If controlplanes run in a different cluster, we need to delete the resources associated with
+		// the k0smotron.Cluster. This is because the owner references cannot be used in this case.
+		if !kmcScope.inClusterHCP {
+			var errors []error
+			for _, descendant := range kmcScope.getDescendants(ctx, kmc) {
```
apedriza, line 154: This is what I like the least since ownerreferences cannot be used for a background deletion. Happy to hear other options 😃

internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
```diff
@@ -538,32 +537,33 @@ func (r *ClusterReconciler) addMonitoringStack(kmc *km.Cluster, statefulSet *app
...
 
-		return nil
+	if !isStatefulSetsEqual(&statefulSet, foundStatefulSet) {
+		return scope.client.Patch(ctx, &statefulSet, client.Apply, patchOpts...)
 	}
 
-	return err
+	if foundStatefulSet.Status.ReadyReplicas == 0 {
+		return fmt.Errorf("%w: no replicas ready yet for statefulset '%s' (%d/%d)", ErrNotReady, foundStatefulSet.GetName(), foundStatefulSet.Status.ReadyReplicas, kmc.Spec.Replicas)
+	}
```
apedriza: I think we can set cluster status to ready if there is one replica already running instead of wait for all replicas to be ready. This way It decreases time for deploying worker nodes

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -32,6 +32,10 @@ import (
 
 // ClusterSpec defines the desired state of K0smotronCluster
 type ClusterSpec struct {
+	// HostingClusterKubeconfigRef is the reference to the kubeconfig of the hosting cluster.
+	// This kubeconfig will be used to deploy the k0s control plane.
+	//+kubebuilder:validation:Optional
+	HostingClusterKubeconfigRef *HostingClusterKubeconfigRef `json:"hostingClusterKubeconfigRef,omitempty"`
```
makhov: I'd call it just `kubeconfigRef` or `kubeconfigSecretRef`

apedriza: done!

.golangci.yml
```diff
@@ -8,7 +8,6 @@ run:
 linters:
   default: none 
   enable:
-    - depguard
```
apedriza, line 11: After the recent changes in the golangci-lint configuration, some noisy errors appeared. I think it’s safe to remove the `depguard` linter for now. It deserves some investigation on how to configure it properly in follow-up changes.

apedriza:
This is the contract fulfillment for k0smotron integrating ClusterProfile api https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/4322-cluster-inventory/README.md#secret-format. It is the responsibility of the user to locate, label and store with the proper key value for the kubeconfig secret accordingly to ClusterProfile API. That way we allow user use ClusterProfile API together with k0smotron but we are not strongly coupled to that API

internal/controller/bootstrap/worker_bootstrap_controller.go
```diff
@@ -71,6 +73,11 @@ type Scope struct {
 	Cluster     *clusterv1.Cluster
 }
 
+type hostingClusterScope struct {
```
jnummelin: dunno if this could've been "merged" with the existing scope object? This way we'd be able to pass a single scope object around instead of multiple. Might simplify the code a bit?

internal/controller/k0smotron.io/k0smotroncluster_controller.go
jnummelin: Like we discussed, creating a "dummy" configmap or something acting as the "root" owner for everything _might_ work. Requires some further testing and exploring so could be a separate PR, if that even works. 😂

internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -243,16 +312,100 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		}
+		kmcScope.secretCachingClient = kmcScope.client
+	}
+
+	return kmcScope, nil
+}
+
+// getDescendants returns the list of resources that are associated with the k0smotron.Cluster
+// and that must be deleted when the k0smotron.Cluster is deleted.
+func (scope *kmcScope) getDescendants(ctx context.Context, kmc *km.Cluster) []client.Object {
```
jnummelin, line 354: This could be maybe enhanced to more generic approach, something like:
```
var descendantTypes = []client.Object{&v1.ConfigMapList{}, &v1.ServiceList{}}

func (scope *kmcScope) getDescendants(ctx context.Context, kmc *km.Cluster) ([]client.Object, error) {
	descendants := make([]client.Object, 0)
	lOpt := client.MatchingLabels(kutil.DefaultK0smotronClusterLabels(kmc))
	
	for _, dt := range descendantTypes {
		err := c.List(ctx, dt, lOpt)
		if err != nil {
			return nil, fmt.Errorf("failed to list descendants of type %s K0smotronControlPlane %s/%s: %w", reflect.TypeOf(variable).String(), kcp.Namespace, kcp.Name, err)
		}
		descendants = append(descendants, dt.Items...)
	}

}
```

Could be a follow-up PR too

jnummelin:
@apedriza any plans for docs on this? In a separate PR?

internal/controller/bootstrap/worker_bootstrap_controller.go
apedriza: yes, I think using `scope` struct simplifies the change. I'll change it in this PR, thanks!

internal/controller/k0smotron.io/k0smotroncluster_controller.go
apedriza: since this method is only used for the removal of resources belonging to the cluster and using that proposed dummy object for centralize the deletion is another different approach I will test both solutions is a separate PR

apedriza:
> @apedriza any plans for docs on this? In a separate PR?

@jnummelin I can add docs in this PR. Any ideas about what form it could take? A simple mention in the standalone documentation that this possibility exists?

jnummelin:
> A simple mention in the standalone documentation that this possibility exists?

yeah. plus some docs around the expected format of the secret for the access

Also I _think_ we should add some notes on the RBAC expected for the "user" of the external client kubeconfig?

apedriza:
Added docs and some refactoring for simplicity, as suggested. Other improvements will be addressed in a subsequent PR

kahirokunn:
This is fantastic—exactly what I hoped for. Thank you so much!

PR #999, opened 2025-04-22
Fix k0smotron cluster spec updates

makhov:
Fixes a bug with constant spec update, which leads to API over-usage and OOM


apedriza:
Just to be sure I understand the issue:
1.  K0smotronControlPlane controller updates or creates KMC.spec using the K0sControlPlane spec.
2. KMC controller then generates new fields in KMC.spec, which triggers reconciliation again in step 1 using K0sControlPlane.spec.

This leads to an infinite loop.

IMO, the most intuitive approach would be to always update KMC.spec at the end (e.g., using defer). That way, we don’t need to account for these kinds of edge cases when future updates are made to the spec resource.

To prevent the K0sControlPlane controller from reconciling again, we could slightly customize the `SetupWithManager` function by adding an event filter. Alternatively, we could ensure that the K0sControlPlane controller only updates the fields in KMC.spec that we know will not be modified by the KMC controller.

If we don’t persist the changes to KMC.spec, things generated by the kmc controller like the API server certificates or k0s configuration will only exist in-memory for sts generation, which can be confusing during debugging since KMC.spec wouldn't reflect how the cluster is actually configured. WDYT?

PR #1005, opened 2025-04-26
Add jinja template cloud init

Alt3rius:
Add the jinja templating header to cloud init file to allow for templating when generating k0s service from cloud init. Useful for injecting metadata from cloud instances into k0s, like provider-id or node public ip. 

The idea comes from a similar idea in Kubeadm node bootstraper for clusterapi

### DIFF
--- a/internal/cloudinit/cloudinit.go
+++ b/internal/cloudinit/cloudinit.go
@@ -41,11 +41,11 @@ func (c *CloudInit) AsBytes() ([]byte, error) {
 	var b bytes.Buffer
 
 	// Write the "header" first
-	_, err = b.WriteString("## template: jinja\n")
+	_, err := b.WriteString("## template: jinja\n")
 	if err != nil {
 		return nil, err
 	}
-	_, err := b.WriteString("#cloud-config\n")
+	_, err = b.WriteString("#cloud-config\n")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


Alt3rius:
Fixed the unittest and retested build tests locally, should pass now without problems. Workflow can be re-run.

jnummelin:
hmm, I'm not sure why you need jinja templating since k0smotron does NOT add any templates in the cloud init? Care to elaborate bit more on the use case where you'd need this?

Alt3rius:
For example, when creating clusters with dedicated CCM on DigitalOcean, the provider-id needs to be provided to kubelet extra args during installation. The way they have it implemented in their examples for ClusterAPI, they use kubeadm bootstrap provider, that uses jinja templating in cloud init to pull the variables from cloud init metadata and template them into the kubelet service.

I've tried to find any other way to implement this, but so far, the only one that worked was by using jinja templating in cloud init. Also, I've researched the implementation in Kubeadm Bootstrap and they're adding the templating header to the cloud init by default.

jnummelin:
@Alt3rius thanks for the example, makes much more sense to me now 😄 

Could you maybe sqush the commits into one since the two last ones are kinda unnecessary as separate commits.

Alt3rius:
@jnummelin Squashed everything into 1 commit

makhov (approved):

makhov:
@Alt3rius Thanks for the PR!

PR #1007, opened 2025-04-29
Recreate control plane machines on K0sControlPlane spec changes

makhov:
Triggers machine recreation on K0sControlPlane change

### DIFF
--- a/internal/controller/bootstrap/controlplane_bootstrap_controller.go
+++ b/internal/controller/bootstrap/controlplane_bootstrap_controller.go
@@ -173,6 +173,7 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 
+	oldConfig := config.DeepCopy()
 	defer func() {
 		// Always report the status of the bootsrap data secret generation.
 		conditions.SetSummary(config,
@@ -180,7 +181,7 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
 				bootstrapv1.DataSecretAvailableCondition,
 			),
 		)
-
+		config.Spec = oldConfig.Spec
 		err := patchHelper.Patch(ctx, config)
 		if err != nil {
 			log.Error(err, "Failed to patch K0sControllerConfig status")
@@ -207,7 +208,7 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	scope.machines = machines
 
-	bootstrapData, err := c.generataBootsrapDataForController(ctx, log, scope)
+	bootstrapData, err := c.generateBootstrapDataForController(ctx, log, scope)
 	if err != nil {
 		// if the bootstrap data generation corresponds to a controller that is not the initial one, it is common to try to obtain
 		// the IP of the first controller when has not yet been surfaced. This is required to create a join token. It is needed to
@@ -266,7 +267,7 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
 	return ctrl.Result{}, nil
 }
 
-func (c *ControlPlaneController) generataBootsrapDataForController(ctx context.Context, log logr.Logger, scope *ControllerScope) ([]byte, error) {
+func (c *ControlPlaneController) generateBootstrapDataForController(ctx context.Context, log logr.Logger, scope *ControllerScope) ([]byte, error) {
 	var (
 		files      []cloudinit.File
 		installCmd string
--- a/internal/controller/controlplane/helper.go
+++ b/internal/controller/controlplane/helper.go
@@ -19,6 +19,7 @@ package controlplane
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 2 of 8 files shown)


internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -173,14 +173,15 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
...
+	oldConfig := config.DeepCopy()
 	defer func() {
 		// Always report the status of the bootsrap data secret generation.
 		conditions.SetSummary(config,
 			conditions.WithConditions(
 				bootstrapv1.DataSecretAvailableCondition,
 			),
 		)
-
+		config.Spec = oldConfig.Spec
```
apedriza, line 184: Can we add a comment here about why we reset config.spec? 

nit: for me to use the k0sconfigcontroller as a memory store for the data needed to initialize the cloud init, but not persist those modifications, sounds strange. I think using just a variable and not the config resource to store the necessary changes to form the k0sconfig file and the install command (where the controller adds additional data) would be more understandable and to have less friction if in the future it is necessary to persist some changes in the config.spec.

internal/controller/controlplane/helper.go
```diff
@@ -264,6 +282,34 @@ func (c *K0sController) generateMachineFromTemplate(ctx context.Context, name st
 	return infraMachine, nil
 }
 
+func (c *K0sController) needsRecreation(bootstrapConfigs map[string]bootstrapv1.K0sControllerConfig, kcp *cpv1beta1.K0sControlPlane, machine *clusterv1.Machine) bool {
```
apedriza, line 285:
```suggestion
func (c *K0sController) hasControllerConfigChanged(bootstrapConfigs map[string]bootstrapv1.K0sControllerConfig, kcp *cpv1beta1.K0sControlPlane, machine *clusterv1.Machine) bool {
```
Maybe more descriptive?

internal/controller/controlplane/helper.go
```diff
@@ -264,6 +282,34 @@ func (c *K0sController) generateMachineFromTemplate(ctx context.Context, name st
...
+
+	bootstrapConfig, found := bootstrapConfigs[machine.Name]
+	if !found {
+		return false
+	}
+
+	// k0s config will be reconciled using dynamic config, so leave it out of the comparison
+	bootstrapConfig.Spec.K0sConfigSpec.K0s = kcp.Spec.K0sConfigSpec.K0s
+	// leave out the tunneling spec for the bootstrap config
+	bootstrapConfig.Spec.K0sConfigSpec.Tunneling = kcp.Spec.K0sConfigSpec.Tunneling
```
apedriza, line 309: Do we need to reset tunneling? I thought k0scontrollerconfig tunneling is the same as kcp tunneling, isn't it?

apedriza (approved):
no blocking comments

aglarendil:
I tested the image built by @makhov with PRs #1007 and #1008 and it seems that 1007 does not work as expected - k0scontrolplane gets updated, but nodes do not get re-rolled. supposedly, because, a different set of parameters needs to be monitored, not only k0scontrolplane spec, but also k0s.api.extraArgs 
here is my k0scontrolplane config change (wanted to test enabling audit logs)

```
apiVersion: v1
items:
- apiVersion: controlplane.cluster.x-k8s.io/v1beta1
  kind: K0sControlPlane
  metadata:
    annotations:
      meta.helm.sh/release-name: vkuklin-kcm-03-kof-03-eu-openstack-cluster-deployment
      meta.helm.sh/release-namespace: kcm-system
    creationTimestamp: "2025-05-05T11:33:30Z"
    finalizers:
    - k0s.controlplane.cluster.x-k8s.io
    generation: 3
    labels:
      app.kubernetes.io/managed-by: Helm
      cluster.x-k8s.io/cluster-name: vkuklin-kcm-03-kof-03-eu-openstack-cluster-deployment
      helm.toolkit.fluxcd.io/name: vkuklin-kcm-03-kof-03-eu-openstack-cluster-deployment
      helm.toolkit.fluxcd.io/namespace: kcm-system
    name: vkuklin-kcm-03-kof-03-eu-openstack-cluster-deployment-cp
    namespace: kcm-system
    ownerReferences:
    - apiVersion: cluster.x-k8s.io/v1beta1
      blockOwnerDeletion: true
      controller: true
      kind: Cluster
      name: vkuklin-kcm-03-kof-03-eu-openstack-cluster-deployment
      uid: 17fd9ad3-83cc-42a2-9f6e-c921496c967b
    resourceVersion: "12428671"
    uid: d0ce449b-0bc5-4397-9831-befabdd6b879
  spec:
    k0sConfigSpec:
      args:
      - --enable-worker
      - --enable-cloud-provider
      - --kubelet-extra-args="--cloud-provider=external"
      - --disable-components=konnectivity-server
...
      k0s:
        apiVersion: k0s.k0sproject.io/v1beta1
        kind: ClusterConfig
        metadata:
          name: k0s
        spec:
          api:
            externalAddress: 172.19.122.224
            extraArgs:
              anonymous-auth: "true"
              audit-log-path: /var/log/audit/kubernetes-audit.log <- added this one
```

makhov:
@aglarendil Could you try this image `makhov/k0smotron:v1.5.2-with-1007-and-1008-2`? It should trigger an update on the `spec.api.extraArgs` change. 

Regarding the rolling update on the k0smotron controller update, could you also provide a `K0sControllerConfig` object from one of the controllers.

PR #1009, opened 2025-04-29
feat(bootstrap): add secretMetadata field to propagate metadata to bootstrap Secret

kahirokunn:
## Description

This PR adds a new `secretMetadata` field to K0sWorkerConfig that allows users to explicitly specify which labels and annotations should be propagated to the generated bootstrap Secret.

## Motivation

Currently, there is no way to add custom metadata (labels and annotations) to the bootstrap Secrets generated from K0sWorkerConfig. This feature enables users to:
- Add organizational metadata for better resource management
- Apply labels for cost allocation and team ownership
- Include annotations for documentation and operational purposes

## Changes

- Added new `SecretMetadata` type with `Labels` and `Annotations` fields
- Added `secretMetadata` field to `K0sWorkerConfigSpec`
- Updated the bootstrap controller to apply metadata from `secretMetadata` to the generated Secret
- Added comprehensive tests for the new functionality
- Created sample configuration demonstrating usage

## Example Usage

```yaml
kind: K0sWorkerConfig
metadata:
  name: worker-config
  namespace: default
spec:
  version: v1.30.0+k0s.0
  secretMetadata:
    labels:
      environment: production
      team: platform
      app: my-application
    annotations:
      cost-center: engineering
      owner: platform-team@example.com
      description: "Bootstrap configuration for production workers"
```

## Testing

- [x] Unit tests added for the new functionality
- [x] Existing tests pass
- [x] Sample configuration added

## Documentation

- [x] Added inline documentation for the new field
- [x] Created sample YAML demonstrating usage


apedriza:
bootstrap secrets add owner references over k0sworkerconfig, shouldn't that be enough for selection? or maybe just propagate the k0sworkerconfig name as label? I'm thinking in cases where we don't want to propagate to the secret metadata from workerconfig

kahirokunn:
Hi @apedriza ,

Thank you for the detailed feedback.

I’m building a component that decides whether to upload a cloud-init file to S3 by correlating each `K0sWorkerConfig` with the bootstrap `Secret` it produces. Because the `Secret` currently lacks the labels and annotations from the originating `K0sWorkerConfig`, I have to fetch the parent object and reconcile the two on every request—logic that becomes quite involved, especially in a multi-tenant setup.

In other controllers (for example, Deployment → ReplicaSet → Pod) it’s common to propagate selected metadata so higher-level tooling can filter, audit, or enforce policy without an extra lookup. Having similar capabilities here would greatly simplify integration and improve ecosystem compatibility.

To avoid copying *all* metadata—and the concerns that come with it—one proven approach is to propagate only labels or annotations that match a configurable prefix. The actions-runner-controller project recently adopted this pattern (see [actions-runner-controller#3607](https://github.com/actions/actions-runner-controller/pull/3607)) and it works well.

With that in mind, would either of the following options be acceptable?

1. **Prefix-based propagation** – copy or exclude only labels/annotations whose keys begin with a user-defined prefix (e.g. `k0s.konvoy.dev/`).  

I’m happy to adjust the PR toward whichever solution you prefer. Please let me know your thoughts.

Thanks again for the constructive discussion!

Best regards,  
Kahiro

kahirokunn:
This makes it easier to link with Sveltos, but since it is not a standard CAPI specification, you may close it. 🙏

apedriza:
> This makes it easier to link with Sveltos, but since it is not a standard CAPI specification, you may close it. 🙏

yeah that is my main concern, that is not strictly necessary on the k0smotron side. It seems more like a workaround for a particular need. WDYT @makhov ?

apedriza:
> > This makes it easier to link with Sveltos, but since it is not a standard CAPI specification, you may close it. 🙏
> 
> yeah that is my main concern, that is not strictly necessary on the k0smotron side. It seems more like a workaround for a particular need. WDYT @makhov ?

I've discussed this with @jnummelin about allow that propagation from workerconfig into the secret and what would possibly be a good idea is to add a new field in the workerConfig to place values we want to append in the secret metadata. In this case something like:

```yaml
kind: K0sWorkerConfig
. . . 
spec:
  secretMetadata:
     labels:
       your-label: here
     annotations:
       your-annotation: here
```
that way we don't propagate not desired data from workerconfig into secret, just the one configured in `secretMetadata`

kahirokunn:
> > > This makes it easier to link with Sveltos, but since it is not a standard CAPI specification, you may close it. 🙏
> > 
> > 
> > yeah that is my main concern, that is not strictly necessary on the k0smotron side. It seems more like a workaround for a particular need. WDYT @makhov ?
> 
> I've discussed this with @jnummelin about allow that propagation from workerconfig into the secret and what would possibly be a good idea is to add a new field in the workerConfig to place values we want to append in the secret metadata. In this case something like:
> 
> ```yaml
> apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
> kind: K0sWorkerConfig
> . . . 
> spec:
>   secretMetadata:
>      labels:
>        your-label: here
>      annotations:
>        your-annotation: here
> ```
> 
> that way we don't propagate not desired data from workerconfig into secret, just the one configured in `secretMetadata`

I agree with that.
If we implement it, can you merge it? 👀

apedriza:
> > > > This makes it easier to link with Sveltos, but since it is not a standard CAPI specification, you may close it. 🙏
> > > 
> > > 
> > > yeah that is my main concern, that is not strictly necessary on the k0smotron side. It seems more like a workaround for a particular need. WDYT @makhov ?
> > 
> > 
> > I've discussed this with @jnummelin about allow that propagation from workerconfig into the secret and what would possibly be a good idea is to add a new field in the workerConfig to place values we want to append in the secret metadata. In this case something like:
> > ```yaml
> > apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
> > kind: K0sWorkerConfig
> > . . . 
> > spec:
> >   secretMetadata:
> >      labels:
> >        your-label: here
> >      annotations:
> >        your-annotation: here
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > that way we don't propagate not desired data from workerconfig into secret, just the one configured in `secretMetadata`
> 
> I agree with that. If we implement it, can you merge it? 👀

yeah I think that way is safe to merge. Do you want to take care of those changes?

kahirokunn:
Hi @apedriza,

Sure thing—I've implemented the requested changes and pushed the update. When you have a moment, I’d really appreciate your review.

Thank you!

kahirokunn:
🚀 **Functionality Verification Report**

I manually tested the **`spec.secretMetadata`** feature added in the latest commit `4e79111`.

### Steps

1. Applied the following `K0sWorkerConfig` using `kubectl apply -f -`

```yaml
kind: K0sWorkerConfig
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"bootstrap.cluster.x-k8s.io/v1beta1","kind":"K0sWorkerConfig","metadata":{"annotations":{},"name":"docker-test-0","namespace":"default"},"spec":{"version":"v1.27.2+k0s.0"}}
  creationTimestamp: "2025-06-19T05:04:44Z"
  generation: 2
  labels:
  name: docker-test-0
  namespace: default
  ownerReferences:
  - apiVersion: cluster.x-k8s.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: Machine
    name: docker-test-0
    uid: c0d6b822-c38b-4ee6-8f83-140c180ec07a
  resourceVersion: "7217"
  uid: f1fd68f5-0b67-4feb-be70-153b43ec5418
spec:
  secretMetadata:
    annotations:
      description: Bootstrap configuration for production workers
      owner: platform-team@example.com
    labels:
      app: my-application
      team: platform
  useSystemHostname: false
  version: v1.27.2+k0s.0
status:
  conditions:
  - lastTransitionTime: "2025-06-19T05:06:20Z"
    status: "True"
    type: Ready
  - lastTransitionTime: "2025-06-19T05:06:20Z"
    status: "True"
    type: DataSecretAvailable
  dataSecretName: docker-test-0
  ready: true
```

2. Verified the generated Secret

```yaml
apiVersion: v1
data:
  value: xxx
kind: Secret
metadata:
  annotations:
    description: Bootstrap configuration for production workers
    owner: platform-team@example.com
  creationTimestamp: "2025-06-19T05:20:51Z"
  labels:
    app: my-application
    team: platform
  name: hoge-fuga
  namespace: default
  ownerReferences:
  - apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
    controller: true
    kind: K0sWorkerConfig
    name: hoge-fuga
    uid: 1c34ffe2-f054-473f-b749-4b8e2bf72bf4
  resourceVersion: "7500"
  uid: 621788e4-a4d5-4eaf-9aa6-31d01d0dcea3
type: cluster.x-k8s.io/secret
```

### Results

* The values specified in `secretMetadata.labels / annotations` were **successfully transferred to the bootstrap Secret**.

Please review and confirm! 🙌

internal/controller/bootstrap/worker_bootstrap_controller_test.go
```diff
@@ -577,3 +577,119 @@ func TestReconcileGenerateBootstrapData(t *testing.T) {
 		}
 	}, 20*time.Second, 100*time.Millisecond)
 }
+
+func TestReconcileCopyLabelsAndAnnotations(t *testing.T) {
```
apedriza: Good to add test for this feature 👍 . 

nit: I wonder if would make sense to encapsulate secret creation given a worker config in a func to test that specific logic instead of create a full test environment for testing the addition of this metadata. It looks excessive 
 IMO, and it is faster run that proposed unit test than the current integration test. Alternative would be to reuse `TestReconcileGenerateBootstrapData` test and check there the secret created has the expected values (I don't know how I could forget to check secret is created in that test 😶‍🌫️ )

kahirokunn: Thank you very much!
I tried this.

kahirokunn:
@apedriza I’ve addressed all of your review comments. Could you please review it again? Thank you. 🙏

internal/controller/bootstrap/worker_bootstrap_secret_test.go
```diff
@@ -0,0 +1,154 @@
+/*
+
```
apedriza, line 3: can you add the **Copyright**  as other files includes ?

kahirokunn: Thx. I did it.

kahirokunn:
@apedriza I’ve addressed all of your review comments again. Could you please review it again? 
I actually created the cluster and manually retested everything.
Thank you. 🙏

kahirokunn:
I'm looking forward to using this feature.
Thank you for your support until the merge!

PR #1017, opened 2025-05-08
feat(etcd): allow custom ResourceRequirements for etcd StatefulSet

kahirokunn:
## What this PR does / why we need it  
Adds first-class support for configuring CPU / memory requests & limits for the
etcd StatefulSet created by k0smotron in HCP mode.

## What this PR adds  
* `EtcdSpec.Resources` field (optional) in `api/v1beta1`  
* CRD schemas and sample manifest updated  
* Controller now applies the specified `ResourceRequirements`  
* Documentation section “Resource Requirements” added to `docs/etcd.md`  
* Unit tests for resource propagation

## Special notes for reviewers  
* The new field is fully optional and backwards-compatible.  
* If `spec.etcd.resources` is omitted the behavior is unchanged.  
* Follow-up work: expose default requests/limits via Helm chart values.

## Checklist  
- [x] PR title follows Conventional Commits  
- [x] Unit tests added/updated  
- [x] Docs updated  
- [x] Rebased on latest `main`


internal/controller/k0smotron.io/k0smotroncluster_etcd.go
```diff
@@ -345,6 +345,7 @@ func (r *ClusterReconciler) generateEtcdStatefulSet(kmc *km.Cluster, existingSts
 							{Name: "ETCDCTL_KEY", Value: "/var/lib/k0s/pki/etcd/server.key"},
 							{Name: "ETCD_INITIAL_CLUSTER", Value: r.initialCluster(kmc, replicas)},
 						},
+						Resources: getEtcdResourceRequirements(kmc),
```
apedriza, line 348: we could directly use the resources without the pointer. This way we don't need the func that checks if the pointer is null. Same behavior that the controlplane resources have.

kahirokunn:
Hi @apedriza, could you review it again? Thx.

apedriza (approved):
LGTM!

PR #1019, opened 2025-05-09
Fix panic on non-CAPI installations

makhov:
Currently it panics due to missing CRDs


apedriza:
what if CAPI is installed after k0smotron? Shouldn't be up to the user to install needed stuff or just throw a warning from k0smotorn side?

makhov:
I've added log message, I think it's better like this than panic.

apedriza:
I been testing this and I still see the panic 

```
2025-06-19T09:32:50Z	ERROR	controller-runtime.source.EventHandler	if kind is a CRD, it should be installed before calling Start	{"kind": "Machine.cluster.x-k8s.io", "error": "no matches for kind \"Machine\" in version \"cluster.x-k8s.io/v1beta1\""}
sigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind[...]).Start.func1.1
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.5/pkg/internal/source/kind.go:71
k8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext.func2
	/go/pkg/mod/k8s.io/apimachinery@v0.30.3/pkg/util/wait/loop.go:87
k8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext
	/go/pkg/mod/k8s.io/apimachinery@v0.30.3/pkg/util/wait/loop.go:88
k8s.io/apimachinery/pkg/util/wait.PollUntilContextCancel
	/go/pkg/mod/k8s.io/apimachinery@v0.30.3/pkg/util/wait/poll.go:33
sigs.k8s.io/controller-runtime/pkg/internal/source.(*Kind[...]).Start.func1
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.18.5/pkg/internal/source/kind.go:64
```

I think this could be related to the webhook implementation, at least it starts to appear after `v1.5.0` where webhook was included (among other things but looks like the more notable change that can be related, not sure)

This makes me think in the possibility to have two different installations one for k0smotron standalone an another *full k0smotron* including capi integration. That it makes sense?

PR #1020, opened 2025-05-12
[K0smotronControlPlane] Add Infrastructure status patching

s3rj1k:
Implements automatic patching of Infrastructure object (referenced via cluster.Spec.InfrastructureRef) ready status for objects with the `k0smotron` managed-by annotation.

Related: https://github.com/k0smotron/k0smotron/issues/668

### DIFF
--- a/internal/controller/controlplane/k0smotron_controlplane_controller.go
+++ b/internal/controller/controlplane/k0smotron_controlplane_controller.go
@@ -18,6 +18,7 @@ package controlplane
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
@@ -55,6 +56,14 @@ import (
 	"github.com/k0sproject/version"
 )
 
+const (
+	// AnnotationKeyManagedBy is the annotation key that indicates which controller manages the infrastructure object
+	AnnotationKeyManagedBy = "cluster.x-k8s.io/managed-by"
+
+	// AnnotationValueManagedByK0smotron is the value for the managed-by annotation
+	AnnotationValueManagedByK0smotron = "k0smotron"
+)
+
 type K0smotronController struct {
 	client.Client
 	SecretCachingClient client.Client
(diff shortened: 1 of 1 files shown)


internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -442,6 +458,69 @@ func (c *K0smotronController) getComparableK0sVersionRunningInPod(ctx context.Co
...
+	if err != nil {
+		return fmt.Errorf("failed to get ready status: %w", err)
+	}
+
+	// Only patch if status is not set or different from desired value
+	if !found || currentReady != ready {
+		log.Info("Patching Infrastructure object status", "ready", ready)
+
+		// Convert patch to JSON
+		b, err := json.Marshal(map[string]any{
```
makhov, line 499: It's a simple patch, we can stick with somthing like:
```
readyPatchBytes := []byte(fmt.Sprintf(`{"status": {"ready": %t}}`, ready))
```

internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -55,6 +55,14 @@ import (
 	"github.com/k0sproject/version"
 )
 
+const (
+	// AnnotationKeyManagedBy is the annotation key that indicates which controller manages the infrastructure object
+	AnnotationKeyManagedBy = "cluster.x-k8s.io/managed-by"
+
+	// AnnotationValueManagedByK0smotron is the value for the managed-by annotation
+	AnnotationValueManagedByK0smotron = "k0smotron"
```
jnummelin, line 63: The value of the annotation is not important. CAPI spec says the value can be anything.

s3rj1k: yea, but someone can actually want to handle that object with personal controller, without matching value it won't be possible

s3rj1k: I see some edge-cases when simple approach that is done here is not enough and one would need to have extra patching steps in place by some other controller

jnummelin: This is a good point that looking only for this specific _value_ we still leave the door open for someone/something else handle some special cases 👍

jnummelin (approved):

PR #1043, opened 2025-06-03
Correct spelling of "Initialized" field name

kahirokunn:


api/controlplane/v1beta1/k0s_types.go
```diff
@@ -145,7 +145,7 @@ type K0sControlPlaneStatus struct {
 	// The value of this field is never updated after provisioning is completed. Please use conditions
```
apedriza, line 145: can we also fix the comment of the field ? `KubeadmControlPlane` to `K0sControlPlane`

kahirokunn: Thx. I did it.

makhov:
I'm a bit worried about that we changing CRDs, what happens to existing clusters? Maybe it makes sense to deprecate the field, add a new one and use it?

kahirokunn:
@makhov 
Thank you for the concern!
In this PR we only corrected the typo in the Go struct field name (Inititalized → Initialized).
The JSON tag remains exactly the same (json:"initialized"), so the generated CRD definition does not change at all and no stored data on existing clusters is touched. Because the on-wire field name is unchanged, there’s no need to deprecate or migrate anything—everything will continue to work as before.

makhov:
Oh, you're right! LGTM then!

PR #1046, opened 2025-06-03
fix: Resolve K0smotronControlPlane reconciliation loop caused by Etcd resources initialization

### LINKED ISSUE #1047: K0smotronControlPlane status.ready never becomes true when etcd field is partially configured
### What happened?

K0smotronControlPlane gets stuck in a reconciliation loop and its `.status.ready` field never becomes `true` when the `etcd` field is partially configured without the `resources` field.

The controller continuously keeps updating the k0smotron Cluster spec, preventing the reconciliation from completing. This issue started occurring from v1.5.3.

Reproduction in CI: https://github.com/k0sproject/k0smotron/actions/runs/15410066553/job/43360285044?pr=1046

### Involved Areas

Cluster API integration

K0smotronControlPlane should reconcile successfully and `.status.ready` should become `true` when the cluster is ready, even when the etcd field is partially configured without explicitly defining the resources field.

### Steps to reproduce

1. Create a K0smotronControlPlane with partial etcd configuration:

```yaml
kind: K0smotronControlPlane
metadata:
  name: docker-test-cp
spec:
  version: v1.33.1-k0s.0
  etcd:
    autoDeletePVCs: true
    # resources:
    #   requests:
    #     cpu: 100m
    #     memory: 128Mi
    #   limits:
    #     cpu: 200m
    #     memory: 256Mi
  persistence:
    type: pvc
    persistentVolumeClaim:
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 50Mi
  service:
    type: NodePort
```

2. Apply the manifest
3. Check the controller logs: `kubectl logs -n k0smotron deployment/k0smotron-controller-manager`
4. Observe that:
   - `.status.ready` remains `false` indefinitely
   - The reconciliation loop continues

### k0smotron version

v1.5.3

### k0s version

v1.33.1-k0s.0

**Root Cause**: The kubebuilder default annotation for the `Etcd` field doesn't include `resources`, causing `Etcd.Resources` to be initialized inconsistently (sometimes `nil`, sometimes `{}`). The `isClusterSpecEqual` function treats these as different values, triggering continuous spec updates.

**Workaround**: Explicitly define the resources field in the etcd configuration:
```yaml
etcd:
  autoDeletePVCs: false
  image: quay.io/k0sproject/etcd:v3.5.13
  resources: {}  # Add this line
```

**Impact**: This is a regression from v1.5.3 that prevents users from deploying K0smotronControlPlane with a common configuration pattern (partial etcd configuration).

kahirokunn:
## Summary

Fixes #1047

This PR resolves a critical regression introduced in v1.5.3 where K0smotronControlPlane enters an infinite reconciliation loop when the `etcd` field is partially configured without the `resources` field.

## Problem

When users configure K0smotronControlPlane with partial etcd settings (e.g., only `autoDeletePVCs` or `image`), the controller gets stuck in a reconciliation loop:

- `.status.ready` never becomes `true`
- The k0smotron Cluster spec is repeatedly updated

**Live reproduction**: https://github.com/k0sproject/k0smotron/actions/runs/15410066553/job/43360285044?pr=1046

## Root Cause

The issue stems from inconsistent initialization of `Etcd.Resources`:

1. The kubebuilder default for `Etcd` was missing the `resources` field:
   ```go
   //+kubebuilder:default={"image":"...","persistence":{}}  // No resources!
   ```

2. This caused `Etcd.Resources` to be initialized differently:
   - Sometimes as `nil`
   - Sometimes as empty struct `{}`

3. The `isClusterSpecEqual` function using `reflect.DeepEqual` treats these as different, causing endless updates

### 1. Fix the root cause (commit 1)
Added `"resources":{}` to the kubebuilder default:
```diff
-//+kubebuilder:default={"image":"quay.io/k0sproject/etcd:v3.5.13","persistence":{}}
+//+kubebuilder:default={"image":"quay.io/k0sproject/etcd:v3.5.13","persistence":{},"resources":{}}
```

### 2. Prevent regression (commit 2)
Enhanced the test to cover partial etcd configuration:
```yaml
etcd:
  autoDeletePVCs: true  # Partial config without resources
```

## Verification

- ✅ K0smotronControlPlane reconciles successfully
- ✅ `.status.ready` becomes `true` when cluster is ready
- ✅ No more spec comparison loops in controller logs
- ✅ Test now covers the problematic scenario

## Impact

- **Fixes regression**: Restores functionality broken since v1.5.3
- **Backward compatible**: Empty resources struct equals no resource constraints
- **Common use case**: Partial etcd configuration is a typical pattern

### DIFF
--- a/inttest/capi-docker/capi_docker_test.go
+++ b/inttest/capi-docker/capi_docker_test.go
@@ -268,6 +268,8 @@ spec:
       type: sa
     - name: docker-test-etcd
       type: etcd
+  etcd:
+    autoDeletePVCs: true
   persistence:
     type: pvc
     persistentVolumeClaim:


apedriza:
is this still needed? I think the fix was already merged

kahirokunn:
Since I rebase in main, let's try to see if CI passes.

kahirokunn:
It is still not ready.
This change is necessary.

kahirokunn:
It seems that it no longer occurs.
In that case, we will close this PR.
Thx!

PR #1049, opened 2025-06-03
A minor worker version detection improvement. 

makhov:
Allows setting a plain Kubernetes version on `K0sWorkerConfig.spec.version` without `+k0s.0` suffix


kahirokunn:
This is awesome 🙌

kahirokunn:
How about add the same functionality to the control plane? 👀

makhov:
For the `K0sContollerConfig` it doesn't make much difference since we choose the version on `K0sControlPlane` level, but it's good to keep it consistent. Thanks for pointing it out.

kahirokunn:
Shouldn't `Cluster` resources created from `K0smotronControlPlane` also have the `-k0s` suffix? 👀

apedriza:
I wonder if this logic should be part of the admission webhook. WDYT?

kahirokunn:
I'm concerned that when these resources are managed by CD tools such as ArgoCD and changes are made using mutating admission webhooks, diffs will always occur.

kahirokunn:
I didn't know that the k0smotron control plane already had a version detection feature.
And I encountered a bug related to that.
https://github.com/k0sproject/k0smotron/issues/1052

I'm a little concerned about whether K0sWorkerConfig's status.version also needs to be careful not to violate the cluster API specification.

PR #1051, opened 2025-06-04
Use k0smotron SA for HCP

makhov:
Using a k0smotron SA for HCP and etcd statefulsets instead of the default one

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_controller.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_controller.go
@@ -22,6 +22,8 @@ import (
 	"time"
 
 	apps "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
@@ -38,6 +40,8 @@ import (
 
 	km "github.com/k0sproject/k0smotron/api/k0smotron.io/v1beta1"
 )
+
+const serviceAccountName = "k0smotron-hcp"
 
 var patchOpts []client.PatchOption = []client.PatchOption{
 	client.FieldOwner("k0smotron-operator"),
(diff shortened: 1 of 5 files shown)


jnummelin:
So now we create SA, roles and bindings for each HCP "instance", right?  We also need to make sure that we delete all those when user deletes the HCP CR. One idea would be to use ownerReferences for this so kube should cascade the delete automatically

internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -246,6 +259,71 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			Namespace: kmc.Namespace,
+		},
+	}
+	if err := r.Get(ctx, client.ObjectKey{Name: serviceAccountName, Namespace: kmc.Namespace}, sa); client.IgnoreNotFound(err) != nil {
+		return err
+	} else if err != nil {
+		if err := r.Create(ctx, sa); err != nil {
+			return err
+		}
+	}
```
jnummelin: While this pattern is good in general, I think this creates a bit of a problem. Say in some future version we need to change the roles/bindings, we would not be able to do so since the old ones would already exist

internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -39,6 +41,8 @@ import (
 	km "github.com/k0sproject/k0smotron/api/k0smotron.io/v1beta1"
 )
 
+const serviceAccountName = "k0smotron"
```
jnummelin: I'd maybe rename this to something like `k0smotron-hcp` just to say that it's NOT for k0rmotron itself

makhov:
It's actually not per cluster, but per namespace. Maybe cluster+owner reference makes more sense, but will lead to some resource duplication in the cluster.

internal/controller/k0smotron.io/k0smotroncluster_controller.go
makhov: Fixed

internal/controller/k0smotron.io/k0smotroncluster_controller.go
makhov: Done

apedriza:
Don't know if I'm missing something, but what about let the user set a SA in the Cluster spec and use it if provided? that way we don't worry about what roles to add into the SA and it is more flexible

makhov:
@apedriza Yeah, I think you're right. I was thinking about the automated flow, but looks like currently it's easier for everyone is just to use a given SA

Issue #1052, opened 2025-06-05
MachineSet Preflight Checks fail due to version mismatch between K0smotronControlPlane spec.version and status.version


kahirokunn:
## What happened?

When using k0smotron with Cluster API's MachineSet Preflight Checks feature enabled, the `ControlPlaneIsStable` preflight check continuously fails, causing MachineDeployment scale-up operations to be held indefinitely.

Specifically, the K0smotronControlPlane has `spec.version` set to `v1.33.1` while `status.version` is `v1.33.1-k0s.0`. Due to this build metadata difference, Cluster API incorrectly determines that the control plane is "upgrading".

The following message is repeatedly logged every 15 seconds:
```
Scale up on hold because K0smotronControlPlane k0smotron-k1/k1-dhd6j is upgrading ("ControlPlaneIsStable" preflight check failed)
```

## Involved Areas

Cluster API integration

## What did you expect to happen?

When the K0smotronControlPlane is in a stable state, the `ControlPlaneIsStable` check in MachineSet Preflight Checks should pass, allowing MachineDeployment to scale up normally.

## Steps to reproduce

1. Prepare a Cluster API management cluster
2. Set up the environment with the following commands:

```bash
export CLUSTER_TOPOLOGY="true"
export EXP_RUNTIME_SDK="true"
export EXP_MACHINE_SET_PREFLIGHT_CHECKS="true"

clusterctl init \
  --bootstrap k0sproject-k0smotron \
  --control-plane k0sproject-k0smotron \
  --infrastructure k0sproject-k0smotron,docker \
  --core cluster-api
```

3. Create a cluster using k0smotron + Docker provider cluster class (see [documentation](https://github.com/k0sproject/k0smotron/blob/main/docs/capi-clusterclass.md))
Example:

```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  name: test
spec:
  topology:
    class: k0smotron-docker-etcd
    classNamespace: k0smotron
    version: v1.33.1
    workers:
      machineDeployments:
        - class: docker-worker
          name: default
          replicas: 2
  clusterNetwork:
    pods:
      cidrBlocks:
      - 192.168.0.0/16
    serviceDomain: cluster.local
    services:
      cidrBlocks:
      - 10.128.0.0/12
```

4. Create a MachineDeployment and attempt to scale up worker nodes
5. Check the capi-controller-manager logs

## k0smotron version

[...]

Issue #1027, opened 2025-05-20
wrong `version` field in status of `k0smotronControlPlane`


cwrau:
### What happened?

I have `v1.28.9` in the spec which resulted in `v1.28.9-k0s.0` in the status field of the k0smotronControlPlane.

As `-k0s.0` denotes as prerelease, this is perpetually stuck in upgrading as far as CAPI is concerned.

As far as I can tell, k0smotron [only ever deploys prereleases](https://github.com/apedriza/k0smotron/blob/main/api/k0smotron.io/v1beta1/k0smotroncluster_types.go#L116-L137), is there a reason for this?

### What did you expect to happen?

That a finished release would be used and not a prerelease.

If k0s doesn't follow semver, #sad, it should trim this suffix so CAPI can work with this.

### Steps to reproduce

Deploy a cluster with CAPI, upgrade it and see it being stuck.

### k0smotron version

1.5.2

### k0s version

N/A

jnummelin:
`v1.28.9-k0s.0` is actually the same as `v1.28.+-k0s.0` from k0smotron and k0s point of view. k0smotron replaces the `+` to `-` since Docker hub and some other registries disallow to use `+` in the tag nameing scheme. 🤦 

@makhov @apedriza I think we need to do some similar magic which we did on the `K0sControlPlane` provider with the reported version?

apedriza:
> `v1.28.9-k0s.0` is actually the same as `v1.28.9+k0s.0` from k0smotron and k0s point of view. k0smotron replaces the `+` to `-` since Docker hub and some other registries disallow to use `+` in the tag nameing scheme. 🤦
> 
> [@makhov](https://github.com/makhov) [@apedriza](https://github.com/apedriza) I think we need to do some similar magic which we did on the `K0sControlPlane` provider with the reported version?

@jnummelin you mean [this kind of logic](https://github.com/k0sproject/k0smotron/blob/main/internal/controller/controlplane/status.go#L248-L252) for being consistent with the version format between `spec.version` and `status.version`?

jnummelin:
@cwrau yes, something along those lines

kahirokunn:
The **mismatch between `spec.version` and `status.version`** discussed in this Issue is causing a derivative issue where MachineSet Preflight Checks (`ControlPlaneIsStable`) fail.
The specific symptoms and logs are documented in **Issue #1052**, and the PR to fix the root cause of this Issue has been created in #1053. I would appreciate it if you could review them together.
[...]

PR #1053, opened 2025-06-05
Fix K0smotronControlPlane version mismatch for Cluster API compatibility

kahirokunn:
## Main Content

This PR resolves the K0smotronControlPlane and Cluster API compatibility issue reported in Issue #1052:

### Problem
- Version format discrepancy between `spec.version` (e.g., `v1.33.1`) and `status.version` (e.g., `v1.33.1-k0s.0`)
- Cluster API's MachineSet Preflight Checks incorrectly recognize this as "upgrading in progress"
- MachineDeployment scale-up operations are indefinitely suspended

### Solution
1. **Version normalization function**: Unified version format using the `FormatStatusVersion` function
2. **Complete version information retention**: Store complete k0s version information in the `K0sVersion` field

The detailed PR description is saved in the `pr-title-and-description.md` file. This PR enables k0smotron to function correctly as a standard control plane provider within the Cluster API ecosystem.


jnummelin:
I wonder if it would be possible to create or extend some existing e2e test to cover this for the MD part?

internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -344,6 +344,23 @@ func (c *K0smotronController) ensureCertificates(ctx context.Context, cluster *c
 	return certificates.LookupOrGenerateCached(ctx, c.SecretCachingClient, c.Client, capiutil.ObjectKey(cluster), *metav1.NewControllerRef(kcp, cpv1beta1.GroupVersion.WithKind("K0smotronControlPlane")))
 }
 
+// FormatStatusVersion formats the status version to match the spec version format.
+// If spec.version doesn't contain "-k0s." suffix, it removes the suffix from status.version as well.
+func FormatStatusVersion(specVersion, statusVersion string) string {
+	specHasK0sSuffix := strings.Contains(specVersion, "-k0s.")
+
+	// spec.versionの形式に合わせてstatus.versionを調整
```
jnummelin: Japanese text? 🧐

internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -344,6 +344,23 @@ func (c *K0smotronController) ensureCertificates(ctx context.Context, cluster *c
...
+	if !specHasK0sSuffix && strings.Contains(statusVersion, "-k0s.") {
+		// spec.versionに-k0s.サフィックスがない場合は、status.versionからも除去
```
jnummelin: Japanese again, pls change to English 🙂

internal/controller/controlplane/k0smotron_controlplane_controller.go
kahirokunn: Sorry, I fixed.

internal/controller/controlplane/k0smotron_controlplane_controller.go
kahirokunn: Sorry, I fixed.

kahirokunn:
I've added e2e tests that cover the MachineDeployment scaling scenario. The tests verify:
Version format consistency between spec.version and status.version
MachineDeployment scale-up operations work correctly (no longer hang due to version mismatch)
Both status.version and status.k0sVersion fields are properly populated
Thanks for the feedback!

kahirokunn:
Rebased with the latest main branch.

inttest/capi-docker-machinedeployment/CHANGES.md
```diff
@@ -0,0 +1,72 @@
+# Changes Summary for K0smotronControlPlane Version Format E2E Test
```
makhov: We don't need this info in the repo. We can keep it in history like PR description, but not in the repo

inttest/capi-docker-machinedeployment/README.md
```diff
@@ -0,0 +1,57 @@
+# CAPI Docker MachineDeployment Version Format Test
```
makhov: Same thing, I don't see much value in this file because it gets outdated very quickly.

inttest/capi-docker-machinedeployment/capi_docker_test.go
```diff
@@ -66,9 +70,20 @@ func (s *CAPIDockerSuite) SetupSuite() {
 }
 
 func (s *CAPIDockerSuite) TestCAPIDocker() {
+	// Test with k0s suffix (default/backward compatible)
+	s.Run("WithK0sSuffix", func() {
```
makhov: I don't like this approach. We keep it one file - one test for integration tests.

inttest/capi-docker-clusterclass/capi_docker_clusterclass_test.go
```diff
@@ -63,6 +63,121 @@ func (s *CAPIDockerClusterClassSuite) SetupSuite() {
...
+	s.dumpClusterState()
+
+	// Collect node logs if available
+	s.collectNodeLogs()
+
+	// Call the parent teardown
+	s.FootlooseSuite.TearDownSuite()
+}
+
+func (s *CAPIDockerClusterClassSuite) dumpClusterState() {
```
makhov: We do use supportbundle for this: https://github.com/k0sproject/k0smotron/blob/main/.github/workflows/capi-smoke-tests.yml#L109

inttest/basic/basic_test.go
```diff
@@ -65,6 +65,9 @@ func (s *BasicSuite) TestK0sGetsUp() {
 	s.createK0smotronCluster(s.Context(), kc)
 	s.Require().NoError(common.WaitForStatefulSet(s.Context(), kc, "kmc-kmc-test", "kmc-test"))
 
+	s.T().Log("waiting for all deployments in kmc-test namespace to be ready")
+	s.Require().NoError(util.WaitForDeploymentsInNamespace(s.Context(), kc, "kmc-test"))
```
makhov: Do we have any deployments in the namespace? In general, I'd prefer to avoid this common func and check every object I care about explicitly.

inttest/capi-docker-machinedeployment/capi_docker_test.go
```diff
@@ -150,8 +169,88 @@ func (s *CAPIDockerSuite) TestCAPIDocker() {
...
+		Name(name).
+		Do(s.ctx).Into(kcp)
+	if err != nil {
+		return nil, err
+	}
+	return &kcp.Status, nil
+}
+
+// verifyK0smotronControlPlaneVersionFormat verifies that the version format is consistent
+func (s *CAPIDockerSuite) verifyK0smotronControlPlaneVersionFormat(clusterName, expectedSpecVersion string, expectK0sSuffix bool) error {
```
makhov: I think it's too complex func for the task. Could we just keep it simple?

inttest/capi-docker-machinedeployment/capi_docker_test.go
```diff
@@ -150,8 +169,88 @@ func (s *CAPIDockerSuite) TestCAPIDocker() {
...
+
+	s.T().Logf("K0smotronControlPlane status: version=%s, k0sVersion=%s", status.Version, status.K0sVersion)
+
+	// Verify version format
+	if expectK0sSuffix {
+		// When spec.version has -k0s. suffix, status.version should also have it
+		if !strings.Contains(status.Version, "-k0s.") {
+			return fmt.Errorf("expected status.version to contain '-k0s.' suffix, but got: %s", status.Version)
+		}
+		if status.Version != expectedSpecVersion {
```
makhov: We use testify library for asserts

makhov (changes requested):
Thanks for the PR! I appreciate the effort, but I think a lot of changes here are out of the scope, I think we can simplify it a lot

inttest/capi-docker-machinedeployment/CHANGES.md
kahirokunn: Thx. I deleted it.

inttest/capi-docker-machinedeployment/README.md
kahirokunn: Thx. I deleted it.

inttest/capi-docker-machinedeployment/capi_docker_test.go
kahirokunn: Thx. I did it.

inttest/capi-docker-machinedeployment/capi_docker_test.go
kahirokunn: Thx. I did it.

inttest/capi-docker-clusterclass/capi_docker_clusterclass_test.go
kahirokunn: Sorry, this commit is a remnant from when I was trying to fix a flaky CI issue on this branch.
I moved it to PR #1073, and some issues have been resolved, but the root cause has not been fixed.

inttest/basic/basic_test.go
kahirokunn: Sorry, this commit is a remnant from when I was trying to fix a flaky CI issue on this branch.
I moved it to PR https://github.com/k0sproject/k0smotron/pull/1073, and some issues have been resolved, but the root cause has not been fixed.

inttest/capi-docker-machinedeployment/capi_docker_test.go
kahirokunn: Thx. I did it.

kahirokunn:
Thank you for the feedback. I apologize for any confusion—because I work in the JST time zone, I assumed the team was still offline and briefly tested some CI fixes on this branch. I’ve now addressed all your comments and reverted the temporary changes. When you have a moment, could you please review it again? I’d really appreciate it.

kahirokunn:
Overcoming flaky tests and achieving a miracle I am thrilled that all CI tests passed. 😍 👍

kahirokunn:
Bump up 🤚

kahirokunn:
Conflict resolved

kahirokunn:
@makhov Hi! ✋ 
Could you please advise me on what steps are needed to move forward? 👀 
Thank you 🙏

kahirokunn:
If left as is, it causes conflicts, requiring frequent corrections, which is a bit of a hassle... 😢

cwrau (approved):
Logic looks good to me, especially since https://github.com/k0sproject/k0smotron/pull/1053/files#diff-0f7bdf411afd541628aba6e0c9feee81f9c85ad4c462de03021792f6ae2bbdc6R116 should catch anything

kahirokunn:
I'm glad you said that. 😭 Thank you 😭

apedriza:
@kahirokunn I will take a look to it next, I need to prioritize other stuff right now. Sorry for the inconvenience 🙏

kahirokunn:
I was relieved to hear from you.
Thank you for always working on such wonderful projects!
It's totally fine.
Thank you 🙏

internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -468,7 +485,11 @@ func (c *K0smotronController) computeStatus(ctx context.Context, cluster types.N
 	kcp.Status.UnavailableReplicas = int32(unavailableReplicas)
 
 	if kcp.Status.ReadyReplicas > 0 {
-		kcp.Status.Version = minimumVersion.String()
+		statusVersion := minimumVersion.String()
+		// Store the full k0s version
+		kcp.Status.K0sVersion = statusVersion
+		// Store the formatted version for Cluster API compatibility
+		kcp.Status.Version = FormatStatusVersion(kcp.Spec.Version, statusVersion)
```
apedriza: I would store in `.status.k0sVersion` just the [k0s version](https://docs.k0sproject.io/v1.31.3+k0s.0/releases/#:~:text=(k0s.0)%20reflects%20the%20k0s%20version%2C%20which%20is%20built%20on%20top%20of%20the%20certain%20Kubernetes%20version.) and in `.status.version` the k8s version. For example, for `v1.30.0-k0s.0` :

- `.status.k0sVersion` = `k0s.0`
- `.status.version` = `v1.30.0`

this way we avoid redundancy.

internal/controller/controlplane/version_format_test.go
```diff
@@ -0,0 +1,71 @@
+/*
```
apedriza: can we place this test in `k0smotron_controller_test.go` file?

inttest/capi-docker-machinedeployment-no-suffix/capi_docker_test.go
```diff
@@ -0,0 +1,164 @@
...
+	kubeClient, err := kubernetes.NewForConfig(restCfg)
+	s.Require().NoError(err)
+	s.Require().NotNil(kubeClient)
+	s.client = kubeClient
+
+	s.ctx, _ = util.NewSuiteContext(s.T())
+}
+
+func (s *CAPIDockerSuite) TestCAPIDocker() {
+	s.testCAPIDockerWithVersion("docker-md-test-no-suffix", "v1.27.2", "../../config/samples/capi/docker/cluster-with-machinedeployment-no-suffix.yaml", false)
```
apedriza: I think we should try to place the cluster declaration in the same file as the other integration tests.
On the other hand, I believe that having just one test for the MachineDeployment is enough.
For me, adding two new e2e tests to cover both version format possibilities seems excessive. We can always add an integration or unit test to go deeper in that logic.

Have you taken a look at the new e2e framework we're migrating to?
It might be a good idea to start deprecating the practice of adding tests to the `inttest` folder in favor of using the `e2e` folder.

internal/controller/controlplane/k0smotron_controlplane_controller.go
kahirokunn: Thank you for your feedback, apedriza.

According to the Cluster API specification, **`.status.version` must exactly match `.spec.version`**. Any discrepancy here causes the MachineSet Pre-flight Checks to incorrectly determine that an "upgrade is in progress," which stops the MachineDeployment scaling process (this is the root cause of the bugs reported in Issues #1052 and #1027).

Therefore, in this PR we have implemented:

* Aligning `.status.version` with `.spec.version` (e.g., `v1.33.1`) to **resolve the inconsistency**
* Storing the detailed k0s version string (e.g., `v1.33.1-k0s.0`) in the newly created field **`.status.k0sVersion`**

Your proposed approach of:

```
.status.k0sVersion = k0s.0
.status.version    = v1.30.0
```

would cause `.status.version` and `.spec.version` to be mismatched again, violating the CAPI specification mentioned above and causing the bug to reoccur. For more details on the background, please also refer to the comments in #1068.

For these reasons, we believe the current implementation is necessary to maintain CAPI compatibility. We hope you understand. Thank you for your consideration.

internal/controller/controlplane/version_format_test.go
kahirokunn: Thank you. I fixed it.

inttest/capi-docker-machinedeployment-no-suffix/capi_docker_test.go
kahirokunn: Thx. I fixed it.

kahirokunn:
@apedriza I made the corrections as instructed. How does it look? 👀

internal/controller/controlplane/k0smotron_controlplane_controller.go
apedriza: yeah, what I mean is that k0sVersion could only contained k0s version `k0s.0` in this case. But it is not a blocking thing, I'm OK with this too. I just wanted to avoid redundancy as much as possible

apedriza (commented):
from e2e output: `testing: warning: no tests to run`

seems something is missing and test is not running

e2e/machinedeployment_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+func TestMachineDeployment(t *testing.T) {
+	setupAndRun(t, func(t *testing.T) {
+		t.Run("without-k0s-suffix", func(t *testing.T) {
+			machineDeploymentSpec(t, "v1.27.2", "v1.27.2")
+		})
+
+		t.Run("with-k0s-suffix", func(t *testing.T) {
+			machineDeploymentSpec(t, "v1.27.2+k0s.0", "v1.27.2+k0s.0")
+		})
```
apedriza: I would suggest running only an e2e spec using the classic version format, without involving the k0s version. I recommend avoiding parameterizing the spec function. Everything required for the test should either be self-contained or rely on environment variables.

There is no need to introduce configurability to the spec if only a single variant is intended to run. The unit tests you created cover the different configurations and scenarios

e2e/machinedeployment_test.go
```diff
@@ -0,0 +1,233 @@
...
+	controlPlane, err := util.DiscoveryAndWaitForControlPlaneInitialized(ctx, capiframework.DiscoveryAndWaitForControlPlaneInitializedInput{
+		Lister:  bootstrapClusterProxy.GetClient(),
+		Cluster: cluster,
+	}, util.GetInterval(e2eConfig, testName, "wait-control-plane"))
+	require.NoError(t, err)
+
+	err = util.WaitForControlPlaneToBeReady(ctx, bootstrapClusterProxy.GetClient(), controlPlane, util.GetInterval(e2eConfig, testName, "wait-control-plane"))
+	require.NoError(t, err)
+
+	t.Log("Verifying K0smotronControlPlane version format")
```
apedriza: we use just fmt.Print for logs. Nothing against but simply to unify styles.

kahirokunn: Thx. I fixed it.

e2e/machinedeployment_test.go
kahirokunn: Thx. I fixed it.

kahirokunn:
@apedriza I made the corrections as instructed. How does it look? 👀 Thank you 🙏

kahirokunn:
Conflict resolved

kahirokunn:
Will this be available in `v1.7.0` ? 👀

broboa:
I am eager to test this out, because I have just laid my hands on CAPI and k0smotron, but the bug that seems to be resolved by this PR is leading me to a dead end. Thanks!

kahirokunn:
@broboa I hope you can try it out soon. 👍

apedriza:
@kahirokunn  @broboa have you try to set same version format in `Cluster.spec.topology.version` than `K0smotronControlPlaneTemplate.spec.version` ? for example, `v1.27.2-k0s.0`. MachineDeployment version comes from `Cluster.spec.version` so in that case the comparison succeed. I don't know further implications but should work. Letm me know if works for your case.

kahirokunn:
Thank you for your suggestion. While setting the same version format in both `Cluster.spec.version` and `K0smotronControlPlaneTemplate.spec.version` (e.g., v1.27.2-k0s.0) might work as a workaround in some cases, I think this PR addresses important underlying issues that need to be resolved.

Here are my main concerns with relying solely on version format alignment:

**1. Inconsistent Examples and User Experience**
The k0smotron repository contains numerous examples with inconsistent version formats - some include the `-k0s.0` suffix while others don't. Users typically start by following these examples, and this inconsistency can lead to confusion and errors. This PR ensures the system works correctly regardless of which format users choose.

**2. k0s Patch Version Requirements**
The k0s.X suffix (k0s.0, k0s.1, k0s.2, etc.) was originally designed to indicate k0s-specific patches and fixes. Looking at Docker Hub (https://hub.docker.com/r/k0sproject/k0s/tags?name=k0s.1), we can see that k0s.1 and k0s.2 versions are actively released and used. Users need the flexibility to specify these patch versions when necessary.

**3. Security and Best Practices**
If we restrict users to only k0s.0, it would force operations to rely on mutable container tags when new k0s patches are released. This practice is discouraged by security frameworks like CIS Benchmarks and increases the risk of supply chain attacks. Users should be able to pin specific k0s patch versions for security and stability.

**4. Future Compatibility**
What happens when a critical k0s-specific patch (k0s.1, k0s.2) needs to be deployed? Users would be unable to update without modifying their entire version strategy.

This PR provides a robust solution that:
- Maintains backward compatibility
- Works with any version format users choose
- Preserves the full k0s version information in `status.k0sVersion`
- Ensures Cluster API compatibility through proper version formatting

I think this approach offers the flexibility and reliability needed for various use cases, allowing users to choose the version format that best suits their requirements.

I'm happy to discuss this further or make any adjustments to the implementation if you have any concerns or suggestions.

apedriza:
What I understand is that the main issue here is that we currently have two different ways to declare a k0s version, due to the format accepted by Docker Hub.

For a better user experience, I would suggest we simplify this by accepting only one format (using `+`), and handling any required conversion at runtime when creating the StatefulSet. At the end, this is an internal technicality related to Docker Hub, and it should remain transparent to the user. To preserve backward compatibility with cases using `-`, we could add a warning in the validation webhook notifying about future deprecation. There should be no problem with k0s specific versions.

On the other hand, we should clearly document that machine preflight checks expect the same version format used by k0smotron, for example: `v1.31.1+k0s.1` So, whenever a user defines `MachineDeployment.spec.version` or `Cluster.spec.topology.version`, it should match the k0smotron format.

IMO adding multiple version fields would only increase maintenance overhead and make things harder to understand but maybe Im missing something, sorry for the back and forth @kahirokunn.

@makhov, @jnummelin what do you think?

broboa:
@apedriza are you sure it is exactly `cluster.spec.version` and not `cluster.spec.topology.version`? It seems that `cluster.spec.version` doesnt exist.

apedriza:
> @apedriza are you sure it is exactly `cluster.spec.version` and not `cluster.spec.topology.version`? It seems that `cluster.spec.version` doesnt exist.

yes `cluster.spec.topology.version` sorry

broboa:
@apedriza defining `cluster.spec.topology` requires both `version` and `class`. A ClusterClass seems like a big thing to define, or? https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-class/write-clusterclass

apedriza:
> @apedriza defining `cluster.spec.topology` requires both `version` and `class`. A ClusterClass seems like a big thing to define, or? https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-class/write-clusterclass

@broboa if you don't need a clusterclass you only need use format with `<k8s-version>-<k0s-version>` in the `K0smotronControlPlane.spec.version` to avoid the machineset preflight being stuck

kahirokunn:
@apedriza 

Thank you for your feedback. I'd like to clarify my understanding and align on the path forward.

## My Initial Understanding

I initially understood this PR was addressing the **presence/absence of the suffix** issue:
- `v1.31.1` (shortened form without suffix)
- `v1.31.1-k0s.0` (full form with suffix)

This mismatch was causing the Cluster API compatibility issue where MachineDeployment scaling would hang.

## Your Actual Proposal (if I understand correctly)

After re-reading your comment, I now understand you're proposing to address a different aspect - the **symbol standardization** between:
- `v1.31.1-k0s.0` (hyphen format for Docker Hub)
- `v1.31.1+k0s.0` (plus format as Kubernetes standard)

Your suggestion is to:
1. Accept only the `+` format from users (e.g., `v1.31.1+k0s.0`)
2. Automatically convert to `-` format internally when creating StatefulSets for Docker Hub
3. Make this Docker Hub technicality transparent to users

I wasn't aware that container tags don't allow `+` symbols. I found that other projects like Helm also auto-convert `+` to `-` (https://github.com/opencontainers/distribution-spec/issues/154#issuecomment-2978081486), so I agree with this automatic conversion approach.

## Moving Forward

While the `+` to `-` conversion is indeed a valuable improvement, this PR currently solves a concrete problem: **the version format mismatch that prevents MachineDeployment from scaling**. This is affecting users today, as reported in #1052 and #1027.

Is there anything specific that needs to be addressed in this PR to move it forward? I'm happy to make any necessary adjustments.

apedriza:
> this PR currently solves a concrete problem: **the version format mismatch that prevents MachineDeployment from scaling**. This is affecting users today, as reported in #1052 and #1027.

@kahirokunn have you try to use `<k8s-version>-<k0s-version>` format for the `K0smotronControlPlane.spec.version` ? that should avoid the failing comparison between `.spec.version` and `.status.version` and make the preflight check succeed. Probably is failing is your case because you are using a plain k8s version format for the K0smotronControlPlane. Can you confirm?

kahirokunn:
@apedriza Thank you very much for your kind response! 🙏 
However, I am not seeking a temporary workaround. Since k0smotron is a ClusterAPIProvider, I created this PR with the intention of addressing the parts that do not comply with the ClusterAPI specification, making fundamental fixes, and making many people happy. 😆

kahirokunn:
Many of the current issues are due to non-compliance with ClusterAPI KEPs.
Compliance with ClusterAPI KEPs will resolve many issues and enable integration with more providers.
And many people will be happy. 👍

apedriza:
> Many of the current issues are due to non-compliance with ClusterAPI KEPs. Compliance with ClusterAPI KEPs will resolve many issues and enable integration with more providers. And many people will be happy. 👍

maybe I'm missing something. Is there any documentation in CAPI that references the version format? As far as I can see the only thing it says 
 > // version defines the desired Kubernetes version for the control plane. 
 > // The value must be a valid semantic version; also if the value provided by the user does not start with the prefix v, it
 > // must be added
 
The k0smotron controlplane provider is always used together with the k0smotron bootstrap provider, so that the version format is always understood between providers. On the other hand, `K0sControlPlane` also uses this format so that is why I think keep using that format (or convert it using the mutating webhook + a deprecation notice) is the simplest change

kahirokunn:
Hi @apedriza,

Thank you for your review. I'd like to provide some context about why this PR is structured this way.

This PR addresses a critical requirement from the Cluster API Control Plane Contract, which states:

> NOTE: The minimum Kubernetes version, and more specifically the API server version, will be used to determine when a control plane is fully upgraded (spec.version == status.version) and for enforcing Kubernetes version skew policies when a Cluster derived from a ClusterClass is managed by the Topology controller.

[Source: CAPI Control Plane Contract](https://github.com/kubernetes-sigs/cluster-api/blob/063edb3b/docs/book/src/developer/providers/contracts/control-plane.md#controlplane-version)

The implementation in Cluster API enforces this requirement here:
https://github.com/kubernetes-sigs/cluster-api/blob/063edb3b/internal/contract/controlplane.go#L205-L208

Currently, K0smotronControlPlane violates this contract because:
- `spec.version` might be `v1.33.1`
- `status.version` returns `v1.33.1-k0s.0`

This mismatch (`spec.version != status.version`) is causing the issues reported in:
- #1052 
- #1027

This PR simply ensures compliance with the CAPI contract by making `status.version` match the format of `spec.version`. The full k0s version information is preserved in the new `status.k0sVersion` field.

The +/- notation concerns you mentioned are a separate issue and not related to this contract compliance fix. This PR focuses solely on implementing the CAPI specification correctly, which will resolve many current integration problems.

I hope this clarifies the rationale behind the implementation. Happy to discuss further if you have any questions!

apedriza:
@kahirokunn yea, the main issue is that CAPI detects the control plane as upgrading whenever `.spec.version` differs from `.status.version`. To avoid this, I suggest we avoid accepting or storing plain Kubernetes version formats. Instead, we should enforce using the k0s-specific format (always with the `+` suffix). Internally, we can transform `-` to `+` before performing comparisons. All the issues should be solved following this approach if we add default k0s suffix for current plain k8s version and store it again, we don't introduce a new field and we unify the accepted format for the version

jnummelin:
@apedriza @kahirokunn I agree with both of you 😂  Let me explain what I'm thinking.

I think we should have some mutating webhook to standardize the version format we take in, like @apedriza is mentioning. This would mean we actually store the version always in the same format "internally".

But even if we'll create something like that say in some coming 1.X.0 version, it still leaves this problem for already existing resources. And for that, we'd need what this PR implements, otherwise those existing cluster will continue to suffer from this since the webhooks only kick-in when users create/modify the resources in the API.

What I'm not sure is the introduction of `K0sVersion` field. As far as I see we don't _really_ use the field for anything. I mean we don't use it internally in k0smotron code for anything nor do we show it to users, other that if they dig deep into the status field. So is this really needed? 🤔

apedriza:
> @apedriza @kahirokunn I agree with both of you 😂 Let me explain what I'm thinking.
> 
> I think we should have some mutating webhook to standardize the version format we take in, like @apedriza is mentioning. This would mean we actually store the version always in the same format "internally".
> 
> But even if we'll create something like that say in some coming 1.X.0 version, it still leaves this problem for already existing resources. And for that, we'd need what this PR implements, otherwise those existing cluster will continue to suffer from this since the webhooks only kick-in when users create/modify the resources in the API.
> 
> What I'm not sure is the introduction of `K0sVersion` field. As far as I see we don't _really_ use the field for anything. I mean we don't use it internally in k0smotron code for anything nor do we show it to users, other that if they dig deep into the status field. So is this really needed? 🤔

yes, we definitely need some kind of transformation for the existing resources as proposed in the PR 💯 . For creating new resources, mutating them if they comes with a  plain k8s versions by adding the k0s suffix should also fix the current bug. On the other hand, I think we should write some warning in the webhook about it would be default to `k0s.0` version

kahirokunn:
The currently released `.status.version` always includes the k0s version like `+k0s.0`, etc. 

So I added `.status.k0sVersion` to keep track of which k0s version it corresponds to, just in case. However, if it's not needed, I can remove it right away. (To be honest, this was just me being proactive, thinking others might want it.)

Is there anything else you'd like me to address?

kahirokunn:
> write some warning in the webhook about it would be default to k0s.0 version

I will check how to do it. 👀

inttest/capi-docker-clusterclass-k0smotron/capi_docker_clusterclass_k0smotron_test.go
```diff
@@ -184,7 +183,6 @@ metadata:
 spec:
   template:
     spec:
-      version: v1.27.2-k0s.0
```
apedriza, line 187: if we set `v1.27.2` as version we can check the e2e works without need new ones. I think is better reuse this tests. wdyt? Sorry for the back and forth @kahirokunn 🙏

internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -468,7 +485,9 @@ func (c *K0smotronController) computeStatus(ctx context.Context, cluster types.N
 	kcp.Status.UnavailableReplicas = int32(unavailableReplicas)
 
 	if kcp.Status.ReadyReplicas > 0 {
-		kcp.Status.Version = minimumVersion.String()
+		statusVersion := minimumVersion.String()
+		// Store the formatted version for Cluster API compatibility
+		kcp.Status.Version = FormatStatusVersion(kcp.Spec.Version, statusVersion)
```
apedriza, line 490: couple lines below, if `status.version` not contain k0s suffix, will requeue always because `desiredVersion` always contains k0s suffix

inttest/capi-docker-clusterclass-k0smotron/capi_docker_clusterclass_k0smotron_test.go
kahirokunn: Thx. I did it.

internal/controller/controlplane/k0smotron_controlplane_controller.go
kahirokunn: Thx. I did it.

kahirokunn:
I have addressed all requests and added validating webhooks.

CleanShot 2025-07-25 at 13 49 23@2x

```yaml
kind: K0smotronControlPlane
metadata:
  name: docker-test-cp
  namespace: default
spec:
  version: v1.33.2
  etcd:
    resources:
      requests:
        cpu: 100m
        memory: 128Mi
      limits:
        cpu: 200m
        memory: 256Mi
  persistence:
    type: pvc
    persistentVolumeClaim:
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 50Mi
  service:
    type: LoadBalancer
---
kind: K0sControlPlane
metadata:
  name: docker-test
spec:
  replicas: 1
  version: v1.30.0
  updateStrategy: Recreate
  k0sConfigSpec:
    args:
      - --enable-worker
    k0s:
      apiVersion: k0s.k0sproject.io/v1beta1
      kind: ClusterConfig
      metadata:
        name: k0s
      spec:
        api:
          extraArgs:
            anonymous-auth: "true"
        telemetry:
          enabled: false
  machineTemplate:
    infrastructureRef:
      kind: DockerMachineTemplate
      name: docker-test-cp-template
      namespace: default
```

Is there anything else we need to do?
I would appreciate your confirmation.
Thank you for your cooperation.

inttest/util/util.go
```diff
@@ -437,3 +439,64 @@ func dowloadStableK0smotronOperator() (string, error) {
 	}
 	return installFile.Name(), nil
 }
+
+// GetK0smotronControlPlaneStatus retrieves the K0smotronControlPlane status
+func GetK0smotronControlPlaneStatus(ctx context.Context, client *kubernetes.Clientset, name, namespace string) (*cpv1beta1.K0smotronControlPlaneStatus, error) {
```
makhov: the changes in this file is not used, let's delete them then

internal/controller/controlplane/k0smotron_controlplane_webhook.go
```diff
@@ -0,0 +1,111 @@
...
+// as this struct is used only for temporary operations and does not need to be deeply copied.
+type K0smotronControlPlaneValidator struct{}
+
+var _ webhook.CustomValidator = &K0smotronControlPlaneValidator{}
+
+// validateVersionSuffix checks if the version has a k0s suffix and returns a warning if it doesn't
+func (v *K0smotronControlPlaneValidator) validateVersionSuffix(version string) admission.Warnings {
+	warnings := admission.Warnings{}
+	if version != "" && !strings.Contains(version, "-k0s.") {
+		warnings = append(warnings, fmt.Sprintf("The specified version '%s' requires a k0s suffix (-k0s.<number>). Using '%s-k0s.0' instead.", version, version))
```
makhov, line 47: Is it real warnings or it will end up as errors? I don't want to change any user-faced behavior in the PR

kahirokunn: This is a warning. User behavior will not change. Please refer to the image below.
https://github.com/k0sproject/k0smotron/pull/1053#issuecomment-3116381776
image

inttest/util/util.go
kahirokunn: While working on many tasks, I left some unused code behind. I deleted it. Thank you 🙏

internal/controller/controlplane/k0smotron_controlplane_webhook.go
kahirokunn: CleanShot 2025-07-25 at 19 10 28@2x

```yaml
kind: Cluster
metadata:
  name: docker-test
  namespace: default
spec:
  controlPlaneRef:
    kind: K0smotronControlPlane
    name: docker-test-cp
    namespace: default
  infrastructureRef:
    kind: DockerCluster
    name: docker-test
    namespace: default
---
kind: K0smotronControlPlane
metadata:
  name: docker-test-cp
  namespace: default
spec:
  version: v1.33.2
  etcd:
    resources:
      requests:
        cpu: 100m
        memory: 128Mi
      limits:
        cpu: 200m
        memory: 256Mi
  persistence:
    type: pvc
    persistentVolumeClaim:
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 50Mi
  service:
    type: LoadBalancer
---
kind: DockerCluster
metadata:
  name: docker-test
  namespace: default
  annotations:
    cluster.x-k8s.io/managed-by: k0smotron # This marks the base infra to be self managed. The value of the annotation is irrelevant, as long as there is a value.
spec: {}
  # More details of the DockerCluster can be set here
---
kind: MachineDeployment
metadata:
  name:  docker-test-md
  namespace: default
spec:
  clusterName: docker-test
  replicas: 1
  selector:
    matchLabels:
      pool: worker-pool-1
  template:
    metadata:
      labels:
        pool: worker-pool-1
    spec:
      clusterName: docker-test
      version: v1.33.2 # Docker Provider requires a version to be set (see https://hub.docker.com/r/kindest/node/tags)
      bootstrap:
        configRef:
          kind: K0sWorkerConfigTemplate
          name: docker-test-machine-config
      infrastructureRef:
        kind: DockerMachineTemplate
        name: docker-test-mt
---
kind: DockerMachineTemplate
metadata:
  name: docker-test-mt
  namespace: default
spec:
  template:
    spec: {}
  # More details of the DockerMachineTemplate can be set here
---
kind: K0sWorkerConfigTemplate
metadata:
  name: docker-test-machine-config
spec:
  template:
    spec:
      version: v1.33.2
      # More details of the worker configuration can be set here
```

👍

kahirokunn:
I have responded to all reviews and confirmed that everything is working properly.
I would appreciate it if you could review us again. 🙏

kahirokunn:
Thank you for your reviews and follow-ups😆

PR #1060, opened 2025-06-09
k0smotron Cluster mutating webhook

makhov:
Mutating webhook that sets default values to the `k0smoton.Cluster`

### DIFF
--- a/inttest/Makefile
+++ b/inttest/Makefile
@@ -5,7 +5,7 @@ K0SMOTRON_IMAGES_BUNDLE ?= $(realpath ../k0smotron-image-bundle.tar)
 K0SMOTRON_INSTALL_YAML ?= $(realpath ../install.yaml)
 LOCAL_STORAGE_INSTALL_YAML ?= $(realpath ./footloose-alpine/local-path-storage.yaml)
 CERT_MANAGER_INSTALL_YAML ?= $(realpath ./footloose-alpine/cert-manager.yaml)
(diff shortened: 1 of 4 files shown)


inttest/Makefile
```diff
@@ -5,6 +5,7 @@ K0SMOTRON_IMAGES_BUNDLE ?= $(realpath ../k0smotron-image-bundle.tar)
 K0SMOTRON_INSTALL_YAML ?= $(realpath ../install.yaml)
 LOCAL_STORAGE_INSTALL_YAML ?= $(realpath ./footloose-alpine/local-path-storage.yaml)
 CERT_MANAGER_INSTALL_YAML ?= $(realpath ./footloose-alpine/cert-manager.yaml)
+WEBHOOK_CHECKEr_INSTALL_YAML ?= $(realpath ./footloose-alpine/webhook-checker.yaml)
```
apedriza:
```suggestion
WEBHOOK_CHECKER_INSTALL_YAML ?= $(realpath ./footloose-alpine/webhook-checker.yaml)
```

apedriza (commented):
there are kubebuilder default annotations for some fields like prometheus image, etcd image.. Should be moved all that defaults? I think webhook is always deploy if cluster controller is deploy so..

inttest/Makefile
makhov: Great catch, thanks! Fixed!

makhov:
I've added some other default fields, thanks!

apedriza (approved):

PR #1067, opened 2025-06-18
Change the call order in the webhook initialization

makhov:
Not sure why it matters, but previous version didn't work locally


apedriza:
is this still needed anymore? Im able to run both webhooks locally using development version.

PR #1068, opened 2025-06-18
Use plain kubernetes version in K0smotronControlPlane status

### LINKED ISSUE #1027: wrong `version` field in status of `k0smotronControlPlane`

makhov:
Fixes #1027 

Put plain kubernetes version in K0smotronControlPlane status instead of docker complaint k0s version that is treated as prerelease version


kahirokunn:
Dear @makhov ,
I am always grateful for the reviews you provide for me.

Thank you also for creating PR #1068!
I feel very reassured that you are already addressing the CAPI issue regarding "matching spec.version and status.version."

However, I believe that the current changes in PR #1068 do not yet fully cover *cases where `spec.version` has the `-k0s.N` suffix*, and this would cause CAPI's `IsUpgrading()` to return true again, leaving the issue unresolved.

I believe there is actually a conflict with **PR #1053** that I created two weeks ago.
**PR #1053** covers the following:

1. Handles both cases where `spec.version` has or doesn't have a suffix
2. Always normalizes `status.version` to the same format
3. Preserves the original full notation in `status.k0sVersion`

The implementation follows this flow, and **e2e tests have already confirmed that "upgrade determination converges for both cases with and without suffixes."**
I have also updated the documentation and CRD accordingly.

I have addressed jnummelin's review comments and understand it is currently awaiting merge.

Therefore, I apologize for the request, but would it be possible for you to **consider PR #1053 instead?**
Multiple test patterns have already been verified, which should reduce review effort. In that PR, I have completed all the work comprehensively to support all version formats, and after reviewing CAPI's specifications and implementation code, I am confident it is fully compatible.

I apologize for the inconvenience during your busy schedule, but I would greatly appreciate your review. Thank you very much!

Best regards,
@kahirokunn

makhov:
Yeah, sorry, I've created the PR and realized that we have yours already. So I'm closing this one in favor of #1053

PR #1082, opened 2025-06-26
docs: refactor custom resources generation

apedriza:
iOS safari cannot display custom resources page, probably because the current size of it. With this change each resource is in an individual page. This change has also caused the refactoring of the crds generation.


jnummelin (approved):
LGTM

PR #1085, opened 2025-06-30
feat(etcd): add advanced scheduling and runtime configuration options

kahirokunn:
Add support for etcd-specific pod scheduling and runtime configurations to provide fine-grained control over etcd pod placement and behavior.

Changes include:
- Add RuntimeClass field for container runtime selection
- Add Tolerations field for node taint handling
- Add TopologySpreadConstraints field for pod distribution control
- Add NodeSelector field for node targeting
- Add Affinity field for custom affinity/anti-affinity rules
- Add ImagePullSecrets field for private registry access
- Add PriorityClassName field for pod priority scheduling

These etcd-specific settings take precedence over cluster-level configurations when specified, with automatic fallback to cluster-level settings when etcd-specific ones are not provided.

The implementation includes:
- Updated EtcdSpec type definition with new fields
- Generated deepcopy methods for proper object copying
- Updated CRD definitions with comprehensive field validation
- Enhanced etcd StatefulSet generation logic to apply settings
- Comprehensive test coverage for all new functionality

This enables users to optimize etcd deployment for specific infrastructure requirements while maintaining backward compatibility.


kahirokunn:
@apedriza I would appreciate it if you could review this PR when you have time.
Thank you 🙏

internal/controller/k0smotron.io/k0smotroncluster_etcd_test.go
```diff
@@ -119,6 +121,372 @@ func TestEtcd_resourceRequirements(t *testing.T) {
...
+								WhenUnsatisfiable: corev1.DoNotSchedule,
+								LabelSelector: &metav1.LabelSelector{
+									MatchLabels: map[string]string{"app": "etcd"},
+								},
+							},
+						},
+					},
+				},
+			},
+			want: func(t *testing.T, podSpec *corev1.PodSpec) {
```
apedriza: we use `require` for assertions.

IMO is more transparent to move assertions to the test body and configure values in each test case than encapsulate validation in each test case. I could write a null `want` fn and still pass

apedriza (commented):
@makhov wdyt about this api extension?

kahirokunn:
@apedriza Thank you for raising an excellent point regarding the alignment with k0s configuration patterns.

I would be pleased to share an interesting use case that we have been implementing: utilizing runtimeClass not only for container runtime selection, but also as an alias for scheduling configurations (including nodeSelector, tolerations, and other related settings).

The primary advantage of this approach is the effective separation of application specifications from infrastructure scheduling concerns.

internal/controller/k0smotron.io/k0smotroncluster_etcd_test.go
kahirokunn: Thank you. I have addressed the issue as you pointed out. How does that look? 👀

kahirokunn:
I rebased with the latest main.

kahirokunn:
Bump 👀

makhov:
Thanks for the PR! Looking on the given changes, I start thinking, should we embed the whole pod spec instead of just some fields? I may imagine that some users might want to change another field. WDYT?

kahirokunn:
Thanks for the suggestion—embedding the entire pod spec makes sense and will give users the flexibility they need. I’m on board! 🙏

kahirokunn:
keep

kahirokunn:
Keep

kahirokunn:
keep

kahirokunn:
Keep

kahirokunn:
> Thanks for the PR! Looking on the given changes, I start thinking, should we embed the whole pod spec instead of just some fields? I may imagine that some users might want to change another field. WDYT?

@makhov Thank you for your review. 🙏 
I have implemented according to your review instructions and conducted thorough manual testing locally with various scenarios. Everything is working correctly. Would you be able to review it again when you have a chance? 🙏

kahirokunn:
I've updated the PR title and description to reflect this as well. There are various records and verifiable code available, so please take a look when you get a chance. 🙏

Issue #1087, opened 2025-07-01
Bug: ControlPlane .status.version not set to empty string during initialization (CAPI contract violation)


kahirokunn:
### What happened?

k0smotron controlplane does not set `.status.version` to an empty string during the period before the ControlPlane becomes ready. Currently, k0smotron sets `.status.version` to either null or the same value as `.spec.version` during initialization, which violates the CAPI (Cluster API) contract requirements.

According to the CAPI implementation specification, during the period until the ControlPlane is created and ready, `.status.version` must be an empty string, not null.

Reference: https://github.com/kubernetes-sigs/cluster-api/blob/063edb3b/internal/contract/controlplane.go#L185-L203

### Involved Areas

Cluster API integration

### What did you expect to happen?

The k0smotron ControlPlane implementation should comply with the CAPI contract by:

Setting `.status.version` to an empty string (`""`) during the initialization period before the ControlPlane becomes ready

### Steps to reproduce

1. Create a new k0smotron ControlPlane resource using CAPI
2. Observe the `.status.version` field during the initialization phase before the ControlPlane becomes ready
3. Verify that `.status.version` is set to null or contains the same value as `.spec.version` instead of being an empty string
4. Compare this behavior with the CAPI contract requirements at https://github.com/kubernetes-sigs/cluster-api/blob/063edb3b/internal/contract/controlplane.go#L185-L203

Expected observation: `.status.version` should be `""` (empty string) during initialization
Actual observation: `.status.version` is either null during initialization

### k0smotron version

v1.5.4

### k0s version

v1.33.2-k0s.0

apedriza:
can we close this @kahirokunn ?

kahirokunn:
This also needs to be merged🙇‍♂️
https://github.com/k0sproject/k0smotron/pull/1088

PR #1088, opened 2025-07-01
feat: Set default empty string for K0smotronControlPlane status.version to comply with CAPI

kahirokunn:
## Summary

This PR ensures that `K0smotronControlPlane.status.version` is initialized as an empty string instead of `null` to comply with Cluster API (CAPI) implementation requirements.

## Background

According to the CAPI specification, the `status.version` field of ControlPlane resources must be an empty string (not `null` or `spec.version`) during the period between ControlPlane creation and when it becomes fully ready. This is a strict requirement for CAPI compliance and proper interaction with CAPI controllers.

## Changes

- Added `// +kubebuilder:default:={version:""}` annotation to the `Status` field in `K0smotronControlPlane` struct
- This ensures that when a `K0smotronControlPlane` resource is created, the `status.version` field is automatically initialized as an empty string

## Impact

- **CAPI Compliance**: Ensures proper compliance with CAPI specification requirements
- **Controller Behavior**: Prevents potential issues with CAPI controllers that expect `status.version` to be a string type rather than `null`
- **Initialization**: Provides consistent initialization behavior across different CAPI implementations

## Testing

- [x] CRD generation works correctly with the new default annotation
- [x] New `K0smotronControlPlane` resources are created with `status.version: ""`
- [x] Existing functionality remains unaffected

## Breaking Changes

None. This change only affects the initial state of newly created resources and does not modify existing behavior.

Fixes: #1087


kahirokunn:
@apedriza Hello, would you mind reviewing this? Thank you 🙏

makhov:
The PR looks very similar to #1090. Can we change K0smotronControlPlane.status to comply with CAPI in one PR?

kahirokunn:
> The PR looks very similar to #1090. Can we change K0smotronControlPlane.status to comply with CAPI in one PR?

Okay 👍

PR #1094, opened 2025-07-02
Add generated deepcopy files

kahirokunn:
zz_generated.deepcopy.go was missing, so I restored it.


apedriza (approved):
Thanks! I deleted them by mistake in the last commit

PR #1099, opened 2025-07-04
Check for the controller config change with etcd node name set

makhov:
Fixes https://github.com/k0rdent/kcm/issues/1709

At some point we were putting the things that we generate for internal use to the object. Now we don't do it, but upgrading k0smotron triggers machine recreation, it the old version was doing that. 

The PR adds a small workaround to prevent it.


apedriza:
I wonder if in the future we should evaluate whether it's better to modify the watched resource directly, instead of relying on internal variables during the reconcile process, to avoid this kind of unexpected behavior. For example, we could use an internal variable to store the k0s config and its additions, rather than relying on the K0sControllerConfig resource.

But we can address it later, it is not blocking from my review.

makhov:
Yeah, we need to build smarter solution for this eventually. Probably we can calculate a hash of meaningful fields and store them in K0sControllerConfig's annotations. And then just compare hashes. Not sure that in-memory vars will be reliable in such a dynamic environment.

apedriza:
Thinking about this change, why every time there was a machine reconcile, the previous active machines were not recreated if their k0scontroller config had been modified?

internal/controller/controlplane/helper.go
```diff
@@ -292,17 +292,28 @@ func (c *K0sController) hasControllerConfigChanged(bootstrapConfigs map[string]b
...
+	// Bootstrap controller did set etcd name to the K0sControllerConfig, so we need to compare it with the name set in the K0sControlPlane
+	kcpStorageConfigEtcdWithName, _, _ := unstructured.NestedMap(kcp.Spec.K0sConfigSpec.K0s.Object, "spec", "storage")
+	_ = unstructured.SetNestedField(kcpStorageConfigEtcdWithName, machine.Name, "etcd", "extraArgs", "name")
 	bootstrapConfig.Spec.K0sConfigSpec.K0s = kcp.Spec.K0sConfigSpec.K0s
 	// leave out the tunneling spec for the bootstrap config
 	bootstrapConfig.Spec.K0sConfigSpec.Tunneling = kcp.Spec.K0sConfigSpec.Tunneling
 	return !reflect.DeepEqual(kcp.Spec.K0sConfigSpec, *bootstrapConfig.Spec.K0sConfigSpec) ||
 		!reflect.DeepEqual(kcpAPIConfig, bootstrapAPIConfig) ||
-		!reflect.DeepEqual(kcpStorageConfig, bootstrapStorageConfig)
+		(!reflect.DeepEqual(kcpStorageConfig, bootstrapStorageConfig) && !reflect.DeepEqual(kcpStorageConfigEtcdWithName, bootstrapStorageConfig))
```
zerospiel, line 314: probably nit-picking: would it be beneficial to separate these comparisons and put them into debug logs to ease the debugging process (at least in that particular case)?

and a follow-up question: should we consider adding some (golden?) unit tests? Or is it not worth it since there is no shared code between the boostrap-controller and the cp-controller?

makhov:
@apedriza because setting those fields was introduced by the patchHelper change, which was fixed later. And recreation removes old K0sControllerConfig and creates a new one.

internal/controller/controlplane/helper.go
makhov: Yep, but it's temporary fix, we will come up with the proper solution eventually

makhov:
@apedriza All tests have passed. 
The fix also covers the following case since I removed `break` from the loop.
```
- --config
- /etc/k0s.yaml
- --config
- /etc/k0s.yaml
```

apedriza:
I'm testing this fix. I'll try to update our e2e upgrade test covering this case

apedriza:
I have updated our current e2e tests for k0smotron upgrades to cover this case.

apedriza (approved):

PR #1112, opened 2025-07-16
Change machine naming strategy

### LINKED ISSUE #1022: Change machine naming strategy

Currently, when a K0sControlPlane controller creates Machine resources, they are named using sequential numeric suffixes. This behavior introduces several practical limitations:

- Error-prone maintenance.

- When a Machine is deleted or replaced, the controller typically reuses the same name (e.g., recreates my-cluster-1), which can cause name collisions or reconciliation issues if the previous object is still terminating or being cleaned up.

- Sequential names imply stable identity, which is misleading unless identity persistence is required. Like `Cluster` or `K0sControlPlane`. 

- It adds complexity in highly dynamic clusters where Machines are frequently created and deleted because name reuse makes it harder to track history. For instance, name reuse makes it harder to trace resource history or correlate them with events. 

- It deviates from the naming pattern used by Kubernetes for non-identity-based resources (Pods, Jobs, etc.), owned by a higher-level resource (k0scontrolplane in this case) and the pattern used by CAPI providers.

add a random suffix in the machine name

_No response_

### Additional context

This change involves changing the tests that are based on the machine name.

_No response_

apedriza:
Changing machine naming should not create issues when scaling machines

fix #1022


internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -973,26 +977,15 @@ func (c *K0sController) createFRPToken(ctx context.Context, cluster *clusterv1.C
...
-	if len(machineToDelete) == 0 {
-		for i := 0; i < int(kcp.Spec.Replicas); i++ {
-			name := fmt.Sprintf("%s-%d", kcp.Name, len(desiredMachines)-i)
-			_, ok := desiredMachines[name]
-			if !ok {
-				return name
-			}
-		}
-	}
+func machineName(kcp *cpv1beta1.K0sControlPlane) string {
```
makhov: I think we can get rid of this func, `names.SimpleNameGenerator.GenerateName` will do it for you

apedriza: nice, thanks!

apedriza:
I will keep this PR as a draft because there are some issues that seems to break the cluster when rolling out the controlplane

apedriza:
> I will keep this PR as a draft because there are some issues that seems to break the cluster when rolling out the controlplane

This was a bad test update. Ready for review now! 👍

e2e/k0smotron_upgrade_test.go
```diff
@@ -49,7 +49,7 @@ func TestK0smotronUpgrade(t *testing.T) {
 // Important: This version MUST match the ones set in the e2e config file because that guarantees that the clusterctl
 // local repository contains the required versions of the providers when running the upgrade tests. We should test
 // development version against the latests stable version of k0smotron.
-var k0smotronMinorVersionsToCheckUpgrades = []string{"1.5"}
+var k0smotronMinorVersionsToCheckUpgrades = []string{"1.5", "1.6"}
```
apedriza, line 52: Not really related, but this change deserves to be included in the upgrade test

inttest/capi-docker-machine-template-update-recreate/__debug_bin1768403602
makhov, line 1: Is this file needed in the PR?

apedriza: 😯 removed and added debug files to gitignore. Good catch!

PR #1139, opened 2025-08-08
Switch default registry from docker.io to mirror.gcr.io to avoid docker.io’s strict rate limits by default

kahirokunn:
**Summary**
This PR changes our default container registry from **`docker.io`** to **`mirror.gcr.io`**. By defaulting to Google’s public mirror we side-step Docker Hub’s pull-rate restrictions, reducing the risk of unexpected `429 Too Many Requests` errors in local development, CI pipelines, and ephemeral environments.

**Background**
Docker Hub currently throttles anonymous pulls to **100 pulls / 6 hours** per IP and free-tier authenticated pulls to **200 pulls / 6 hours**. Paid (Pro/Team/Business) accounts remain unlimited. These limits frequently surface in shared runners and NAT’d build hosts.

**Change Details**

* Update the default registry constant from `docker.io` → `mirror.gcr.io`.


makhov:
Thanks for the PR! It's totally a good idea, but we don't want to rely on the Google mirror. We use quay.io for all our images and we work on adding k0s image there as well. I expect the image repo to be `quay.io/k0sproject/k0s`.

kahirokunn:
okay， I will change to it.

kahirokunn:
I fixed. 🙏

makhov:
Thanks! The test is failing for a reason, we need to change the image there as well

kahirokunn:
Thank you! Found and corrected a test change omission.

makhov (approved):

Issue #1143, opened 2025-08-11
move controller config defaults to mutating webhook


apedriza:
### Is your feature request related to a problem? Please describe.

currently, default are set on the controller logic so changes in this lines can update the controller config resource, for example during a controller upgrade, and trigger unexpected rollout in the control plane controller as we faced recently. 

### Describe the solution you would like

If we move defaults to the mutating webhook, only explicit actions from the user like create/update the resource will trigger the rollout mechanism (if needed). On the other hand, it seems to be a better place to locate this default logic

PR #1147, opened 2025-08-11
add probe timeout of 5 seconds

### LINKED ISSUE #1141: CAPI controlplane fails to become 'ready' when 'k0s status' takes more than 1 second
### What happened?

While setting up a cluster using k0smontron as the control plane and bootstrap provider, I found that in certain management clusters, the control plane would never hit a 'ready' status due to the ready & health checks hitting a 'timeout'.

it appears that when the controlplane statefulset is generated, since a timeout is not specified in [k0smotroncluster_statefulset.go near line 145](https://github.com/k0sproject/k0smotron/blob/main/internal/controller/k0smotron.io/k0smotroncluster_statefulset.go#L145) for both the ready check and health check, it uses the kubernetes default value of 1 [kubernetes readiness probe doc](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) and if the 'k0s status' takes longer than 1 second to return, it is considered unhealthy/not-ready

I was able to exec into the control plane pod and perform  

`time k0s status` and received the following output

```
Version: v1.30.1+k0s.0
Process ID: 27
Role: controller
Workloads: false
SingleNode: false
real    0m 2.17s
user    0m 0.11s
sys     0m 2.11s
```

Can a 'timeoutSeconds' of 5 seconds (less than the PeriodSeconds of 10) be defined in the the liveness/readiness probes in k0smotroncluster_statefulset.go so that there is a bit more grace in the health/readiness checks as the command is indeed executing, but just taking a bit longer than 1 second.

Another option would be to expose the health/readiness values as part of the K0smotronControlPlane CRD so that users can set values that work in their environments

### Involved Areas

Cluster API integration

controlplane to come up as ready

### Steps to reproduce

unsure how to reproduce as on some kubernetes clusters it executes under 1 second

### k0smotron version

v1.6.2

### k0s version

v1.30.1+k0s.0

_No response_

brinker-tbaker:
this PR adds a add probe timeout of 5 seconds up from the kubernetes default of 1 second
Fixes #1141


jnummelin (approved):

makhov:
Something is off with commits. Also, the test fails because we change the StatefulSet which causes pod to be recreated. I'll think about how to deal with that

apedriza:
@brinker-tbaker can you rebase latest changes?

brinker-tbaker:
rebased and same test failure. @apedriza i see you opened #1143, was that related to this?

apedriza:
> rebased and same test failure. @apedriza i see you opened #1143, was that related to this?

sorry @brinker-tbaker totally missed your comment. That issue you pointed out was related to a different topic related to refactor logic to controlplane config defaults

PR #1148, opened 2025-08-12
Fix golang.org/x/oauth2 and golang.org/x/crypto HIGH vulnerabilities

apedriza:
While accidentally pushing to origin, I noticed server-side 'remote:' messages from GitHub about vulnerabilities. Not critical, but I think it's worth bumping them. The remaining HIGH vulnerability related to Helm will be fixed in a follow-up PR,
since updating Helm pulls quite a few transitive changes.

https://github.com/k0sproject/k0smotron/security/dependabot/49
https://github.com/k0sproject/k0smotron/security/dependabot/48
https://github.com/k0sproject/k0smotron/security/dependabot/40


apedriza:
obsolete

PR #1156, opened 2025-08-14
Exclude some fields in STS hash calculation

makhov:
To avoid unnecessary rollouts

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_statefulset.go
@@ -142,18 +142,6 @@ func (r *ClusterReconciler) generateStatefulSet(kmc *km.Cluster) (apps.StatefulS
 							},
 						}},
 						Resources: kmc.Spec.Resources,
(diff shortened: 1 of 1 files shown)


apedriza:
sts behavior for pod templates changes is to rollout so trigger arollout for new changes on that maybe is not that bad. Maybe open the api to overwrite our current probes can avoid the rollout and if user set values different that defaults (current used) rollout is triggered. wdyt?

apedriza:
> sts behavior for pod templates changes is to rollout so trigger arollout for new changes on that maybe is not that bad. Maybe open the api to overwrite our current probes can avoid the rollout and if user set values different that defaults (current used) rollout is triggered. wdyt?

Im thinking about the case where timeout needs to increase. With this changes, the pod is not recreated

makhov:
Yeah, but the intention was to keep there "non-important" fields to avoid rollouts. If everything is working I don't need to change the timeout. And there is an option to trigger it manually by removing the annotation.

I'm not sure that it's the best solution, but adding new config param also doesn't feel right

PR #1157, opened 2025-08-14
Deduplicate k0s args in comparison

makhov:
Finally fixes the issue when duplicated args in K0sControlPlane causes machine recreation

### DIFF
--- a/internal/controller/controlplane/helper.go
+++ b/internal/controller/controlplane/helper.go
@@ -293,27 +293,40 @@ func (c *K0sController) hasControllerConfigChanged(bootstrapConfigs map[string]b
 	}
 
 	kcpK0sConfigSpecCopy := kcp.Spec.K0sConfigSpec.DeepCopy()
+	bootstrapConfigCopy := bootstrapConfig.DeepCopy()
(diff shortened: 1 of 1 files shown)


apedriza:
A few changes have been around this logic lately so I'm a bit confused right now 😅  @makhov can you give me an idea about what wasn't working with the latest `uniqueArgs` logic?

makhov:
Yeah, it was a bit tricky. The previous logic still re-triggers machine recreation once, because it was modifying K0sControlPlane. I've "fixed" it in #1132, which brought the initial issue back. So this PR covers all edge cases.
We've tested it manually with the k0rdent folks, it works now as expected.

apedriza (approved):
I think it's a good idea to work with copies 👍 

On the other hand, @makhov  I wonder if we should update the e2e upgrade test. I recently added some lines covering the unexpected rollout but it seems last regression should have caused the test to fail and it didn't.

makhov:
Could you point out to those changes? I'll update the test in a separate PR.

apedriza:
> Could you point out to those changes? I'll update the test in a separate PR.

This are the lines -> https://github.com/k0sproject/k0smotron/blob/main/e2e/k0smotron_upgrade_test.go#L221-L232. I added them for `v1.6.0` iirc

PR #1161, opened 2025-08-15
Push each PR image to ttl.sh with the commit SHA

jnummelin:
In PR build we already built the image but never pushed it to anywhere. This PR adds pushing an image to ttl.sh with the commit SHA. The image will be named as: `ttl.sh/k0smotron-<commit SHA>`

This enables anyone to easily test any PR image outside of the CI.


apedriza:
I think this uses default time limit which is `1h`. Perhaps we could expand it to have a little more margin when pulling the image. Maximum is `1d`

jnummelin:
ttl.sh says:
> The default is 24 hours

apedriza:
> ttl.sh says:
> 
> > The default is 24 hours

yeah, I misread it then

PR #1162, opened 2025-08-18
refactor(ci): reduce steps duplication

apedriza:
Some CI logic was duplicated—for example, building the k0smotron binary and running unit tests across multiple jobs/steps. This PR adopts a “run it only once” approach so each CI stage executes a single time and its outputs are reused, reducing overall workflow duration.

### DIFF
--- a/hack/capi-ci/config.yaml
+++ b/hack/capi-ci/config.yaml
@@ -1,10 +1,10 @@
 providers:
   - name: "k0sproject-k0smotron"
-    url: "file://%pwd%/k0sproject-k0smotron/bootstrap-k0sproject-k0smotron/v0.0.0/bootstrap-components.yaml"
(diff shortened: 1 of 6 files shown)


hack/capi-ci/config.yaml
```diff
@@ -1,10 +1,10 @@
 providers:
   - name: "k0sproject-k0smotron"
-    url: "file://%pwd%/k0sproject-k0smotron/bootstrap-k0sproject-k0smotron/v0.0.0/bootstrap-components.yaml"
+    url: "file://${CLUSTERCTL_REPOSITORY_PATH}/k0sproject-k0smotron/bootstrap-k0sproject-k0smotron/v0.0.0/bootstrap-components.yaml"
```
apedriza, line 3: when generating `config.yaml` artifact once in a different runner where it is consumed needs to allow setting the path from the runner that uses it (GitHub-hosted vs self-hosted runner has different paths)

Issue #1164, opened 2025-08-20
Want to improve safety and readability of multi-cluster operations by generating kubeconfig with cluster names (e.g., `<cluster-name>-k0s` / `<cluster-name>-admin`)


kahirokunn:
### Is your feature request related to a problem? Please describe.

The kubeconfig generated by k0smotron has fixed names for `cluster`/`context`/`user` (`k0s` / `admin`) as shown below, making it difficult for humans to identify which cluster's kubeconfig it is. As a result, there's an increased risk of human error, such as thinking you're operating on `wl1` when you're actually on `wl2`.

Current example:

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: xxx
    server: https://k8s-default-xxx-xxx-xxx.elb.ap-northeast-1.amazonaws.com:6443
  name: k0s
contexts:
- context:
    cluster: k0s
    user: admin
  name: k0s
current-context: k0s
kind: Config
preferences: {}
users:
- name: admin
  user:
    client-certificate-data: xxx
    client-key-data: xxx
```

Screenshot (readability issue illustration):
![kubeconfig readability](https://github.com/user-attachments/assets/96c276b5-8afa-43a7-adaf-7d8222810549)

### Describe the solution you would like

I would like k0smotron to generate each kubeconfig with **unique naming based on the cluster name** by default (or as an option). Specifically, I propose the following naming convention:

Expected output (example for `wl1`):

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: xxx
    server: https://k8s-default-xxx-xxx-xxx.elb.ap-northeast-1.amazonaws.com:6443
  name: wl1-k0s
contexts:
- context:
    cluster: wl1-k0s
    user: wl1-admin
  name: wl1-k0s
current-context: wl1-k0s
kind: Config
preferences: {}
users:
- name: wl1-admin
  user:
    client-certificate-data: xxx
    client-key-data: xxx
```

With this format, multiple kubeconfigs can be safely merged using `kubectl config view --flatten`, contributing to improved readability and prevention of misoperations.

### Describe alternatives you've considered

Currently working around this with the following approach, but relying on manual work or scripts leaves room for human error:

```bash
KUBECONFIG=./mgmt-cluster.yaml clusterctl get kubeconfig wl1 > .kube/wl1.yaml
KUBECONFIG=./mgmt-cluster.yaml clusterctl get kubeconfig wl2 > .kube/wl2.yaml

[...]

PR #1165, opened 2025-08-20
feat: add cluster-specific naming to generated kubeconfig

kahirokunn:
## Description

Fixes: #1164

`K0smotronCluster` now generates kubeconfigs with **cluster-scoped context and user names**.

* context: `<cluster-name>-k0s`
* user: `<cluster-name>-admin`

## Impact

* On reconciliation, existing kubeconfigs are updated so that only the **keys (context/user names)** change.
* **No behavioral impact**: credentials, cluster endpoints, and permissions remain the same; no downtime.


jnummelin:
Like mentioned in the issue discussion, we can't rely on https://github.com/k0sproject/k0s/pull/6303 in k0smotron side. That'll create a version dependency between the project which we want to avoid completely. That feature would be available only from certain k0s version onward and thus k0smotron would have to do `if k0sVersion >= x.y.z` type of logic, or to try to detect if that flag is available or not. I think we should avoid that.

PR #1183, opened 2025-08-29
Add support for ignition v3

apedriza:
e2e test using ignition is not wired up at the moment to CI but included a configuration to run it "locally" if desired


internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -230,6 +231,11 @@ func (c *ControlPlaneController) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 
+	format := "cloud-config"
```
makhov: maybe move it to constants?

internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -354,37 +361,48 @@ func (c *ControlPlaneController) generateBootstrapDataForController(ctx context.
...
+		customUserData, err = resolveContentFromFile(ctx, c.Client, scope.Cluster, scope.Config.Spec.CustomUserDataRef)
 		if err != nil {
 			return nil, fmt.Errorf("error extracting the contents of the provided custom controller user data: %w", err)
 		}
-		ci.CustomCloudInit = customCloudInit
 	}
 
-	return ci.AsBytes()
+	var p provisioner.Provisioner = &provisioner.CloudInitProvisioner{}
+	if scope.Config.Spec.K0sConfigSpec != nil && scope.Config.Spec.K0sConfigSpec.Ignition != nil {
```
makhov: just an idea that this could be a func `IsIgnition()` or `GetProvisionerType`.

internal/provisioner/ignition.go
```diff
@@ -0,0 +1,130 @@
...
+
+// IgnitionProvisioner implements the Provisioner interface for Ignition.
+type IgnitionProvisioner struct {
+	Variant          string
+	Version          string
+	AdditionalConfig string
+}
+
+// ToProvisionData converts the input data to Ignition user data.
+func (i *IgnitionProvisioner) ToProvisionData(input *InputProvisionData) ([]byte, error) {
```
makhov, line 56: would nice to have some stupid unit tests for this

internal/controller/util/download.go
```diff
@@ -18,10 +18,56 @@ package util
...
 	"fmt"
+
+	"github.com/k0sproject/k0smotron/internal/provisioner"
 )
 
+// downloadScript is the script used to download and install k0s on a node. It bassically mimics the official get.k0s.sh script
+// but allows specifying the install path and architecture via environment variables.
+// TODO: Use the official get.k0s.sh script when it supports these features.
+// See: https://github.com/k0sproject/get/pull/22
+const downloadScript = `#!/bin/sh
```
makhov: Note, that having it as a file will increase user-data size, so I'd love to keep it on get.k0s.sh. If we are won't agree on merging these changes to the main script, we probable can have one specifically for k0smotron like `get.k0s.sh/for-k0smoton`

apedriza:
@makhov thanks for the feedback! I addressed all the suggestions. Once the script modification is clarified, I will use it instead of rewriting it

internal/controller/bootstrap/worker_bootstrap_controller.go
```diff
@@ -258,20 +261,27 @@ func (r *Controller) generateBootstrapDataForWorker(ctx context.Context, log log
...
-		customCloudInit, err := resolveContentFromFile(ctx, r.Client, scope.Cluster, scope.Config.Spec.CustomUserDataRef)
+		customUserData, err = resolveContentFromFile(ctx, r.Client, scope.Cluster, scope.Config.Spec.CustomUserDataRef)
 		if err != nil {
 			return nil, fmt.Errorf("error extracting the contents of the provided custom worker user data: %w", err)
 		}
-		ci.CustomCloudInit = customCloudInit
 	}
 
-	return ci.AsBytes()
+	var p provisioner.Provisioner = &provisioner.CloudInitProvisioner{}
```
makhov: nitpick: we can probably add a provisioner to scope

internal/controller/infrastructure/ssh_provisioner.go
```diff
@@ -39,7 +39,7 @@ var regex = regexp.MustCompile(`--kubelet-root-dir[ =](/[/a-zA-Z0-9_-]+)+`)
 
 type SSHProvisioner struct {
 	bootstrapData []byte
-	cloudInit     *cloudinit.CloudInit
+	cloudInit     *provisioner.InputProvisionData
```
makhov: I think cloudInit is not a good name for it anymore. maybe `bootstrapData`. And the current bootstrapData field is not used, so can be removed

internal/provisioner/provisioner.go
```diff
@@ -0,0 +1,38 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package provisioner
+
+// InputProvisionData holds the data needed for provisioning a machine.
+type InputProvisionData struct {
+	Files          []File   `yaml:"write_files" json:"files,omitempty"`
+	Commands       []string `yaml:"runcmd" json:"cmds,omitempty"`
```
makhov: Maybe leave some comment about why tags are different

internal/controller/bootstrap/worker_bootstrap_controller.go
apedriza: done, thanks!

internal/provisioner/provisioner.go
apedriza: I reverted it and removed the json tag. I think we dont need it

internal/controller/infrastructure/ssh_provisioner.go
apedriza: I left the field name as `cloudInit` because I didn’t see much value in using the Ignition format here. Ignition is designed to run during a machine’s early boot, while provisioning over SSH happens post-boot, which doesn’t feel like a natural fit to me.

Another option could be to support a translation from Ignition into files+commands, but Ignition allows declaring more specific things (like users, disks etc) that dont map well to the idea of “post-boot provisioning” as in the SSH case. WDYT?

makhov: Yeah, makes sense. Thanks for the explanation!

PR #1188, opened 2025-09-09
refactor `config` structure for more flexibility during installation modes

### LINKED ISSUE #1179: refactor `config` structure for more flexibility in k0smotron installation
### What happened?

Currently, it is not possible to install k0smotron standalone only. Furthermore, the generation of components necessary for each provider in the cluster API integration seems to overlap when generating RBACs, i.e., each `*-provider.yaml` contains references to other components. For a project consisting of different components that can be used independently (capi-bootstrapprovider/capi-controlplaneprovider/capi-infraprovider/standalone), we should have more independence between them when installing it.

On the other hand, it seems that some resources from the project’s initial scaffolding are no longer needed: rbac roles kube-rbac-proxy resources and k0smotron cluster editor/viewer, and prometheus monitoring.

### Involved Areas

Cluster API integration, Standalone

I think it would be a good idea to offer an alternative to installing k0smtoron standalone. I also believe that each k0smtoron capi provider should not have references to other crds other than those related to the provider's api group. 

### Steps to reproduce

NA

### k0smotron version

NA

### k0s version

NA

Regarding the k0smotron installation process: the current documentation does not clearly explain that a full k0smotron installation will not set up the CAPI controller unless CAPI is already installed. As a result, users might accidentally mix installation methods, which can lead to issues such as deploying k0smotron “twice" like https://github.com/k0sproject/k0smotron/issues/1180. I propose:

- Provide a dedicated installation method for k0smotron standalone, including only the resources required for standalone usage.
- Clarify in the documentation that the “full installation” will only deploy the CAPI-related controller if CAPI has been installed beforehand. Ideally, I think offer a single standard installation method for CAPI integration using clusterctl. This aligns with the standard process for installing a CAPI provider, although it would break the current installation approach for some users.
- Delete roles not needed for the provider installed (check `x-provider.yaml` roles for each k0smotron provider)

apedriza:
fix #1179

## Changes
- Three ways to install k0smotron:
  - only standalone mode. The Install manifest only includes resources related to standalone mode.
  - full installation. All in one manifest (as it is now)
  - per CAPI provider. `*-components.yaml` only includes resources related to it. Some RBAC roles have been added for each provider because the "bug" was masked when including all providers' RBACs per `*-components.yaml`

- Now that standalone mode can be installed only using required manifests, if CAPI core or other CAPI providers were installed before installing standalone, the k0smotron manager failed because it detects there are CAPI resources + k0smotron CAPI CRDs were not installed with the new standalone manifest. Added a new flag to enable only the standalone controller to have more control over which controller needs to be set up.


Makefile
```diff
@@ -185,49 +203,43 @@ ifndef ignore-not-found
   ignore-not-found = false
 endif
 
-.PHONY: install
-install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl create -f -
-
-.PHONY: uninstall
-uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
```
apedriza, line 208: I dont think this is really used since we need to tweak the dev environment to not deploy the webhook locally. I will take a general look to improve devx soon.

cmd/main.go
```diff
@@ -297,6 +279,9 @@ func main() {
 				os.Exit(1)
 			}
 		}
+	} else if isControllerEnabled(standaloneController) {
```
apedriza, line 321: new flag value `--enable-controller=standalone` allows us to run only controllers for standalone mode. If not, combining cluster-api core controller + standalone k0smotron is not possible (weird combination but someone might run CAPI without k0smotron provider and k0smotron in standalone mode)

internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -84,7 +84,7 @@ type kmcScope struct {
 // +kubebuilder:rbac:groups=k0smotron.io,resources=clusters/scale,verbs=get;update;patch
 // +kubebuilder:rbac:groups=k0smotron.io,resources=clusters/finalizers,verbs=update
 // +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list
+// +kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch
```
makhov, line 87: is it in a scope of this PR?

config/crd/bootstrap.cluster.x-k8s.io_k0scontrollerconfigs.yaml
makhov, line 1: Couldn't we do it without having another copy of the file, just using kustomize?

internal/controller/bootstrap/worker_bootstrap_controller.go
```diff
@@ -83,6 +83,10 @@ type Scope struct {
 // +kubebuilder:rbac:groups=exp.cluster.x-k8s.io,resources=machinepools;machinepools/status,verbs=get;list;watch
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinepools;machinepools/status,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=secrets;events;configmaps,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=k0smotroncontrolplanes/status,verbs=get;list;watch;create;update;patch;delete
```
makhov, line 89: Why we need additional permissions?

apedriza: Now that roles have been split, the bootstrap provider requires a new rule to retrieve the control plane resource and determine whether it is an HCP (for the new federated HCP feature). Previously, all providers shared the same roles, so the missing rule for the bootstrap provider was hidden by this overlap

config/crd/bootstrap.cluster.x-k8s.io_k0scontrollerconfigs.yaml
apedriza: mm yes we should be able. I will take a look to it

apedriza:
@makhov this is ready for another round 😁

config/manager/manager.yaml
```diff
@@ -0,0 +1,76 @@
...
+    spec:
+      securityContext:
+        runAsNonRoot: true
+      containers:
+      - command:
+        - /manager
+        args:
+        - "--health-probe-bind-address=:8081"
+        - "--metrics-bind-address=127.0.0.1:8080"
+        - "--leader-elect"
```
makhov, line 47: The current state of the patch file is:
```
        - "--health-probe-bind-address=:8081"
        - "--diagnostics-address=127.0.0.1:8080"
        - "--leader-elect"
        - "--enable-controller=bootstrap"
        - --feature-gates=${K0SMOTRON_FEATURE_GATES:=""}
```

And why we switching from patches to full files for controllers?

config/clusterapi/controlplane/manager/manager.yaml
```diff
@@ -0,0 +1,76 @@
...
+    spec:
+      securityContext:
+        runAsNonRoot: true
+      containers:
+      - command:
+        - /manager
+        args:
+        - "--health-probe-bind-address=:8081"
+        - "--metrics-bind-address=127.0.0.1:8080"
+        - "--leader-elect"
```
makhov: this again will start full set of controllers, right?

config/manager/manager.yaml
apedriza: this is the same what we currently use for the make `*-component.yaml` targets where we use the manager declaration (by the way I removed manager config duplication and use the same for all in the last changes) and then when we call kustomize build where we use the patch file in every component with the proper `enable-controller` value

config/clusterapi/controlplane/manager/manager.yaml
apedriza: every component has the proper patch file for the manager setting the expected enable-controller flag

apedriza:
@makhov can you PTAL again?

config/clusterapi/all/manager/manager.yaml
```diff
@@ -0,0 +1,110 @@
...
+      #             operator: In
+      #             values:
+      #               - amd64
+      #               - arm64
+      #               - ppc64le
+      #               - s390x
+      #           - key: kubernetes.io/os
+      #             operator: In
+      #             values:
+      #               - linux
```
jnummelin: I don't think we needs this

config/clusterapi/all/manager/manager.yaml
```diff
@@ -0,0 +1,110 @@
...
+      securityContext:
+        runAsNonRoot: true
+        # TODO(user): For common cases that do not require escalating privileges
+        # it is recommended to ensure that all your Pods/Containers are restrictive.
+        # More info: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
+        # Please uncomment the following code if your project does NOT have to work on old Kubernetes
+        # versions < 1.19 or on vendors versions which do NOT support this field by default (i.e. Openshift < 4.11 ).
+        # seccompProfile:
+        #   type: RuntimeDefault
```
jnummelin: I don't think we needs this

config/clusterapi/all/manager/manager.yaml
apedriza: done!

config/clusterapi/all/manager/manager.yaml
apedriza: done!

PR #1190, opened 2025-09-10
RecreateDeleteFirst update strategy

makhov:


internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -454,7 +458,9 @@ func (c *K0sController) reconcileMachines(ctx context.Context, cluster *clusterv
 		}
 	}
 
-	if infraMachineMissing || (len(machineNamesToDelete)+len(desiredMachines) > int(kcp.Spec.Replicas)) {
+	if infraMachineMissing ||
+		(len(machineNamesToDelete)+len(desiredMachines) > int(kcp.Spec.Replicas)) ||
+		((clusterIsUpdating || configurationHasChanged) && isRecreateDeleteFirstPossible(kcp, machineNamesToDelete, desiredMachines)) {
```
apedriza: this is becoming a bit hard to understand IMO. Also, for me this seems that this condition is for RecreateDeleteFirst strategy only. Maybe something like: 

```go
if machineNeedsToBeDeleted(...) {
      if kcp.Spec.UpdateStrategy == cpv1beta1.UpdateRecreateDeleteFirst && !isRecreateDeleteFirstPossible{
              continue
      }
}
```

Now that this logic has several conditions, maybe wrapping them into functions might improve readability

inttest/capi-docker-machine-template-update-recreate-delete-first/capi_docker_machine_template_update_recreate_delete_test.go
apedriza, line 1: can we test node is deleted first before recreating the new one? Maybe with some watcher over the number of machines (but I dont know if it might create flakiness in the e2e). We have also the option to add some cases  for this in some of the existing integration tests `k0s_controlplane_controller_test.go`. 

Worth to move this e2e to use the e2e framework?

chramb:
on controller + worker nodes k0smotron and capi-operator have race condition where k0smotron removes node from etcd triggering https://github.com/k0sproject/k0s/issues/6526 which triggers https://github.com/k0sproject/k0s/issues/6528 because of which pods are stuck in terminating state and prevent capi-operator from removing the node because drain never finishes.

Will be testing with https://github.com/k0sproject/k0smotron/pull/1248 to see if it fixes the issue

Tizaniy:
> on controller + worker nodes k0smotron and capi-operator have race condition where k0smotron removes node from etcd triggering [k0sproject/k0s#6526](https://github.com/k0sproject/k0s/issues/6526) which triggers [k0sproject/k0s#6528](https://github.com/k0sproject/k0s/issues/6528) because of which pods are stuck in terminating state and prevent capi-operator from removing the node because drain never finishes.
> 
> Will be testing with #1248 to see if it fixes the issue

I face the same issue. I successfully tested the RecreateDeleteFirst update strategy on the BM cluster with 3 control planes but needed some manual intervention to play around a stuck drain.

makhov:
@Tizaniy what is your setup? Is it controller+worker? What LB so you use?

Tizaniy:
@makhov I did it on the simplest virtualized environment with just three control+worker nodes. Keepalived as a load balancer, and custom heathcheck scripts suggested by the team to avoid the issue with routing to unresponsive nodes https://github.com/k0sproject/k0s/issues/6527.

inttest/capi-docker-machine-template-update-recreate-delete-first/capi_docker_machine_template_update_recreate_delete_test.go
makhov: Done!

internal/controller/controlplane/k0s_controlplane_controller.go
makhov: Tried to make it clearer and easier to read

PR #1192, opened 2025-09-12
Add k0s leave scripts only for older k0s versions

makhov:
Adding those scripts increases user-data size and in general it is a workaround. EtcdMember was introduced in v1.30.x and stabilized in v1.30.10 and v1.31.6. 
So since v1.31.6+k0s.0 it should be safe to stop adding those scripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: $(ENVTEST)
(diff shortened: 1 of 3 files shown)


apedriza:
worth encapsulating over a func the logic to build the runcmd given some params like the k0s version?

makhov:
@apedriza done

apedriza:
> @apedriza done

Sorry I miss to mention why encapsulating, for unit test it

makhov:
Good point, I've added some tests

PR #1194, opened 2025-09-12
Set file permissions in job provisioner

makhov:
Currently we don't set permissions, so shell scripts can't be executed


apedriza:
are last fixes related to some bug/regression reported?

makhov:
No, this one I noticed working on another feature.

PR #1201, opened 2025-09-19
Ingress support v2 

makhov:
Another approach for ingress support, based on other k0s modifications


apedriza:
@makhov I'll take a closer look at this. I want to test it properly, so it might take me some time

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -115,6 +115,28 @@ type ClusterSpec struct {
...
+	//+kubebuilder:validation:Optional
+	KonnectivityHost string `json:"konnectivityHost,omitempty"`
+	// ClassName defines the ingress class name to be used by the ingress controller.
+	//+kubebuilder:validation:Optional
+	ClassName string `json:"className,omitempty"`
+	// Annotations defines extra annotations to be added to the ingress controller service.
+	//+kubebuilder:validation:Optional
+	//+kubebuilder:default={haproxy.org/ssl-passthrough: "true"}
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
```
apedriza, line 139: I'm thinking that cases where the user sets fields in the spec but deploy is `false` can be misleading. Additionally, there is the possibility of declaring an external or built-in ingress when `APIHost` or `KonnectivityHost` is empty, or even with a value different from the one provided in the ingress. 

Another approach, more strict in declaring ingress details might be:

```go
type IngressSpec struct {
	// Ref references an existing Ingress resource.
    // If not set, the system will create and manage one automatically.
    Ref *corev1.LocalObjectReference `json:"ref,omitempty"`
    // BuiltIn defines parameters for the built-in ingress, if Ref is not set.
    BuiltIn *BuiltInIngressSpec `json:"managed,omitempty"`
}

type BuiltInIngressSpec struct {
	// Port defines the port used by the ingress controller
	//+kubebuilder:default=443
	Port int64 `json:"port,omitempty"`
	APIHost string `json:"apiHost,omitempty"`
	KonnectivityHost string `json:"konnectivityHost,omitempty"`
	// ClassName defines the ingress class name to be used by the ingress controller.
	ClassName string `json:"className,omitempty"`
	// Annotations defines extra annotations to be added to the ingress controller service.
	//+kubebuilder:default={haproxy.org/ssl-passthrough: "true"}
	Annotations map[string]string `json:"annotations,omitempty"`
}
```

This way it is more explicit that the fields in the managed spec are only used for the built-in ingress, otherwise, get all needed values from the provided ingress. Then, for better UX, we can add some lines in the webhook checking that both (`Ref` and `BuiltIn`) are not set at the same time, or if we want to allow it, throw a warning about using the provided ref as default in this case, or forcing one of them to be set. Also, this way, we drop the `Deploy` field

internal/controller/k0smotron.io/k0smotroncluster_webhook.go
apedriza, line 1: Should we add the same checks for the `K0smotronControlPlane` resource? Otherwise, a user who applies a non-supported k0s version for ingress + CAPI integration will not immediately see that something is wrong on his side

e2e/util/hcp.go
```diff
@@ -90,12 +90,12 @@ func WaitForHCPToBeReady(ctx context.Context, getter capiframework.Getter, cp *c
 
 		desiredReplicas := controlplane.Spec.Replicas
 		statusReplicas := controlplane.Status.Replicas
-		updatedReplicas := controlplane.Status.UpdatedReplicas
+		//updatedReplicas := controlplane.Status.UpdatedReplicas
```
apedriza: Probably not very useful the updatedReplicas field now but in future e2e tests for HCP upgrades, keeping this field can be useful. Can we add it back?

internal/controller/bootstrap/common.go
```diff
@@ -68,32 +68,36 @@ func resolveFiles(ctx context.Context, cli client.Client, cluster *clusterv1.Clu
...
 			}
 		} else {
 			args = append(args, arg)
 		}
 	}
-	if isWorker && !useSystemHostname {
-		args = append(args, kubeletExtraArgs)
+	if len(kubeletExtraArgs) > 0 {
+		args = append(args, fmt.Sprintf(`--kubelet-extra-args="%s"`, strings.Join(kubeletExtraArgs, " ")))
 	}
```
apedriza, line 101: this looks like a fix for cases where `kubelet-extra-args` was more than one and also a fix where we only add them in case `isWorker && !useSystemHostname`, doesn't it? worth to add a unit test for this method?

internal/controller/bootstrap/worker_bootstrap_controller.go
```diff
@@ -323,13 +335,79 @@ func (r *Controller) getK0sToken(ctx context.Context, scope *Scope) (string, err
...
+		if err != nil {
+			return "", fmt.Errorf("failed to create join token: %w", err)
+		}
+	} else {
+		var err error
+		joinToken, err = kutil.CreateK0sJoinToken(ca.KeyPair.Cert, token, fmt.Sprintf("https://%s:%d", scope.Cluster.Spec.ControlPlaneEndpoint.Host, scope.Cluster.Spec.ControlPlaneEndpoint.Port), "kubelet-bootstrap")
+		if err != nil {
+			return "", fmt.Errorf("failed to create join token: %w", err)
+		}
 	}
```
apedriza, line 358: wdyt?

```go
joinURL := fmt.Sprintf("https://%s:%d", scope.Cluster.Spec.ControlPlaneEndpoint.Host, scope.Cluster.Spec.ControlPlaneEndpoint.Port)
if scope.ingressSpec != nil {
    joinURL = fmt.Sprintf("https://%s:%d", scope.ingressSpec.APIHost, scope.ingressSpec.Port)
}

joinToken, err := kutil.CreateK0sJoinToken(ca.KeyPair.Cert, token, joinURL, "kubelet-bootstrap")
if err != nil {
	return "", fmt.Errorf("failed to create join token: %w", err)
}
```

internal/controller/k0smotron.io/k0smotroncluster_certs.go
```diff
@@ -116,3 +116,83 @@ func (scope *kmcScope) ensureEtcdCertificates(ctx context.Context, kmc *km.Clust
 
 	return etcdCerts.SaveGenerated(ctx, scope.client, util.ObjectKey(kmc), *metav1.NewControllerRef(kmc, km.GroupVersion.WithKind("Cluster")))
 }
+
+func (scope *kmcScope) ensureIngressCerts(ctx context.Context, kmc *km.Cluster) error {
```
apedriza:
```suggestion
func (scope *kmcScope) ensureHAProxyCerts(ctx context.Context, kmc *km.Cluster) error {
```

internal/controller/k0smotron.io/k0smotroncluster_webhook.go
```diff
@@ -3,19 +3,83 @@ package k0smotronio
...
+// ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type Cluster.
+func (c ClusterValidator) ValidateDelete(_ context.Context, _ runtime.Object) (warnings admission.Warnings, err error) {
+	return nil, nil
+}
+
+var ingressCompatibleVersions = []*version.Version{
+	version.MustParse("v1.34.1+k0s.0"),
+	version.MustParse("v1.33.5+k0s.0"),
+	version.MustParse("v1.32.9+k0s.0"),
+	version.MustParse("v1.31.13+k0s.0"),
```
apedriza:
```go
version.MustParse("v1.33.5+k0s.0"),
version.MustParse("v1.32.9+k0s.0"),
version.MustParse("v1.31.13+k0s.0"),
```

I couldn't find none of the above versions

internal/controller/k0smotron.io/k0smotroncluster_ingress.go
```diff
@@ -0,0 +1,245 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        kmc.GetIngressManifestsConfigMapName(),
+			Namespace:   kmc.Namespace,
+			Annotations: kcontrollerutil.AnnotationsForK0smotronCluster(kmc),
+		},
+		Data: map[string]string{
+			// Due to k0s behavior, we need to create a dummy Endpoints object to create a worker profile
+			// Once a local haproxy is up, it will update the Endpoints object to point with the actual proxy's IP
+			"0_temp-kubernetes-ep.yaml": `apiVersion: v1
+kind: Endpoints
```
apedriza, line 76: don't know how it might affect but seems `Endpoints` is deprecated. Maybe worth to use `EndpointSlices`. https://kubernetes.io/docs/concepts/services-networking/service/#endpoints

internal/controller/k0smotron.io/k0smotroncluster_kubeconfig.go
```diff
@@ -110,12 +110,15 @@ func rewriteKubeconfigNames(kubeconfigYAML string, clusterName string) (string,
...
+		srcCluster.Server = fmt.Sprintf("https://%s:%d", kmc.Spec.Ingress.APIHost, kmc.Spec.Ingress.Port)
+	}
 	if len(srcUser.ClientCertificateData) == 0 || len(srcUser.ClientKeyData) == 0 {
 		return "", fmt.Errorf("client certificate/key data not found in kubeconfig")
 	}
 
-	clusterKey := fmt.Sprintf("%s-k0s", clusterName)
-	userKey := fmt.Sprintf("%s-admin", clusterName)
+	clusterKey := fmt.Sprintf("%s-k0s", kmc.Name)
+	userKey := fmt.Sprintf("%s-admin", kmc.Name)
```
apedriza, line 121: why this change? I think is better to just use the cluster name string as param instead the whole struct which not needed

internal/controller/k0smotron.io/k0smotroncluster_ingress.go
```diff
@@ -0,0 +1,245 @@
...
+func (scope *kmcScope) reconcileIngress(ctx context.Context, kmc *km.Cluster) error {
+	if kmc.Spec.Ingress == nil {
+		return nil
+	}
+
+	configMap, err := scope.generateIngressManifestsConfigMap(kmc)
+	if err != nil {
+		return fmt.Errorf("failed to generate ingress manifests configmap: %w", err)
+	}
+	_ = ctrl.SetControllerReference(kmc, &configMap, scope.client.Scheme())
```
apedriza: All this references should be done using the ownerference wrap that takes into account if the hcp runs on the management or not for the proper background deletion

internal/controller/bootstrap/worker_bootstrap_controller.go
makhov: thanks, done!

internal/controller/k0smotron.io/k0smotroncluster_ingress.go
makhov: done, thanks!

internal/controller/k0smotron.io/k0smotroncluster_certs.go
makhov: "Ingress" sounds more like a purpose and "HAProxy" is more like implementation, but I don't have a strong opinion here, so changed it, thanks!

internal/controller/bootstrap/common.go
makhov: Added some test cases, thanks!

e2e/util/hcp.go
makhov: Yeah, sorry, removed it for some debugging and forgot to revert. Fixed!

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
makhov: `Ref` won't help us, since we can't get API and konnectivity addresses from there. It's just too many possible different objects, we can't parse them, so it doesn't make sense for us to have it in the IngressSpec type.

We do need `APIAddress` and `KonnectivityAddress` (I renamed fields and got rid of Port) and for better UX we can deploy an Ingress object, I think it will cover 99.99% use-cases, since the Gateway API is in the Dante's Limbo. 

Maybe we can switch it to something like this, but I'm not sure:
```
ingress:
  apiAddress:my.address.com
  konnectivityAddress: konnectivity.my.address.com
  deploy: # or something, idk
    enabled: true
    className: nginx
    annotations: 
    - haproxy.org/ssl-passthrough: "true"
```

makhov: Haha, getting rid of the Port field was not a good idea, actually. Brought it back

internal/controller/k0smotron.io/k0smotroncluster_webhook.go
makhov: Done

internal/controller/k0smotron.io/k0smotroncluster_kubeconfig.go
makhov: Few lines above we use `kmc` in 
```
	if kmc.Spec.Ingress != nil {
		srcCluster.Server = fmt.Sprintf("https://%s:%d", kmc.Spec.Ingress.APIHost, kmc.Spec.Ingress.Port)
	}
```

internal/controller/k0smotron.io/k0smotroncluster_webhook.go
makhov: I've selected the next k0s releases that was actually never released :)
Removed it

internal/controller/k0smotron.io/k0smotroncluster_ingress.go
makhov: k0s uses endpoints, so we'll switch to slices later

makhov:
@apedriza @jnummelin I would really appreciate if you can take a look on this again, I've been working on it since spring and winter is coming :)

apedriza:
> @apedriza @jnummelin I would really appreciate if you can take a look on this again, I've been working on it since spring and winter is coming :)

I will prioritize the review

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -115,6 +116,50 @@ type ClusterSpec struct {
...
+	//+kubebuilder:validation:Optional
+	APIHost string `json:"apiHost,omitempty"`
+	//+kubebuilder:validation:Optional
+	KonnectivityHost string `json:"konnectivityHost,omitempty"`
+	// ClassName defines the ingress class name to be used by the ingress controller.
+	//+kubebuilder:validation:Optional
+	ClassName string `json:"className,omitempty"`
+	// Annotations defines extra annotations to be added to the ingress controller service.
+	//+kubebuilder:validation:Optional
+	//+kubebuilder:default={haproxy.org/ssl-passthrough: "true"}
```
apedriza: if no annotation is set and ingress controller used is not haproxy we are still adding`{haproxy.org/ssl-passthrough: "true"}` annotation in the ingress

e2e/ingress_test.go
```diff
@@ -0,0 +1,197 @@
...
+	require.NoError(t, bootstrapClusterProxy.GetClient().List(ctx, machineList, client.InNamespace(workloadClusterNamespace), client.MatchingLabels{
+		clusterv1.ClusterNameLabel: workloadClusterName,
+	}), "Should list machines")
+
+	for _, m := range machineList.Items {
+		var (
+			stdout bytes.Buffer
+			stderr bytes.Buffer
+		)
+		cmd := exec.Command("docker", "exec", m.Name, "curl", "https://10.128.0.1/healthz", "--cacert", "/etc/haproxy/certs/ca.crt")
```
apedriza, line 160: Wonder if we might check connectivity using another infra agnostic method. Now that aws is added in CI for our e2e I think worth to add TestIngress also to the nightly e2e suite. It can be a follow up PR

internal/controller/bootstrap/worker_bootstrap_controller.go
```diff
@@ -334,13 +346,75 @@ func (r *Controller) getK0sToken(ctx context.Context, scope *Scope) (string, err
...
+				Path: "/etc/haproxy/certs/server.key",
+			},
+			ContentFrom: &bootstrapv1.ContentSource{
+				SecretRef: &bootstrapv1.ContentSourceRef{
+					Name: secret.Name(scope.Cluster.Name, "ingress-haproxy"),
+					Key:  "tls.key",
+				},
+			},
+		},
+	})
```
apedriza, line 409: probably we need to take into account recent changes using the getter that encapsulates the workinigDir for files

internal/controller/controlplane/k0smotron_controlplane_webhook.go
```diff
@@ -56,7 +57,14 @@ func (v *K0smotronControlPlaneValidator) ValidateCreate(_ context.Context, obj r
...
 
-	warnings := v.validateVersionSuffix(kcp.Spec.Version)
+	warnings, err := v.validate(kcp)
+	if err != nil {
+		return warnings, err
+	}
+
+	vsWarnings := v.validateVersionSuffix(kcp.Spec.Version)
+	warnings = append(warnings, vsWarnings...)
+
```
apedriza: looks like `validate` should contain `validateVersionSuffix` mehtod

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
apedriza: in which cases we should not deploy the ingress? Im thinking about whether remove `Deploy` field is a good option for simplicity

internal/controller/k0smotron.io/k0smotroncluster_ingress.go
```diff
@@ -0,0 +1,244 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+
+	if kmc.Spec.Ingress.ClassName != "" {
+		ingress.Spec.IngressClassName = &kmc.Spec.Ingress.ClassName
+	}
```
apedriza:
```suggestion
	ingress.Spec.IngressClassName = &kmc.Spec.Ingress.ClassName
```
I guess whatever is set in the kmc spec need to be forwarded to the ingress

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
apedriza: also, should be APihost and konnectivityAddress required instead optional? We are using those fields as set in mostly every place

apedriza (commented):
looking good, some comments here

internal/controller/controlplane/k0smotron_controlplane_webhook.go
makhov: Done

internal/controller/bootstrap/worker_bootstrap_controller.go
makhov: This is for HCP, so the path doesn't matter here

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
makhov: Fixed

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
makhov: > in which cases we should not deploy the ingress? Im thinking about whether remove Deploy field is a good option for simplicity

We deploy Ingress object by default as most of the people use one of the popular ingress controllers and we know how to properly configured. But some users may use Gateway API instead or may want to configure Ingress in a weird way, which we can't cover, so there is an option to deploy your own.

> also, should be APihost and konnectivityAddress required instead optional? We are using those fields as set in mostly every place

Done

e2e/ingress_test.go
makhov: I'll try to find a way, but I don't want to change the test AND the code at the same time, so I think the best time to do it is after the PR is merged

internal/controller/k0smotron.io/k0smotroncluster_ingress.go
makhov: Done

makhov:
Yeah, I'll think how to add this info in a separate PR just to avoid new merge conficts

Issue #1197, opened 2025-09-15
Allow More Flexible Templating for UserData


aglarendil:
### Is your feature request related to a problem? Please describe.

Hiya folks, the problem is that seemingly as we are expanding usage of k0smotron for various field scenarios, we need to be able to do execute more complex pre and post steps before or after k0s is triggered to actually configure the node
these things would be:
1. predownload some software including k0s in airgapped env
2. precreate users with predictable ids and create some dirs, for example, to preconfigure audit logging
3. from discussion between Jussi and Serhii

```
Jussi Nummelin
  just to wrap my head better around this, what are the exact things you want to run pre&post?
Serhii Ivanov
  in pre I have some obscure hacks to make network working and in post I need to do some k8s hotfixes\
  but my example is not representing a problem good enough, I can do it in other ways also
  assume something like k0rdent with a lot of abstraction layers and user-data is exposed directly to consumer,  like single CM
  vs
  having multiple places and fields directly in cluster definition
Jussi Nummelin
  but k0smotron is also doing other things in the CI than commands, we need multiple placeholders for things.   kinda feels we'd need some real templating...
```

### Describe the solution you would like

I suggest that all the steps that are generated during cloud init stage become a set of template macros, that can be injected into ANY template for user-data, so a user can create cm with template that will have a placeholder for this k0smotron macro like
MYSTUFF BEFORE
"${{ download_k0s}}"
"${{ run_k0s }}
MYSTUFF LATER

### Describe alternatives you've considered

supposedly, we have a partial solution here that merges userData, but unfortunately, it sometimes allows only for example, for pre-files or post-files be added, but not both at the same time
https://github.com/k0sproject/k0smotron/issues/976

otherwise I would have to wrap my hacks into pre and post commands as base64 encoded bash scripts which is kind of not very neat and might miss some of the corner cases

aglarendil:
related to https://github.com/k0sproject/k0s/issues/6377
because we need some dances around and sometimes post to make audit logs actually work

makhov:
Looks like you don't need neither custom nor merging user-data. k0smotron has `spec.preStartCommands` and `spec.postStartCommands`. They do exactly what you need. Basically, the order of things here:
- preStartCommands
- all k0smotron commands like downloading and installing k0s
- postStartCommands

The name of the fields is a bit misleading tho

aglarendil:
[...]

PR #1202, opened 2025-09-20
expose k0smotron commands as cloud-init vars

makhov:
Putting k0s commands to vars when feature flag is enabled. This allows full control over used-data using customCloudInit. Requires setting `--feature-gates=CloudInitVars=true`.

**Warning:** no user-data will be generated by k0smotron, user is fully responsible for the vars and files usage.

The following example shows the full set of vars and how to use them:
```
runcmd:
 - echo -n "custom" > /tmp/custom
 - {{ k0smotron_k0sDownloadCommands }}
 - {{ k0smotron_k0sInstallCommand }}
 - {{ k0smotron_k0sStartCommand }}

write_files:
{% for f in k0smotron_files %}
  - path: {{ f.path }}
    content: |
      {{ f.content | indent(6) }}
    permissions: "{{ f.permissions }}"
{% endfor %}
  - path: /tmp/my-extra-file
    content: test
    permissions: "0600"
```

**UPD**: I swithched double quotes to single ones, it should fix the templating issue. 
Also I've added `K0SMOTRON_FEATURE_GATES` var, so you can do `K0SMOTRON_FEATURE_GATES=CloudInitVars=true,AnotherGate=…`

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -118,7 +118,10 @@ func main() {
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
 
-	_ = featuregate.Configure(featureGates, os.Getenv(featuregate.EnvVarName))
+	err := featuregate.Configure(featureGates, os.Getenv(featuregate.EnvVarName))
+	if err != nil {
+		setupLog.Error(err, "unable to configure feature gates, the provided flags will be ignored")
+	}
 
 	if enabledController != "" && enabledController != allControllers {
 		enabledControllers = map[string]bool{
@@ -127,14 +130,6 @@ func main() {
 	}
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
-
-	// NOTE: support both the deprecated and capi metrics flags
-	// TODO: remove the next block in favor of:
-	// _, metricsOpts, err := flags.GetManagerOptions(managerOptions)
-	// if err != nil {
-	// 	setupLog.Error(err, "unable to start manager: invalid flags")
-	// 	os.Exit(1)
-	// }
 
 	var metricsOpts metricsserver.Options
 	{


s3rj1k:
Hey @makhov, thanks for working on this, I was able to test this with barebone k0smotron and CAPO infrastructure.

Currently it does not work (cloud-init thing), I am getting this error from cloud-init

```
2025-10-10 13:11:45,624 - jinja_template.py[WARNING]: Ignoring jinja template for part-001. Failed to render template. Unable to parse Jinja template due to syntax error: expected token 'end of statement block', got '=' on line 4: {% set k0smotron_k0sInstallCommand = "/usr/local/bin/k0s install controller --force --enable-dynamic-config --env AUTOPILOT_HOSTNAME=s3rj1k-capo-control-plane-xqm4j --labels=k0smotron.io/machine-name=s3rj1k-capo-control-plane-xqm4j --enable-worker --disable-components=konnectivity-server --no-taints --config /etc/k0s.yaml --kubelet-extra-args="--hostname-override=s3rj1k-capo-control-plane-xqm4j"" %}
```

and user-data file that gets created looks like this (`...` to reduce bloat, all data inside was correctly rendered)

```
# cat user_data
## template: jinja
{% set k0smotron_postStartCommands = "" %}
{% set k0smotron_k0sDownloadCommands = "curl -sSfL --retry 5 https://get.k0s.sh/ | K0S_INSTALL_PATH=/usr/local/bin K0S_VERSION=v1.33.4+k0s.0 sh" %}
{% set k0smotron_k0sInstallCommand = "/usr/local/bin/k0s install controller --force --enable-dynamic-config --env AUTOPILOT_HOSTNAME=s3rj1k-capo-control-plane-xqm4j --labels=k0smotron.io/machine-name=s3rj1k-capo-control-plane-xqm4j --enable-worker --disable-components=konnectivity-server --no-taints --config /etc/k0s.yaml --kubelet-extra-args="--hostname-override=s3rj1k-capo-control-plane-xqm4j"" %}
{% set k0smotron_k0sStartCommand = "/usr/local/bin/k0s start" %}
{% set k0smotron_preStartCommands = "passwd -u root && echo "root:root" | chpasswd" %}
{% set k0smotron_files = [
  {
    "path": "/etc/k0s.yaml",
    "content": "...",
    "permissions": "0644"
  },
  {
    "path": "/var/lib/k0s/pki/ca.crt",
    "content": "...",
    "permissions": "0644"
  },
  {
    "path": "/var/lib/k0s/pki/ca.key",
    "content": "...",
    "permissions": "0644"
  },
  {
    "path": "/var/lib/k0s/pki/etcd/ca.crt",
    "content": "...",
    "permissions": "0644"
  },
  {
    "path": "/var/lib/k0s/pki/etcd/ca.key",
    "content": "...",
    "permissions": "0644"
  },
  {
    "path": "/var/lib/k0s/pki/front-proxy-ca.crt",
    "content": "...",
    "permissions": "0644"
  },
  {
    "path": "/var/lib/k0s/pki/front-proxy-ca.key",
    "content": "...",
    "permissions": "0644"
  },
  {
    "path": "/var/lib/k0s/pki/sa.pub",
    "content": "...",
    "permissions": "0644"
  },
  {
    "path": "/var/lib/k0s/pki/sa.key",
    "content": "...",
    "permissions": "0644"
  }
] %}

#cloud-config
# https://cloudinit.readthedocs.io/en/latest/reference/examples.html

swap:
  size: 0

package_update: true
package_upgrade: true

runcmd:
  - echo -n "custom" > /root/custom
  - "{{ k0smotron_preStartCommands }}"
  - "{{ k0smotron_k0sDownloadCommands }}"
  - "{{ k0smotron_k0sInstallCommand }}"
  - "{{ k0smotron_k0sStartCommand }}"
  - "{{ k0smotron_postStartCommands }}"

write_files:
{% for f in k0smotron_files %}
  - path: {{ f.path }}
    content: |
      {{ f.content | indent(6) }}
    permissions: "{{ f.permissions }}"
{% endfor %}
  - path: /root/my-extra-file
    content: test
    permissions: "0600"
```

Also I would like to ask in addition to extend ability to enable feature-flags to environs also, so it would be possible to export environ `FEATURE_GATE_CLOUD_INIT_VARS` and set it's value to `true` instead of using CLI Args (not replace but in addition to CLI args one can use environs.)

This should make Helm template writers life a bit easier and also setting ENVs in Deployments are safer vs doing something like this

```
kubectl -n kube-system patch deployment k0smotron-controller-manager-bootstrap \
  --type='json' \
  -p='[{"op": "add", "path": "/spec/template/spec/containers/0/args/-", "value": "--feature-gates=CloudInitVars=true"}]'
```

s3rj1k:
I've managed to verify this change to be working with below config

```
  user-data: |2
    swap:
      size: 0

    packages:
      - bash
      - ca-certificates
      - curl
      - dash

    # package_reboot_if_required: true
    package_update: true
    package_upgrade: true

    runcmd:
      - /root/cloud-init.sh && /bin/rm -f /root/cloud-init.sh

    write_files:
      - path: /root/cloud-init.sh
        content: |
          #!/usr/bin/bash
          set -euo pipefail

          export PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"

          {{ k0smotron_k0sDownloadCommands }}
          {{ k0smotron_k0sInstallCommand }}
          {{ k0smotron_k0sStartCommand }}
        permissions: "0755"
    {% for f in k0smotron_files %}
      - path: {{ f.path }}
        content: |
          {{ f.content | indent(6) }}
        permissions: "{{ f.permissions }}"
    {% endfor %}
```

Note that commands are inside a shell script and this is intentional, it seems that putting them directly in `runcmd` steps does not work all the time, at least for me it did not work on Debian 13 cloud-image for some strange reason.

cmd/main.go
```diff
@@ -115,6 +118,8 @@ func main() {
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
 
+	_ = featuregate.Configure(featureGates, os.Getenv(featuregate.EnvVarName))
```
apedriza: throw a warning maybe? can be useful for the user if some funcionality will not be available

aglarendil:
this actually closes https://github.com/k0sproject/k0smotron/issues/1197

internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -367,26 +367,31 @@ func (c *ControlPlaneController) generateBootstrapDataForController(ctx context.
 		files = append(files, genShutdownServiceFiles(scope.Config.Spec.K0sInstallDir)...)
 	}
 
-	commands, err := c.genK0sCommands(scope, installCmd)
+	commands, commandsMap, err := c.genK0sCommands(scope, installCmd)
```
apedriza, line 370: nit: seems like we would be fordwarding data that will not be used to build the provision data, either commands or commandsMap, depending on the flag value. What do you think about outputting a `inputProvisionDataCommands` where we store in a struct all the downlaod, start, install, post.. etc commands and then in the cloud init provisioner decides based on the feature flag how to build the user data? that way we dont fordward unneeded data. So `genK0sCommands` (maybe rename it?) would be generating that `inputProvisionDataCommands` with every field on it depending of the type of commands we have.

makhov: It's a good idea, but requires a deep refactoring, so I'll create a separate PR based on this branch to see how it looks like. So we won't block this one, but can merge and release both if we decide to

cmd/main.go
makhov: Done, thanks!

apedriza:
how about creating a section on feature gates, default values, etc.? even if it's just one

apedriza:
maybe worth to add some functional example (e2e or full k0scontrolplane in docs)

makhov:
As I wrote in docs, I truly believe that no one should use this feature, so I don't really want to provide a lot of docs on it

s3rj1k:
Yay!

PR #1208, opened 2025-09-25
Add pull k0s bin from OCI registry support

apedriza:


api/bootstrap/v1beta1/k0s_types.go
```diff
@@ -289,6 +296,13 @@ type K0sConfigSpec struct {
 	// See: https://cloudinit.readthedocs.io/en/latest/reference/merging.html
 	// +kubebuilder:validation:Optional
 	CustomUserDataRef *ContentSource `json:"customUserDataRef,omitempty"`
+
+	// ArtifactRef is a reference to the artifact to be fetched from an OCI registry.
+	// The format of the reference is: [registry]/[repository]@[digest].
+	// Prerequisite: Oras CLI must be installed on the target machine.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:pattern=`^(([a-zA-Z0-9.-]+(:[0-9]+)?/)?[a-zA-Z0-9._-]+(/[a-zA-Z0-9._-]+)*)@sha256:[a-fA-F0-9]{64}$`
+	ArtifactRef string `json:"artifactRef,omitempty"`
```
makhov: Could we just use the current field and detect the way how we should download it by schema? Like `https://` vs `oci://`?

apedriza: yes good point, that way we don't need to extend the API

internal/controller/util/download.go
```diff
@@ -39,8 +68,8 @@ func DownloadCommands(preInstalledK0s bool, url string, version string, k0sBinPa
 	if version != "" {
 		scriptVars = append(scriptVars, fmt.Sprintf("K0S_VERSION=%s", version))
 	}
-	if k0sBinPath != "" {
-		scriptVars = append(scriptVars, fmt.Sprintf("K0S_INSTALL_PATH=%s", k0sBinPath))
+	if k0sInstallPath != "" {
```
makhov: It's always true

apedriza:
> LGTM in general, but we also need a comprehensive tutorial describing how to use it, eg using files for auth and preStartCommands for installing oras

what do you think is the best option for the guide? a new entry in docs? I don't know where it might fit on it. A new post..?

makhov:
I think somewhere in the Examples section will be a good place for it
https://docs.k0smotron.io/stable/capi-examples/

internal/controller/util/download.go
apedriza: done, thanks

apedriza:
> I think somewhere in the Examples section will be a good place for it https://docs.k0smotron.io/stable/capi-examples/

Added a new entry with that example, thanks for the suggestion!

docs/capi-use-oci-registry.md
```diff
@@ -0,0 +1,182 @@
...
+Configuring the `K0sControlPlane` to pull k0s from an OCI registry is straightforward. **The only requirement is that the machine being bootstrapped needs Oras CLI installed**. You can achieve this in two ways:
+
+- By using `.preStartCommands` to install the Oras CLI on the machine before pulling the binary.
+- By using a machine image with the Oras CLI pre-installed.
+
+```yaml
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+    name: aws-test
```
makhov: I don't really have a personal preferences, but most of examples that I've seen (including all in our docs) uses 2 spaces in yaml, not 4. Can we keep it consistent?

makhov (approved):
Just a nitpick about the yaml formatting, otherwise LGTM

docs/capi-use-oci-registry.md
apedriza: will revert it 👍

PR #1213, opened 2025-10-06
CAPI manager flags

makhov:
To support cluster-api-operator changes


apedriza:
I think we can close this in favor of https://github.com/k0sproject/k0smotron/pull/1215

PR #1215, opened 2025-10-06
fix: support CAPI manager flags

### LINKED ISSUE #1214: Support CAPI flags
## Problem

At the moment, `k0smotron` does not support the [long-deprecated](https://github.com/kubernetes-sigs/cluster-api/pull/9264) and [already removed](https://github.com/kubernetes-sigs/cluster-api/pull/11140) CAPI flag.

However, it supports [a very similar flag](https://github.com/k0sproject/k0smotron/blob/9ba014a9c33574c925c361a7ec56e4f74a7042de/cmd/main.go#L95-L96) that has no relation to the CAPI ecosystem at the moment (it can be modified via neither `capb`/`cacpp`/`caip` objects).

Besides, starting with `v0.24.0` of the CAPI operator, it now [unconditionally sets](https://github.com/fariaass/cluster-api-operator/blob/c9d5d7234f9216bfaf9dee3d741188777846d66c/internal/controller/component_customizer.go#L245) the `insecure-diagnostics` flag, which has no support in `k0smotron`.  

Such behavior causes `k0smotron` `deployments` objects to be configured with that flag, which in turn leads to `CrashLoopBackoff` error of the corresponding `pods` objects. It effectively means that `k0smotron` can't be used with the CAPI operator version `v0.24.0` or higher.

## Workaround

No permanent workaround exists. Patching `deployments` manually removing the flag only temporarily remediates the issue.

## Suggested solution

1. Add `cluster-api` flags to the existing flag set.
2. Deprecate the aforementioned flags along with `secure-metrics` and `enable-http2` for backward compatibility
3. Support both flags (deprecated ones and from the `cluster-api`) for backward compatibility.

**Note**: if deprecation of the flags is not expected, the code could be simplified.

zerospiel:
Adds support of the modern CAPI manager flags to align some of them with the new CAPI operator version (v0.24.0).

Deprecates the following flags:
- metrics-bind-address: interchanged with --diagnostics-address
- metrics-secure: interchanged with --insecure-diagnostics
- enable-http2: not needed anymore

Fixes #1214

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -25,6 +25,7 @@ import (
 
 	"k8s.io/client-go/discovery"
 
+	"github.com/spf13/pflag"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


zerospiel:
Sorry, had to sign off the commit and add the issue to the commit message

makhov (approved):

PR #1220, opened 2025-10-07
Drain contoller+worker nodes

makhov:
Drain node as part of machine removal


makhov:
Yeah, it's probably the third time I'm trying to add it to the k0smotron already :) 
Each time after some other provider misbehaving and preventing CAPI from draining the node properly

PR #1233, opened 2025-10-08
Refactor: make `K0sControllerConfig` immutable

### LINKED ISSUE #1143: move controller config defaults to mutating webhook

apedriza:
fix #1143

### Changes

- Make `K0sControllerConfig` immutable, which means that values related to that resource will not change ever once it is created by the controlplane controller. This way, we dont need to know "external" modifications made by other controllers (bootstrap controller) when comparing if the control plane node needs a rollout because the k0sfig has changed, which improves maintainability.
- Introduce a new `k0s.controlplane.cluster.x-k8s.io/k0s-config` annotation to store k0s config to boot the machine
- Use cmp.Diff for comparisons between kcp k0s config and machine k0s config
- Keep backward compatibility: Deprecate the old comparison method that takes into account changes made outside the controlplane controller

### DIFF
--- a/internal/controller/controlplane/k0s_controlplane_controller.go
+++ b/internal/controller/controlplane/k0s_controlplane_controller.go
@@ -27,6 +27,7 @@ import (
 	"github.com/google/uuid"
 	autopilot "github.com/k0sproject/k0s/pkg/apis/autopilot/v1beta2"
 	"github.com/k0sproject/k0smotron/internal/controller/util"
+	"github.com/k0sproject/version"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -71,6 +72,7 @@ var (
 	FRPConfigMapNameTemplate  = "%s-frps-config"
 	FRPDeploymentNameTemplate = "%s-frps"
 	FRPServiceNameTemplate    = "%s-frps"
+	minVersionForETCDName     = version.MustParse("v1.31.1")
 )
 
 type K0sController struct {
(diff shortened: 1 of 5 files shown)


makhov:
How the annotation is better than the object itself?

apedriza:
> How the annotation is better than the object itself?

It would mainly be used to determine the method for checking configuration changes: if the annotation is found, we use it, knowing that the config controller is immutable, which not requires more than unmarshall and compare annotation vs kcp config spec. Otherwise, we would fall back and use the logic that takes into account the changes made by the other controller. 

As such, the same comparison could be achieved using the object, but it is more complicated and cumbersome when taking into account configurations that were created with the method prior to these changes. In the future, we could stop supporting the fallback block, and I believe it would not be necessary to take into account the config object once created so we would simply check the annotation without retrieving the config object.

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -587,10 +594,7 @@ func (c *K0sController) deleteK0sNodeResources(ctx context.Context, cluster *clu
 	return nil
 }
 
-func (c *K0sController) createBootstrapConfig(ctx context.Context, name string, _ *clusterv1.Cluster, kcp *cpv1beta1.K0sControlPlane, machine *clusterv1.Machine, clusterName string) error {
-
-	k0sConfigSpec := kcp.Spec.K0sConfigSpec.DeepCopy()
-	k0sConfigSpec.Args = uniqueArgs(k0sConfigSpec.Args)
+func (c *K0sController) createBootstrapConfig(ctx context.Context, name string, k0sConfigSpec *bootstrapv1.K0sConfigSpec, kcp *cpv1beta1.K0sControlPlane, machine *clusterv1.Machine, clusterName string) error {
```
makhov, line 597: Sorry, I'm not fully following. We update the annotation on each change and then use the value from annotation to update the bootstrap config, right?

apedriza: annotation is never updated, it is only used to check if we need a rollout, excluding the `k0s` unstructured field which might contain changes that will be reconciled using dynamic config. So bootstrap config will never be modified. <- Only if annotation is present, if not present fallback to the current calculation

PR #1235, opened 2025-10-09
Add 1.7 and 1.8 to e2e

apedriza:


e2e/k0smotron_upgrade_test.go
```diff
@@ -49,7 +49,7 @@ func TestK0smotronUpgrade(t *testing.T) {
 // Important: This version MUST match the ones set in the e2e config file because that guarantees that the clusterctl
 // local repository contains the required versions of the providers when running the upgrade tests. We should test
 // development version against the latests stable version of k0smotron.
-var k0smotronMinorVersionsToCheckUpgrades = []string{"1.5", "1.6"}
+var k0smotronMinorVersionsToCheckUpgrades = []string{"1.5", "1.6", "1.7", "1.8"}
```
apedriza, line 52: At some point, we will have to update this test to the same range that we guarantee backward compatibility.

PR #1237, opened 2025-10-10
Health Checks docs

makhov:


docs/capi-machine-health-checks.md
```diff
@@ -0,0 +1,63 @@
...
+
+k0smotron's control plane controller automatically handles the remediation process when machines are marked as unhealthy by MHC.
+
+Read more about MachineHealthChecks in the [Cluster API documentation](https://cluster-api.sigs.k8s.io/tasks/automated-machine-management/healthchecking.html#configure-a-machinehealthcheckbb).
+
+## How it works
+
+When a MachineHealthCheck detects an unhealthy control plane machine:
+
+1. The MHC controller marks the machine with the `MachineOwnerRemediated` condition
```
apedriza:
```suggestion
1. The MHC controller marks the machine with the `OwnerRemediated ` condition
```

this is the string of the condition. I also think MHC controller needs to add `HealthCheckSucceeded` with `false` value

makhov: thanks, fixed!

apedriza: We still miss the `HealthCheckSucceeded` to `false` I think. Other way is to not mention which conditions are set by the MHC controller a just mention that the machine is marked by it.

makhov: Yeah, sorry, missed it. I removed it and just say that the machine is marked as unhealthy

PR #1245, opened 2025-10-14
fix node recreation when k0s config args changes

apedriza:
args changes in the controlplane do not trigger a recreation of the node. This changes add a fix to detect those changes. Also added a check in the inttest to catch these cases


inttest/capi-controlplane-docker-downscaling/capi_controlplane_docker_downscaling_test.go
```diff
@@ -160,8 +161,21 @@ func (s *CAPIControlPlaneDockerDownScalingSuite) TestCAPIControlPlaneDockerDownS
 	s.T().Log("scaling down control plane")
 	s.updateClusterObjects()
 
-	scaleDownCtx, cancel := context.WithTimeout(s.ctx, 10*time.Minute)
+	scaleDownCtx, cancel := context.WithTimeout(s.ctx, 20*time.Minute)
```
makhov, line 164: Why can't we just use the parent context here?

makhov (commented):
Hm, I absolutely sure that I've tested it and it worked. What is the issue now with it?

apedriza:
> Hm, I absolutely sure that I've tested it and it worked. What is the issue now with it?

If you deploy a cluster without `--enable-worker` and then update it adding that label, the cluster will not rollout its controlplanes. Basically we are copying args from kcp to k0scontrollerconfig before the comparission so that kind of changes will never get caught.

inttest/capi-controlplane-docker-downscaling/capi_controlplane_docker_downscaling_test.go
apedriza: in this test we scale down and also we add a new arg `--enable-worker` so we need to test that the controlplanes scale down properly and that the new controlplane is a worker node too. Before this change (and due to the issue not rolloing out the node) we were checking only that the controlplane scales to 1. So now that I also added a check for waiting for the new node to come up maybe it could take a bit more time

PR #1246, opened 2025-10-14
Enable providerid controller for infra provider

makhov:
Currently, if RemoteMachine provider is used for example with kubeadm, no one sets providerID to the Node object and the process hangs

### DIFF
--- a/internal/controller/bootstrap/providerid_controller.go
+++ b/internal/controller/bootstrap/providerid_controller.go
@@ -39,13 +39,14 @@ func (p *ProviderIDController) Reconcile(ctx context.Context, req ctrl.Request) 
 		return ctrl.Result{}, err
 	}
 
-	// Skip the control plane machines that don't have worker enabled
-	if capiutil.IsControlPlaneMachine(machine) && machine.ObjectMeta.Labels["k0smotron.io/control-plane-worker-enabled"] != "true" {
+	// Skip non-k0smotron managed machines
+	if machine.Spec.Bootstrap.ConfigRef.Kind != "K0sControllerConfig" && machine.Spec.Bootstrap.ConfigRef.Kind != "K0sWorkerConfig" &&
+		machine.Spec.InfrastructureRef.Kind != "RemoteMachine" {
 		return ctrl.Result{}, nil
 	}
 
-	// Skip non-k0s machines
-	if machine.Spec.Bootstrap.ConfigRef.Kind != "K0sControllerConfig" && machine.Spec.Bootstrap.ConfigRef.Kind != "K0sWorkerConfig" {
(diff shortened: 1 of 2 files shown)


internal/controller/bootstrap/providerid_controller.go
```diff
@@ -44,8 +44,9 @@ func (p *ProviderIDController) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
-	// Skip non-k0s machines
-	if machine.Spec.Bootstrap.ConfigRef.Kind != "K0sControllerConfig" && machine.Spec.Bootstrap.ConfigRef.Kind != "K0sWorkerConfig" {
+	// Skip non-k0smotron managed machines
```
apedriza, line 42: will the check `machine.ObjectMeta.Labels["k0smotron.io/control-plane-worker-enabled"] != "true"` allow reach the following lines? I think that label is only added when k0smotron controlplane provider is used

makhov: Right. Thanks, fixed!

internal/controller/bootstrap/providerid_controller.go
```diff
@@ -39,13 +39,14 @@ func (p *ProviderIDController) Reconcile(ctx context.Context, req ctrl.Request)
...
+	if machine.Spec.Bootstrap.ConfigRef.Kind != "K0sControllerConfig" && machine.Spec.Bootstrap.ConfigRef.Kind != "K0sWorkerConfig" &&
+		machine.Spec.InfrastructureRef.Kind != "RemoteMachine" {
 		return ctrl.Result{}, nil
 	}
 
-	// Skip non-k0s machines
-	if machine.Spec.Bootstrap.ConfigRef.Kind != "K0sControllerConfig" && machine.Spec.Bootstrap.ConfigRef.Kind != "K0sWorkerConfig" {
+	// Skip the control plane machines that don't have worker enabled
+	if machine.Spec.Bootstrap.ConfigRef.Kind != "K0sControllerConfig" && capiutil.IsControlPlaneMachine(machine) && machine.ObjectMeta.Labels["k0smotron.io/control-plane-worker-enabled"] != "true" {
```
apedriza:
```suggestion
	if machine.Spec.Bootstrap.ConfigRef.Kind == "K0sControllerConfig" && machine.ObjectMeta.Labels["k0smotron.io/control-plane-worker-enabled"] != "true" {
```
label check only for k0s controlplanes, right? also kind `K0sControllerConfig` guarantees the machine is a controlplane

makhov: thanks, fixed!

PR #1252, opened 2025-10-17
Delete machine before leaving etcd v2

makhov:


internal/controller/controlplane/k0s_controlplane_controller.go
apedriza, line 549: maybe worth to remove `runMachineDeletionSequence` now that there are only one step in this "sequence" and call directly `deleteMachine`

makhov: Let's do it in a separate PR. I want to make the behavior work, then we can improve the code :)

makhov:
> should we include in this changes a wait for the drain condition and that way we avoid manually draining on k0smotron side as https://github.com/k0sproject/k0smotron/pull/1220 does?

I still not sure, if it's a reliable way. I'll test it, but the current solution works and solves the issues we've seen with RecreteDeleteFirst strategy, so I want to proceed with the maybe imperfect but working solution asap and improve it later.

apedriza:
> I still not sure, if it's a reliable way. I'll test it, but the current solution works and solves the issues we've seen with RecreteDeleteFirst strategy, so I want to proceed with the maybe imperfect but working solution asap and improve it later.

Ah ok! I thought that the changes also had to do with the node drain thing. LGTM

PR #1258, opened 2025-10-22
Fix etcd cert secrets owner in remote host cluster mode

### LINKED ISSUE #1259: Etcd certificate Secrets immediately GC'd in remote host mode due to incorrect ownerReferences
### What happened?

In hosting cluster mode (when `Spec.KubeconfigRef` is set), etcd certificate Secrets are created with incorrect ownerReferences pointing to `k0smotron.io/Cluster`. This causes the Secrets to be immediately garbage collected after creation.

When watching with `kubectl get secret -n mgmt -w`, the following Secrets are repeatedly created and immediately GC'd:
- `<cluster>-etcd-server`
- `<cluster>-etcd-peer`
- `<cluster>-apiserver-etcd-client`

As a result, pods fail to start due to missing etcd certificate Secrets.

### Involved Areas

Cluster API integration

The etcd certificate Secrets should be created with correct ownerReferences and persist in the remote host cluster, allowing pods to start successfully.

### Steps to reproduce

1. Install k0smotron v1.8.1 via Cluster API
2. Create a Cluster with `Spec.KubeconfigRef` configured (hosting cluster mode)
3. Observe that pods fail to start
4. Run `kubectl get secret -n mgmt -w` to watch Secret creation
5. Observe Secrets being created and immediately garbage collected in a loop

### k0smotron version

v1.8.1

### k0s version

v1.33.2-k0s.0

This issue is specific to the hosting cluster scenario where control plane components run in an external cluster. The `ensureEtcdCertificates` function was setting ownerReferences that shouldn't exist for Secrets in the hosting cluster.

kahirokunn:
Fixes #1259

## Problem
  - `ensureEtcdCertificates` saved `*-etcd-server`, `*-etcd-peer`, and `*-apiserver-etcd-client` Secrets with an ownerReference pointing to `k0smotron.io/Cluster`.

## Result
The following Secrets are now correctly created and persist in the hosting cluster:
- `<cluster>-etcd-server`
- `<cluster>-etcd-peer`
- `<cluster>-apiserver-etcd-client`

Verified with:

`kubectl -n mgmt get secret` showing the three Secrets present
CleanShot 2025-10-22 at 15 07 49@2x
pod is running on external cluster
CleanShot 2025-10-22 at 15 08 20@2x
CleanShot 2025-10-22 at 15 08 44@2x

Cluster status progressing to Provisioned with ControlPlaneReady = True
CleanShot 2025-10-22 at 15 12 47@2x

cluster works good 👍 
CleanShot 2025-10-22 at 15 11 33@2x

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_certs.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_certs.go
@@ -34,6 +34,7 @@ import (
 	"sigs.k8s.io/cluster-api/util/secret"
 
 	km "github.com/k0sproject/k0smotron/api/k0smotron.io/v1beta1"
+	kutil "github.com/k0sproject/k0smotron/internal/controller/util"
 )
 
 func (scope *kmcScope) ensureEtcdCertificates(ctx context.Context, kmc *km.Cluster) error {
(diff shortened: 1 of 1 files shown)


kahirokunn:
That could be a good addition from a different perspective than the bug fix currently being worked on.
Therefore, I think it would be better to address that in a separate PR.

PR #1261, opened 2025-10-22
Add OWNERS file for approvals and reviews

kahirokunn:


kahirokunn:
Hi @makhov ,
Could you please check if the information are accurate?
Thank you! 🙏

kahirokunn:
@apedriza Hello! Could you please review this PR? 🙏

kahirokunn:
@apedriza 
Thanks for pointing that out!

As far as I understand, you’re right – having an OWNERS file only really takes effect once Prow (or a compatible automation) is running and configured for the repo. My intention with this PR was to start capturing the reviewer/approver structure explicitly, but I also see this as a good opportunity to discuss introducing Prow for this project.

Personally, I think setting up Prow would have a positive impact on the project’s growth: it would give us a more consistent review and approval workflow, make contributor onboarding easier, and align k0smotron with the broader Kubernetes ecosystem practices.

What do you think about adopting Prow for k0smotron, either now or in the near future?

kahirokunn:
@apedriza 
Thanks for pointing that out!

As far as I understand, you’re right – having an OWNERS file only really takes effect once Prow (or a compatible automation) is running and configured for the repo. My intention with this PR was to start capturing the reviewer/approver structure explicitly, but I also see this as a good opportunity to discuss introducing Prow for this project.

Personally, I think setting up Prow would have a positive impact on the project’s growth: it would give us a more consistent review and approval workflow, make contributor onboarding easier, and align k0smotron with the broader Kubernetes ecosystem practices.

What do you think about adopting Prow for k0smotron, either now or in the near future?

jnummelin:
> What do you think about adopting Prow for k0smotron, either now or in the near future?

Personally I'd delay that. IT does require some amount of effort to setup and maintain.

I do agree that it's good to have this clearly documented so people know who can review etc.

kahirokunn:
@jnummelin
Thanks for your feedback!

I completely understand your concern about the effort required to set up and maintain Prow. How about we take a phased approach:

1. **Merge the OWNERS file now** - This will serve as clear documentation of the review/approval structure, which benefits both current maintainers and new contributors
2. **Delay Prow integration** - We can revisit Prow adoption later when the project's needs and resources better align

Even without Prow automation, the OWNERS file provides value as:
- **Documentation**: Clear visibility of who can review and approve changes
- **Onboarding guide**: Helps new contributors understand the project structure
- **Future-ready**: When/if we decide to adopt Prow, we'll already have this foundation in place

This way, we get the documentation benefits immediately without the maintenance overhead. Does this approach sound reasonable to you?

jnummelin:
> How about we take a phased approach:

Sounds good to me. @makhov @apedriza WDYT?

kahirokunn:
Thank you 😊

PR #1271, opened 2025-10-27
Annotations and conditions added to the machines during InPlace update

makhov:
First step to support CAPI in-place updates
https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20240807-in-place-updates.md

### DIFF
--- a/internal/controller/controlplane/k0s_controlplane_controller.go
+++ b/internal/controller/controlplane/k0s_controlplane_controller.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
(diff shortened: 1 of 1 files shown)


apedriza:
As far as I understand, once we set the pending hook annotation, the machine controller will try to call the external extension. In our case we reuse our current inplace logic, which it is not through the CAPI Runtime Extension so I dont know if this might lead unexpected states in the machine status from the machine controller POV. For worker machines I guess we should implement the extension. Should we do same for controlplanes?

apedriza:
@makhov I think we can close this in favor of https://github.com/k0sproject/k0smotron/pull/1332, wdyt?

PR #1274, opened 2025-10-28
Use working dir for join token

makhov:
Override some file paths in RemoteMachine provisioner to use the working dir, eg `/etc/k0s.token` -> `/my/working/dir/k0s.token`

### DIFF
--- a/api/bootstrap/v1beta1/k0s_types.go
+++ b/api/bootstrap/v1beta1/k0s_types.go
@@ -20,6 +20,7 @@ import (
 	"github.com/k0sproject/k0smotron/internal/provisioner"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"path/filepath"
 
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
@@ -118,6 +119,9 @@ type K0sWorkerConfigSpec struct {
 	// SecretMetadata specifies metadata (labels and annotations) to be propagated to the bootstrap Secret.
 	// +kubebuilder:validation:Optional
 	SecretMetadata *SecretMetadata `json:"secretMetadata,omitempty"`
+
+	// WorkingDir specifies the working directory where k0smotron will place its files.
+	WorkingDir string `json:"workingDir,omitempty"`
 }
 
 // SecretMetadata defines metadata to be propagated to the bootstrap Secret
@@ -294,6 +298,9 @@ type K0sConfigSpec struct {
 	// See: https://cloudinit.readthedocs.io/en/latest/reference/merging.html
 	// +kubebuilder:validation:Optional
 	CustomUserDataRef *ContentSource `json:"customUserDataRef,omitempty"`
+
+	// WorkingDir specifies the working directory where k0smotron will place its files.
+	WorkingDir string `json:"workingDir,omitempty"`
 }
 
 type TunnelingSpec struct {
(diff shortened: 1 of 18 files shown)


makhov:
> We might simplify it and just use /etc as dir to place the k0s_install.sh

The issue is that sometimes we can't. In some secure setups users want to use another dir for those files, like `/opt/k0smotron` or something. 

I don't like this solution either, but not yet sure what is the best option to deal with that

apedriza:
> > We might simplify it and just use /etc as dir to place the k0s_install.sh
> 
> The issue is that sometimes we can't. In some secure setups users want to use another dir for those files, like `/opt/k0smotron` or something.
> 
> I don't like this solution either, but not yet sure what is the best option to deal with that

We could "move" the workingDir field to the bootstrap config and use it to create the bootstrap secret according to it if provided. That would require reading the controller config in the remote machine reconcile in order to use the workingDir if provided. That way user data is not updated after creating the secret, but implies api changes so current approach lgtm.

Maybe something to take into account if we definitely decide to promote k0smotron v1beta2.

makhov:
> We could "move" the workingDir field to the bootstrap config and use it to create the bootstrap secret according to it if provided. That would require reading the controller config in the remote machine reconcile in order to use the workingDir if provided. That way user data is not updated after creating the secret, but implies api changes so current approach lgtm.

Yeah, I was thinking about this approach as well, but not yet sure it worth it. So, the options:
- use it as it is in this PR
- add a workingDir field to the bootstrap config. might cause some confusion since we have it bot for bootstrap config and RemoteMachines
- read infra machine and use it from workingDir the RemoteMachine object. Also not really straighforward and increase coupling

@apedriza @jnummelin wdyt, which approach is the best?

apedriza:
> * add a workingDir field to the bootstrap config. might cause some confusion since we have it bot for bootstrap config and RemoteMachines

IMO maybe this + deprecation notice in the workingDir in the remotemachine, suggesting to use the bootstrap config workingDir. For backward compatibility with the current workingDir in remote machine we can use bootstrap config workingDir if remotemachine workingDir is not provided

makhov:
I agree but we can't just deprecate the RemoteMachine's one, since it's basically a dir where the script with the commands will be located when `commandsAsScript=true`. So it's kinda different things

apedriza:
> I agree but we can't just deprecate the RemoteMachine's one, since it's basically a dir where the script with the commands will be located when `commandsAsScript=true`. So it's kinda different things

But couldn't we also use the workDir from the bootstrap config for the remote machine? That way we could use it for the k0s config, for the token, and for the script and user data would not need to be changed so having support also for the workindDir field in remotemachine but trying to use the bootstrapconfig workindDir if remotemachine workingDir is not provided. 

Anyways, If finally is not possible I'm ok with the current proposal + maybe some comment in the remotemachine workingDir field mentioning it overrides the working dir for the user data commands and files.

makhov:
@apedriza I've added a new workingDir field to bootstrap configs, ptal

api/bootstrap/v1beta1/k0s_types.go
```diff
@@ -340,3 +347,35 @@ type IgnitionSpec struct {
 	// +kubebuilder:validation:Optional
 	AdditionalConfig string `json:"additionalConfig,omitempty"`
 }
+
+// GetDefaultK0sConfigPath returns the full path to the k0s.yaml file in the working directory.
+func (kcs *K0sConfigSpec) GetDefaultK0sConfigPath() string {
```
apedriza:
```suggestion
func (kcs *K0sConfigSpec) GetK0sConfigPath() string {
```

api/bootstrap/v1beta1/k0s_types.go
```diff
@@ -340,3 +347,35 @@ type IgnitionSpec struct {
...
+// GetJoinTokenPath returns the full path to the k0s token file in the working directory.
+func (kcs *K0sConfigSpec) GetJoinTokenPath() string {
+	if kcs.WorkingDir == "" {
+		return "/etc/k0s.token"
+	}
+	return filepath.Join(kcs.WorkingDir, "k0s.token")
+}
+
+// GetDefaultK0sConfigPath returns the full path to the k0s.yaml file in the working directory.
+func (k *K0sWorkerConfig) GetDefaultK0sConfigPath() string {
```
apedriza:
```suggestion
func (k *K0sWorkerConfig) GetK0sConfigPath() string {
```

apedriza (approved):
just a detail about the naming of the getter. LGTM

api/bootstrap/v1beta1/k0s_types.go
makhov: The thinking behind the `Default` was that we use it in this check, where we check for the default path. 
```
if arg != "--config" && arg != bootstrapConfig.Spec.GetDefaultK0sConfigPath() {
	argsWithoutConfig = append(argsWithoutConfig, arg)
}
```

apedriza: My thinking was that `GetDefaultK0sConfigPath` might return the workDir field value instead what I consider the default value which is `/etc/k0s.yaml`, that is why I suggested renaming to `GetK0sConfigPath` and encapsulate there that defaulting logic. Basically same concept followed in the `GetJoinTokenPath`

PR #1282, opened 2025-10-30
Add nightly e2e suite against aws infra provider

apedriza:
### Changes
- add workflow to exec ignition e2e using aws
- use OIDC as auth protocol
- Updated CloudFormation stack to include s3 actions because ignition user data stored in s3 requires them for the CAPA controller

Im not sure if excluding aws e2e to run it only at night is a good idea and not including it more often, but this certainly increases our current testing

Execution success before remove manual triggering of the workflow -> https://github.com/k0sproject/k0smotron/actions/runs/19037993893/job/54367814464

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -1,0 +1,149 @@
+name: Nightly E2E Tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 2 * * *'  # Runs every day at 02:00 UTC
+
+permissions:
+  id-token: write   # This is required for requesting the JWT
+  contents: read    # This is required for actions/checkout
+  
+jobs:
+  # TODO: instead of building the image for the nightly tests, we could cuse the one published to ttl.sh. If the image no longer exists there,
+  # the tests would fail, which is acceptable, meaning that the development version has already been tested against AWS the night before.
+  build:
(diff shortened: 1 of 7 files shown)


.github/workflows/e2e-nightly.yml
```diff
@@ -0,0 +1,151 @@
...
+      - name: Configure AWS credentials using OIDC
+        uses: aws-actions/configure-aws-credentials@v5
+        with:
+          role-to-assume: ${{ secrets.AWS_OIDC_ROLE_ARN }}
+          aws-region: eu-west-1
+
+      - name: Run e2e test
+        run: |
+          export AWS_B64ENCODED_CREDENTIALS=$(clusterawsadm bootstrap credentials encode-as-profile)
+          export TEST_NAME=Test$(echo "${{ matrix.e2e-suite }}" | awk -F'-' '{for(i=1;i<=NF;i++) $i=toupper(substr($i,1,1)) substr($i,2)}1' OFS='')
```
makhov: Could we just use the real test names in matrix to avoid this awk magic?

makhov (approved):

.github/workflows/e2e-nightly.yml
apedriza: Ok, I will set the manual triggering of the workflow to check everything is correct and merge this PR after that

PR #1289, opened 2025-11-10
Propagate RemoteMachine labels and annotations to the Machine

### LINKED ISSUE #1288: Sync label between k0smotron and CAPI resources

When provisioning nodes with k0smotron, some `PooledRemoteMachines` could have different hardware installed in them. 

One way to declaratively label nodes is to label the `MachineDeployment` and use the CAPI core controller `--additional-sync-machine-labels` flag to sync those between the `Machines` and `Nodes`.

I'm wondering if it would be possible to add a functionality to the `remote_machine_controller` to propagate the labels/annotations from a `RemoteMachine` to the underlying `Machine` object.

For example, let's say that I want to add two nodes in a cluster to use as storage nodes.
One node is configured with an nvme while the other is uing a hdd.

We could configure the hardware from the `PooledRemoteMachine` and use CAPI to sync the wanted labels down to the nodes on the workload cluster:
```yaml
kind: PooledRemoteMachine
metadata:
  name: node-a
  labels:
    storage.example.com/nvme: "true"
spec:
  pool: storage-nodes
  machine:
    address: node-a.example.com
    port: 22
    user: root
    sshKeyRef:
      name: ssh-key
---
kind: PooledRemoteMachine
metadata:
  name: node-b
  labels:
    storage.example.com/hdd: "true"
spec:
  pool: storage-nodes
  machine:
    address: node-b.example.com
    port: 22
    user: root
    sshKeyRef:
      name: ssh-key
```

While it is possible to achieve this by using different `Pools/MachineDeployment` to have a distinction between the two nodes, it would be easier and more flexible to sync the labels directly to the `Machine` object.

### Additional context

_No response_

makhov:
Fixes #1288

### DIFF
--- a/internal/controller/infrastructure/remote_machine_controller.go
+++ b/internal/controller/infrastructure/remote_machine_controller.go
@@ -282,9 +282,19 @@ func (r *RemoteMachineController) Reconcile(ctx context.Context, req ctrl.Reques
 			Address: rm.Spec.Address,
 		},
 	}
+	for l := range rm.Labels {
+		if _, ok := m.Labels[l]; !ok {
+			m.Labels[l] = rm.Labels[l]
+		}
+	}
+	for k := range rm.Annotations {
+		if _, ok := m.Annotations[k]; !ok {
+			m.Annotations[k] = rm.Annotations[k]
+		}
+	}
 
 	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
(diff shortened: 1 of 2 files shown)


makhov:
I'm not super happy with adding new fields and PooledRemoteMachine is a fully user-managed object, so using real labels is not much different from using a new field. I think propagating labels if they are not exist is safe

PR #1296, opened 2025-11-18
Unify k0s version format

apedriza:


apedriza:
Closed in favor of #1298

PR #1300, opened 2025-11-20
fix(entrypoint): escape amp in sed

### LINKED ISSUE #1299: Incorrect `sed` leads to corrupted kine data source
### What happened?

If a `Secret` with the kine data source has the following example DSN:

```
K0SMOTRON_KINE_DATASOURCE_URL: postgres://<USERNAME>:<PASS>@<HOST>/<DB>?sslmode=verify-ca&sslrootcert=%2Fvar%2Flib%2Fk0s%2Fkine-ca%2Fca.crt
```

then the following error occurs in the k0s during the execution of the entrypoint:

```
time="2025-11-20 08:13:07" level=info msg="time=\"2025-11-20T08:13:07.26945981Z\" level=error msg=\"failed to ping connection: cannot parse `postgres://<USERNAME>:<PASS>@<HOST>/<DB>?sslmode=verify-ca__K0SMOTRON_KINE_DATASOURCE_URL_PLACEHOLDER__sslrootcert%3D%2Fvar%2Flib%2Fk0s%2Fkine-ca%2Fca.crt`: failed to configure TLS (sslmode is invalid)\"" component=kine stream=stderr
```

The root cause is that the initial DSN has `&` amps and the `sed` expression [here](https://github.com/k0sproject/k0smotron/blob/2069e1fe86291b6ccf1d11c9c02881f354d79c17/internal/controller/k0smotron.io/k0smotroncluster_entrypoint.go#L116) incorrectly executes them, inserting the substitutee instead of the first `&` symbol. That also may cause the same issue if a password has amps as well.

To address the issue, the `&` symbol should be escaped

### Involved Areas

Standalone

The given DSN is being correctly passed to k0s without modifications

### Steps to reproduce

Create a `K0smotronControlPlane` with the `kineDataSourceSecretName` field set, where the `Secret` reference has a DSN containing one or more `&` symbols.

### k0smotron version

v1.9.0

### k0s version

v1.32.8+k0s.0

_No response_

zerospiel:
Escapes ampersand (&) symbol in the entrypoint.sh for the k0s, to avoid incorrect substitutions in case kine data source url has two or more options to connect.

Fixes #1299

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_entrypoint.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_entrypoint.go
@@ -113,7 +113,8 @@ const entrypointTemplate = `
 mkdir /etc/k0s && echo "$K0SMOTRON_K0S_YAML" > /etc/k0s/k0s.yaml
 
 # Substitute the kine datasource URL from the env var
-sed -i "s {{ .KineDataSourceURLPlaceholder }} ${K0SMOTRON_KINE_DATASOURCE_URL} g" /etc/k0s/k0s.yaml
+escaped_url=$(printf '%s' "$K0SMOTRON_KINE_DATASOURCE_URL" | sed 's/[&/\]/\\&/g')
+sed -i "s {{ .KineDataSourceURLPlaceholder }} $escaped_url g" /etc/k0s/k0s.yaml
 
 {{if .PrivilegedPortIsUsed}}
 apk add --no-cache libcap
--- a/internal/controller/k0smotron.io/k0smotroncluster_entrypoint_test.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_entrypoint_test.go
@@ -17,14 +17,21 @@ limitations under the License.
 package k0smotronio
 
 import (
+	"os"
+	"os/exec"
+	"strings"
 	"testing"
 
 	km "github.com/k0sproject/k0smotron/api/k0smotron.io/v1beta1"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
 func TestGetControllerFlags(t *testing.T) {
-	var tests = []struct {
+	tests := []struct {
 		name   string
 		kmc    km.Cluster
 		result string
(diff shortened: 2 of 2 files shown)


internal/controller/k0smotron.io/k0smotroncluster_entrypoint.go
```diff
@@ -113,7 +113,7 @@ const entrypointTemplate = `
 mkdir /etc/k0s && echo "$K0SMOTRON_K0S_YAML" > /etc/k0s/k0s.yaml
 
 # Substitute the kine datasource URL from the env var
-sed -i "s {{ .KineDataSourceURLPlaceholder }} ${K0SMOTRON_KINE_DATASOURCE_URL} g" /etc/k0s/k0s.yaml
+sed -i "s {{ .KineDataSourceURLPlaceholder }} ${K0SMOTRON_KINE_DATASOURCE_URL//&/\\&} g" /etc/k0s/k0s.yaml
```
makhov: Looks like it's not supported by posix shell, only in bash, that's why it fails on CI. We can switch it to
```
ESCAPED_K0SMOTRON_KINE_DATASOURCE_URL="$(printf '%s' "$K0SMOTRON_KINE_DATASOURCE_URL" | sed 's/&/\\&/g')"
sed -i "s {{ .KineDataSourceURLPlaceholder }} ${ESCAPED_K0SMOTRON_KINE_DATASOURCE_URL} g" /etc/k0s/k0s.yaml
```

zerospiel: That one is correct, though the CI found the problem with POSIX-incompatible line, I am now trying to do this with an extra line

The comment is outdated, yes, the probllem is with POSIX

zerospiel: Okay, seems the unit tests finally work 🎉

internal/controller/k0smotron.io/k0smotroncluster_entrypoint.go
```diff
@@ -113,7 +113,8 @@ const entrypointTemplate = `
 mkdir /etc/k0s && echo "$K0SMOTRON_K0S_YAML" > /etc/k0s/k0s.yaml
 
 # Substitute the kine datasource URL from the env var
-sed -i "s {{ .KineDataSourceURLPlaceholder }} ${K0SMOTRON_KINE_DATASOURCE_URL} g" /etc/k0s/k0s.yaml
+escaped_url=$(printf '%s' "$K0SMOTRON_KINE_DATASOURCE_URL" | sed 's/[&/\]/\\&/g')
+sed -i "s|{{ .KineDataSourceURLPlaceholder }}|$escaped_url|g" /etc/k0s/k0s.yaml
```
makhov: Why do we switch the separator? I think `|` unlike space might be in the password.

zerospiel: Initially it was a bit hard to debug the problem, so I decided to put it there. If spaces are okay in this context, let me revert it

Upd. Reverted

PR #1301, opened 2025-11-21
Metadata v1.10

makhov:


e2e/config/aws.yaml
```diff
@@ -105,7 +105,7 @@ providers:
         replacements:
           - old: "imagePullPolicy: Always"
             new: "imagePullPolicy: IfNotPresent"
-      - name: v1.9.99 # potentially next release. Manifest from source files (development) are used.
+      - name: v1.10.99 # potentially next release. Manifest from source files (development) are used.
```
apedriza, line 108: We forgot to add configs for 1.7 and 1.8 (and 1.9 after this changes). Also the to update e2e k0smotron-upgrade e2e. https://github.com/k0sproject/k0smotron/blob/main/e2e/k0smotron_upgrade_test.go#L52 (probably we should start removing the old ones for each version bump)

makhov: fixed

e2e/k0smotron_upgrade_test.go
```diff
@@ -49,7 +49,7 @@ func TestK0smotronUpgrade(t *testing.T) {
 // Important: This version MUST match the ones set in the e2e config file because that guarantees that the clusterctl
 // local repository contains the required versions of the providers when running the upgrade tests. We should test
 // development version against the latests stable version of k0smotron.
-var k0smotronMinorVersionsToCheckUpgrades = []string{"1.5", "1.6", "1.7", "1.8"}
+var k0smotronMinorVersionsToCheckUpgrades = []string{"1.5", "1.6", "1.7", "1.8", "1.9", "1.10"}
```
apedriza:
```suggestion
var k0smotronMinorVersionsToCheckUpgrades = []string{"1.6", "1.7", "1.8", "1.9", "1.10"}
```
now that 1.5 has been removed from the config file

apedriza (approved):

PR #1320, opened 2025-12-16
Add `kubeconfigSecretMetadata` field to configure kubeconfig secret m…

### LINKED ISSUE #1318: propagate CAPI cluster labels to kubeconfig secret

In order to expose more configurability to the worklaod kubeconfig secret, we can extend controlplane api to allow user to declare secret metadata.

Same approach we took for labels from workercofnig to data secret 
https://github.com/k0sproject/k0smotron/pull/1009

_No response_

### Additional context

_No response_

apedriza:
…etadata from controlplane resource

fix #1318


makhov:
Do we really need a separate field for this? Could we just propagate K0sControlPlane/K0smotronControlPlanes labels?

apedriza:
> Do we really need a separate field for this? Could we just propagate K0sControlPlane/K0smotronControlPlanes labels?

IMO this is like the template field for higher-level resources. Labeling/annotating resources with same values from parent resource does not allow us to control what we propagate. Seeing labels or annotations only related to the controlplane could be misleading. And also shares same approach we took in the past

PR #1325, opened 2025-12-23
Auto-create namespace in remote cluster for remote HCP

kahirokunn:
fixes: https://github.com/k0sproject/k0smotron/issues/1326

## What this PR does
This PR removes the manual prerequisite of creating the namespace in the hosting (remote) cluster when deploying Remote HCPs.

## Key changes
- Add a small utility that ensures a namespace exists in the remote cluster (GET first; if NotFound → Create).

## Why this is needed
Users currently hit reconciliation failures if the namespace is missing in the hosting cluster, e.g. `namespaces "<ns>" not found`. This PR makes the Remote HCP workflow smoother and less error-prone by automatically creating the namespace when required.

## Testing
The container image for this PR has been published at `quay.io/kahirokunn/k0smotron:auto-create-ns` and has been tested.

### DIFF
--- a/e2e/remote_hcp_test.go
+++ b/e2e/remote_hcp_test.go
@@ -72,15 +72,6 @@ func remoteHCPSpec(t *testing.T) {
 
 	// Setup a Namespace where to host objects for this spec and create a watcher for the namespace events.
 	namespace, _ := util.SetupSpecNamespace(ctx, testName, bootstrapClusterProxy, artifactFolder)
-	// Create same namespace in hosting cluster
-	nsForHostingCluster := &corev1.Namespace{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: namespace.GetName(),
-		},
-	}
-	err = hostingClusterProxy.GetClient().Create(ctx, nsForHostingCluster)
-	require.NoError(t, err)
-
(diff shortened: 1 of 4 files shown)


kahirokunn:
I would be grateful if someone could review it. 🙏

apedriza:
> From my POV, creating a resource in a non-existing namespace should fail like it fails for every resource whether `NamespaceAutoProvision` admission controller is not set. But for an external cluster it is a bit more tricky.
> 
> Ideally, we should run the creating logic in our admission webhook, creating the namespace and notifiyng the user with a warning about that a namespace does not exists in the external cluster so we are creating it.

This makes me think. What if external cluster gets privisioned with `NamespaceAutoProvision` flag?

kahirokunn:
Actually, NamespaceAutoProvision is already deprecated.
https://github.com/kubernetes/kubernetes/issues/31195

kahirokunn:
Tools like Sveltos or Argo CD also support auto-creating namespaces on target clusters (often enabled by default) to reduce friction when managing external resources.

kahirokunn:
Would you mind reconsidering?🙏

kahirokunn:
I'm so glad you said that! Thank you!

PR #1327, opened 2026-01-07
Control-plane endpoint reconcile fix

makhov:
Currently, we don't set control-plane endpoint after HCP is ready and In some setups it happens before we actually have a chance to set it

### DIFF
--- a/internal/controller/controlplane/k0smotron_controlplane_controller.go
+++ b/internal/controller/controlplane/k0smotron_controlplane_controller.go
@@ -204,7 +204,7 @@ func (c *K0smotronController) Reconcile(ctx context.Context, req ctrl.Request) (
 		return c.reconcileDelete(ctx, types.NamespacedName{Name: cluster.Name, Namespace: cluster.Namespace}, kcp)
 	}
 
-	res, ready, err := c.reconcile(ctx, cluster, kcp, kmcScope)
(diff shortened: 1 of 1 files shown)


internal/controller/controlplane/k0smotron_controlplane_controller.go
apedriza, line 207: maybe we can remove that non-used argument from the function signature

I copy link, I dont know what happened with the file review https://github.com/k0sproject/k0smotron/pull/1327/changes/BASE..70f0a19606321e51128673bdc6c5b565db79d6ff#diff-1b5c4513e09b9768b796da70ea34cd28e9613e2713c7b60dcbba46100061271eL207-R207

cmd/main.go
```diff
@@ -197,6 +197,7 @@ func main() {
 				DisableFor: []client.Object{
 					&corev1.ConfigMap{},
 					&corev1.Secret{},
+					&k0smotronv1beta1.Cluster{},
```
apedriza: why disable cache for kmc?

makhov: sorry, it's a leftover from my tests. removed it

internal/controller/controlplane/k0smotron_controlplane_controller.go
makhov: Done

PR #1332, opened 2026-01-14
Support CAPI in-place version updates

### LINKED ISSUE #1307: Support CAPI in-place upgrades in k0smotron
CAPI v1.12.0 will introduces support for in-place upgrades. Currently, k0smotron supports control plane in-place upgrades using our own strategy, where the control plane controller coordinates the upgrade via k0s autopilot. However, we still lack support for in-place upgrades for worker nodes.

CAPI offers a way to externalize the implementation of node updates via Runtime Extension. This extension must be implemented by k0smotron, implementing the interface proposed by CAPI for the correct integration of the controllers that need to communicate with the service responsible for the update. 

In order to adopt the mechanism proposed by CAPI for in-place upgrades k0smotron need changes/additions in two areas: in the control plane controller and in the creation of a webhook server responsible for the upgrade. 

## CONTROLPLANE CONTROLLER CHANGES

To preserve the existing user experience for control plane in-place upgrades, we should keep the same API field currently used to trigger in-place operations, and extend it to support the new CAPI in-place feature flag. If the feature flag is not enabled, we fall back to our current implementation; otherwise, we use the CAPI in-place mechanism. Worker node in-place updates can only go through the CAPI mechanism.

The control plane controller must handle the update of the machines one by one. To do this, once a new k0sversion is detected in the controlplane resource, the controller must:
- (Feature flag on in-place upgrades is active):
  - Iterate over the in-place extensions and check using the `CanUpdateMachine` endpoint if the update can be handled by the extension. If no extension can handle the update, fall back to a machine rollout.
  - If an in-place update extension can handle the update, mark the machine with the annotation hook `runtime.cluster.x-k8s.io/pending-hooks: ExternalUpdate`. This annotation will be received by the machine controller, which will iterate through the in-place update extensions by calling the `UpdateMachine` endpoint until an extension can handle the update. How the upgrade is handled by the webhook server is described below.
  - Once the extension has completed execution, the machine controller removes the hook annotation and then the control plane controller will set the `UpToDate: True` condition.
- (Feature flag on in-place upgrades is NOT active):
  - Current in-place implementation where the control plane controller handles upgrades by creating the Plan for all the machines as targets.

## CREATION OF A WEBHOOK SERVER UPGRADING THE NODE.

The update service is nothing more than an https server which will implement the following handlers:

- `CanUpdateMachine`: Respond to calls from controller owning the machine (controlplane or md controllers) about whether the node can be updated. 
- `UpdateMachine`: Machine controller will call this endpoint to try a node update. To do this, the server must create a Plan pointing only to the node that corresponds to the machine to be updated, specifying the target of that Plan. Currently, the Plan created in our implementation contains all the machines to be updated as targets. In this new case, since it is one by one, a plan must be created for each machine, and the control plane controller will coordinate the cleanup of any plan that may collide with the next in-place update.

### Tasks

- [ ] **Prerequisite**: Bump to CAPI v1.12.0
- [ ] Create Runtime Extension for in-place updates.
  - [ ] Implement a webhook server that provides a Runtime Extension for in-place updates. More info: https://cluster-api.sigs.k8s.io/tasks/experimental-features/runtime-sdk/implement-extensions
  - [ ] Include it in the release process. More info: https://cluster-api.sigs.k8s.io/tasks/experimental-features/runtime-sdk/deploy-runtime-extension
- [ ] Update control plane rollout logic to use the CAPI in-place mechanism when the feature flag is enabled. Fallback if not set to our current in-place mplementation for controlplane nodes
- [ ] Add e2e covering in-place upgrades
- [ ] Update docs/ create new post in k0s blog?

apedriza:
fix #1307


e2e/util/dump.go
```diff
@@ -36,19 +35,19 @@ func DumpSpecResourcesAndCleanup(ctx context.Context, specName string, clusterPr
...
 	if !skipCleanup {
-		err := deleteClusterAndWait(ctx, capiframework.DeleteClusterAndWaitInput{
-			ClusterProxy:   clusterProxy,
-			Cluster:        cluster,
-			ArtifactFolder: artifactFolder,
-		}, interval)
-		if err != nil {
-			fmt.Println(err.Error())
-		}
+		// err := deleteClusterAndWait(ctx, capiframework.DeleteClusterAndWaitInput{
```
makhov: some leftovers from the debug, I guess?

extensions/inplaceversionupdate/autopilot.go
```diff
@@ -0,0 +1,223 @@
...
+	default:
+		return false, nil
+	}
+}
+
+func createAutopilotPlanForMachine(ctx context.Context, clientset *kubernetes.Clientset, desiredMachine *clusterv1.Machine, isControlPlane bool) error {
+	// TODO: use DownloadURL from K0sControllerConfig/K0sWorkerConfig when available
+	amd64DownloadURL := `https://get.k0sproject.io/` + desiredMachine.Spec.Version + `/k0s-` + desiredMachine.Spec.Version + `-amd64`
+	arm64DownloadURL := `https://get.k0sproject.io/` + desiredMachine.Spec.Version + `/k0s-` + desiredMachine.Spec.Version + `-arm64`
+	armDownloadURL := `https://get.k0sproject.io/` + desiredMachine.Spec.Version + `/k0s-` + desiredMachine.Spec.Version + `-arm`
```
makhov: we should add and respect `downloadURL` as we do for control planes now

extensions/inplaceversionupdate/handler.go
```diff
@@ -0,0 +1,230 @@
...
+	return &InPlaceVersionUpdateHandler{
+		client: client,
+		decoder: serializer.NewCodecFactory(scheme).UniversalDecoder(
+			infrastructurev1beta1.GroupVersion,
+			bootstrapv1beta1.GroupVersion,
+		),
+	}
+}
+
+func DoCanUpdateMachine(ctx context.Context, req *runtimehooksv1.CanUpdateMachineRequest, resp *runtimehooksv1.CanUpdateMachineResponse) {
```
makhov, line 52: should we somehow check that workers can be updated only after controllers?

extensions/inplaceversionupdate/autopilot.go
```diff
@@ -0,0 +1,223 @@
...
+						},
+						"linux-arm": {
+							"url": "` + armDownloadURL + `"
+						}
+					},
+					"targets": {
+						"` + target + `": {
+							"discovery": {
+							    "static": {
+									"nodes": ["` + desiredMachine.Name + `"]
```
makhov: I don't really like the idea that we're creating a Plan per machine, but looks like this is how that Extension works. In future we might think and detect a higher object (like Cluster) changes and update the whole cluster at once

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -417,21 +426,32 @@ func (c *K0sController) reconcileMachines(ctx context.Context, cluster *clusterv
...
+		}
+		if m.Spec.Version == "" || !versionMatches(m, kcp.Spec.Version) {
 			clusterIsUpdating = true
 			if kcp.Spec.UpdateStrategy == cpv1beta1.UpdateInPlace {
 				desiredMachines.Insert(m)
+				notUpdatedMachines.Insert(m)
 			} else {
 				machineNamesToDelete[m.Name] = true
 			}
+			notUpdatedMachines.Insert(m)
```
makhov: Hm, we just added it in the `if` above. Do we need both lines?

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -474,6 +494,42 @@ func (c *K0sController) reconcileMachines(ctx context.Context, cluster *clusterv
...
 			}
 		} else {
+			isCAPIVersionSuitableForInplaceUpdate, err := isCAPIVersionSuitableForInplaceUpdate(ctx, c.Client)
+			if err != nil {
+				log.Log.Info("error checking CAPI version suitability for in-place updates: %w", err)
+			}
+
+			isK0smotronExtensionForInplaceUpdateDeployed, err := isK0smotronExtensionForInplaceUpdateDeployed(ctx, c.Client)
+			if err != nil {
+				log.Log.Info("error checking if k0smotron extension for in-place updates is deployed: %w", err)
```
makhov: log.Info doesn't support formatting, it accepts key and values
```
log.Log.Info("error checking if k0smotron extension for in-place updates is deployed", "error", err)
```

e2e/config/aws.yaml
```diff
@@ -117,6 +117,14 @@ providers:
           - old: "image: k0s/k0smotron:latest"
             new: "image: quay.io/k0sproject/k0smotron:latest"
 
+  - name: k0sproject-k0smotron
+    type: RuntimeExtensionProvider
+    versions:
+      - name: v1.10.99 # next; use manifest from source files
+        value: ../../extension/config/default
```
makhov: A typo in extension**s** ?

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -576,6 +632,84 @@ func (c *K0sController) reconcileMachines(ctx context.Context, cluster *clusterv
 	return nil
 }
 
+func isCAPIVersionSuitableForInplaceUpdate(ctx context.Context, c client.Client) (bool, error) {
+	var capiProvider clusterctlv1.Provider
+	err := c.Get(ctx, client.ObjectKey{Name: clusterctlconfig.ClusterAPIProviderName, Namespace: "capi-system"}, &capiProvider)
```
makhov: We need to check, there is a chance that k0rdent installs everything including CAPI providers and objects in `kcm-system` namespace, not standard `capi-system`

e2e/util/dump.go
apedriza: yes, fixed!

extensions/inplaceversionupdate/autopilot.go
apedriza: added the downloadurl from the config associated to the machine

extensions/inplaceversionupdate/handler.go
apedriza: mm I guess that is the general assumption for cluster updates in k8s

internal/controller/controlplane/k0s_controlplane_controller.go
apedriza: fixed!

internal/controller/controlplane/k0s_controlplane_controller.go
apedriza: fixed!

e2e/config/aws.yaml
apedriza: Ive removed aws configuration, we have docker infra atm

internal/controller/controlplane/k0s_controlplane_controller.go
apedriza: changed logic to list core providers in the cluster instead retrieve it from `capi-system`

extensions/inplaceversionupdate/autopilot.go
apedriza: agree, we should try to combine k0s plan declaring multiple nodes with this upgrade extension. We can think about it in the future but I dont know how from capi controllers POV this case can be done..mm

extensions/inplaceversionupdate/autopilot.go
```diff
@@ -0,0 +1,264 @@
...
+					"platforms": {
+						"linux-amd64": {
+							"url": "` + amd64DownloadURL + `"
+						},
+						"linux-arm64": {
+							"url": "` + arm64DownloadURL + `"
+						},
+						"linux-arm": {
+							"url": "` + armDownloadURL + `"
+						}
```
Danil-Grigorev: Hey @apedriza, is this possible to override default behavior for in-place upgrader? To invoke some command on a host for k0s download or similar?

The use case is such that it needs to support actual per-arch resolution of the OCI artifact from airgap registry (custom), and would use something similar to `oras pull` with a artifact URL and a tag, instead of direct fetch of the blob, with verify of the hash sum of the artifact on download. Plus the registry access might require supplying custom certificates for download.

I see that currently in-place updater undertakes all `version` field updates, and executes it via autopilot plan. I can't tell if download of k0s is all it does, or it also performs specific commands to start a new version under the hood, but if it is possible - can this download logic be delegated to external mechanism, without disabling in-place upgrader logic entirely?

extensions/inplaceversionupdate/autopilot.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, fmt.Errorf("error generating %s transport config: %w", cluster.Name, err)
+	}
+	tlsCfg, err := transport.TLSConfigFor(tCfg)
+	if err != nil {
+		return nil, fmt.Errorf("error generating %s tls config: %w", cluster.Name, err)
+	}
+
+	// Disable keep-alive to avoid hanging connections
+	cl := http.DefaultClient
+	cl.Transport = &http.Transport{
```
makhov: I don't think mutating a global `http.Client` is a good idea, let's just create a new one

internal/autopilot/autopilot.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	switch state {
+	case "Completed":
+		return true, nil
+	// Following processing states declared in:
+	// https://docs.k0sproject.io/stable/autopilot-multicommand/?h=schedulablewait#processing-states
+	case "NewPlan", "Schedulable", "SchedulableWait":
+		return false, nil
+	// Any other state is considered as not completed.
```
makhov, line 152: Can we do something about it? Any error will cause an infinite loop, right?

extensions/inplaceversionupdate/autopilot.go
apedriza: It seems a bit out of the k0smtoron scope. From what I understood, you need to download in your node based in some criteria which I think it could be done out of k0smotron. Then you could deploy a local server with your downloaded artifact to still use the current valid http schema for autopilot downlaoder controller

.github/workflows/go.yml
```diff
@@ -90,6 +90,16 @@ jobs:
...
           path: install-standalone.yaml
+      
+      - name: Generate inplace-extension install yaml
+        run: |
+          make release-extension-webhook
+
+      - name: Upload inplace-extension yaml
+        uses: actions/upload-artifact@v4
+        with:
+          name: install-extension-webhook-yaml
```
makhov, line 96: should it be a separate YAML file or a part of our install.yaml?

docs/update/update-capi-cluster.md
```diff
@@ -45,6 +45,27 @@ k0smotron supports three update strategies, configured via `spec.updateStrategy`
 !!! warning
     The `RecreateDeleteFirst` strategy requires at least 3 control plane nodes.
 
+### Use InPlace strategy
+
+Under the `InPlace` strategy, k0smotron drives the update through one of two mechanisms, depending on what's available in your management cluster:
+
+- **k0s autopilot**: the default path. k0smotron orchestrates the update itself using [k0s autopilot](https://docs.k0sproject.io/stable/autopilot/), independently of Cluster API's machine lifecycle.
```
makhov: Not sure I'm fully understand this doc. Both are using autopilot underneath, right? Who decide which one to use? What is the difference from the user's point of view?

docs/update/update-capi-cluster.md
```diff
@@ -45,6 +45,27 @@ k0smotron supports three update strategies, configured via `spec.updateStrategy`
...
+- **Cluster API in-place updates** (experimental): k0smotron instead delegates the rollout to Cluster API's own [in-place updates mechanism](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20240807-in-place-updates.md). This path is used automatically once all of the following are satisfied:
+
+  - Cluster API core **v1.12.0 or newer**.
+  
+  - The `InPlaceUpdates` feature gate enabled on Cluster API core (`EXP_IN_PLACE_UPDATES=true`).
+  
+  - The k0smotron in-place version update extension installed in the management cluster:
+
+      ```bash
+      kubectl apply --server-side=true -f https://docs.k0smotron.io/{{{ extra.k0smotron_version }}}/extension-inplace-version-update.yaml
```
makhov: In makefile it generates a file with some other name iirc

extensions/inplaceversionupdate/autopilot.go
apedriza: fixed!

internal/autopilot/autopilot.go
apedriza: Now we emit an error message when unexpected state is set on the plan. CAPI should report properly the state as machine conditions

.github/workflows/go.yml
apedriza: The extension should be installed separately, as it can be installed using clusterctl using the `--runtime-extension` flag. Also, deploying a webhook server always seems unneeded for cases where the user does not want to use the feature

docs/update/update-capi-cluster.md
apedriza: I tried to improve it the docs by explaining both approaches and what means. I hope it is clearer. Anyway, happy to keep improving this with any suggestions

docs/update/update-capi-cluster.md
apedriza: Fixed, thanks!

extensions/inplaceversionupdate/autopilot.go
```diff
@@ -0,0 +1,132 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/transport"
+	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	"sigs.k8s.io/cluster-api/util/kubeconfig"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func createAutopilotPlanForMachine(ctx context.Context, c client.Client, clientset *kubernetes.Clientset, desiredMachine *clusterv1.Machine, isControlPlane bool) error {
+	downloadURL, err := getDownloadURLFromK0sConfig(ctx, c, client.ObjectKeyFromObject(desiredMachine), isControlPlane)
```
makhov: shouldn't we use `desiredMachine.Spec.Bootstrap.ConfigRef.Name` for `bootstrapObjKey`, not `desiredMachine`? Machine names are generated, no?

extensions/inplaceversionupdate/autopilot.go
```diff
@@ -0,0 +1,132 @@
...
+package inplaceversionupdate
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/cloudflare/cfssl/log"
```
makhov: wrong logger?

extensions/inplaceversionupdate/autopilot.go
apedriza: changed!

extensions/inplaceversionupdate/autopilot.go
apedriza: yes, removed logger usage there, thanks!

e2e/util/machinedeployment.go
```diff
@@ -0,0 +1,122 @@
...
+	input.MachineDeployment.Spec.Template.Spec.Version = input.KubernetesUpgradeVersion
+
+	err = wait.PollUntilContextTimeout(ctx, time.Second, time.Minute, true, func(ctx context.Context) (done bool, err error) {
+		return patchHelper.Patch(ctx, input.MachineDeployment) == nil, nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to patch the new kubernetes version to MachineDeployment %s: %w", klog.KObj(input.MachineDeployment), err)
+	}
+
+	fmt.Println("Waiting for MachineDeployment to be upgraded")
```
makhov, line 97: Should it be `fmt.Println` or logger?

extensions/inplaceversionupdate/autopilot.go
```diff
@@ -0,0 +1,139 @@
...
+
+	err = autopilot.CreatePlan(ctx, clientset, planParams)
+	if err != nil {
+		return fmt.Errorf("error creating autopilot plan: %w", err)
+	}
+
+	return nil
+}
+
+func getWorkloadClusterClientset(ctx context.Context, client client.Client, cluster client.ObjectKey) (*kubernetes.Clientset, error) {
```
makhov: We do have a helper for that already iirc

extensions/inplaceversionupdate/handler.go
```diff
@@ -0,0 +1,270 @@
...
+	client ctrlclient.Client
+}
+
+func NewInPlaceVersionUpdateHandler(client ctrlclient.Client) *InPlaceVersionUpdateHandler {
+	return &InPlaceVersionUpdateHandler{
+		client: client,
+	}
+}
+
+func DoCanUpdateMachine(ctx context.Context, req *runtimehooksv1.CanUpdateMachineRequest, resp *runtimehooksv1.CanUpdateMachineResponse) {
```
makhov, line 52: It's like handler, right? should we check that it's k0smotron and not for example kubeadm cluster?

extensions/inplaceversionupdate/handler.go
```diff
@@ -0,0 +1,270 @@
...
+		return
+	}
+
+	// When the completed plan does not target the current machine, we need to delete the plan and create a new one
+	// for the current machine. Otherwise, keep the latest plan for historical purposes.
+	// This webhook server creates a plan targeting only one machine at a time, following the CAPI contract, but we
+	// need also take into account cases where the plan is manage by external mechanisms, like the standalone
+	// in-place update in k0smotron for control plane nodes or even autopilot plans created by other tools/cluster
+	// operators.
+	if !slices.Contains(currentPlanTargetNodes, desiredMachine.Name) {
```
makhov: We don't check the version here. Is it possible that the user will try to update cluster first from v1.33 to v1.34, then to v1.35 almost immediately and we will report that upgrade is done (but it's done only to v1.34)

apedriza: Added a condition which checks if the current plan target version is the desired for the machine

extensions/inplaceversionupdate/handler.go
apedriza: Added a check for only update machines bootstrapped with k0s, thanks!

extensions/inplaceversionupdate/autopilot.go
apedriza: I've refactor the helper in order to reuse it as the func used clsutercache controller but this webhook server does not runs within a manager

e2e/util/machinedeployment.go
apedriza: All the utilities in the package use fmt for logging, and to change all of them would require changes not very related to the PR itself, plus possibly some refactor to use the testing logger or another logger. I would change it in a follow-up PR

extensions/config/webhook/deployment.yaml
```diff
@@ -0,0 +1,41 @@
+---
+apiVersion: apps/v1
```
makhov, line 2: Sorry, I'm not fully understand the process. Why do we need a separate deployment? 
Also, I see that ExtensionConfig is used in tests only, do we need it in our install process?

apedriza: A runtime extension like this is a mechanism the user may or may not decide to install, so to use it you need to deploy its manifest and have the corresponding feature flag set. This extension handler only makes sense deployed alongside k0smotron, but in theory any provider can make use of an extension. It is up to them decide what suits them and what not. In the end, this implementation adds a new runtime provider, so it shouldn't be tied to k0smotron's own lifecycle, hence the separate deployment.

> I see that ExtensionConfig is used in tests only, do we need it in our install process?

I have included the ExtensionConfig, I was missing it, thanks!

makhov (approved):
I'm still on the fence about a separate deployment since it overcomplicates things a lot. On the other hand it's an experimental feature and it's better to keep it separately.

apedriza:
> I'm still on the fence about a separate deployment since it overcomplicates things a lot. On the other hand it's an experimental feature and it's better to keep it separately.

We can always reduce the steps later by including the installation into the k0smotron deployment if the future is needed. Better to start by asking more of the user and simplify afterward rather than doing it the other way around I think

Issue #1319, opened 2025-12-16
Make ReadinessProbe and LivenessProbe timings configurable via K0smotronCluster and allow templated overrides for StatefulSet container spec


kahirokunn:
### Is your feature request related to a problem? Please describe.

Currently the controller hard-codes probe settings in `internal/controller/k0smotron.io/k0smotroncluster_statefulset.go` (`ReadinessProbe.InitialDelaySeconds=60`, `LivenessProbe.InitialDelaySeconds=90`, plus other values). This makes it difficult to tune boot/startup behavior for different environments (slow storage/VMs, constrained CPU, etc.), and can lead to unnecessary restarts or delayed readiness.

### Describe the solution you would like

Add support to configure these probe parameters from the custom resource (e.g., `K0smotronCluster.spec`), at minimum `readinessProbe.initialDelaySeconds` and `livenessProbe.initialDelaySeconds`.  
Ideally, provide a more general “override/template” mechanism so users can tune additional StatefulSet/container parameters in the future, without needing code changes for each new knob.

### Additional context

The current hard-coded values are here:
- `ReadinessProbe.InitialDelaySeconds: 60`
- `LivenessProbe.InitialDelaySeconds: 90`

It would be great if users could provide a template/override block in the CR to replace/merge probe settings (and potentially other container spec fields) to support different runtime characteristics across clusters.

kahirokunn:
I'm looking to combine KEDA with k0smotron to scale pods to zero. The goal is to spin up the control plane when users try to access it, and keep it at zero cost when idle, similar to how Knative works. Wouldn't that be amazing?
To make this work effectively, I'd like to keep the initialDelaySeconds as small as possible to minimize startup time.

kahirokunn:
keep

kahirokunn:
Keep

kahirokunn:
Complete ✅

PR #1333, opened 2026-01-15
feat(k0s): add PodTemplate field for control plane pod customization

kahirokunn:
fixes: #1319

## Description

This PR improves k0s control plane pod customization in k0smotron by **embedding the entire Kubernetes `PodTemplateSpec`** under the k0smotron cluster spec, instead of adding a growing list of individual pod/probe fields. This addresses the need to tune boot/startup behavior (e.g., readiness/liveness probe delays) across different environments while keeping the API flexible for future pod-level customization.

## What's Changed

- Added **`spec.podTemplate`** (`corev1.PodTemplateSpec`) to the k0smotron cluster/controlplane spec.

## Testing

- Added unit tests covering PodTemplate merge behavior for the k0s control plane StatefulSet:
  - overriding readinessProbe initialDelaySeconds
  - overriding livenessProbe initialDelaySeconds
  - verifying the StatefulSet template hash changes when probe settings change via `podTemplate`
- Regenerated CRDs and resource reference docs.

## Example Configuration

```yaml
kind: Cluster
metadata:
  name: docker-test
  namespace: default
spec:
  clusterNetwork:
    pods:
      cidrBlocks:
      - 192.168.0.0/16
    serviceDomain: cluster.local
    services:
      cidrBlocks:
      - 10.128.0.0/12
  controlPlaneRef:
    kind: K0smotronControlPlane
    name: docker-test-cp
    namespace: default
  infrastructureRef:
    kind: DockerCluster
    name: docker-test
    namespace: default
---
kind: DockerCluster
metadata:
  name: docker-test
  namespace: default
  annotations:
    cluster.x-k8s.io/managed-by: k0smotron
spec: {}
---
kind: K0smotronControlPlane
metadata:
  name: docker-test-cp
  namespace: default
spec:
  version: v1.34.3-k0s.0
  service:
    type: LoadBalancer
    apiPort: 6443
    konnectivityPort: 8132
  podTemplate:
    metadata:
      annotations:
        skip-kyverno-policy-validation: "true"
    spec:
      nodeSelector:
        kubernetes.io/arch: arm64
      tolerations:
        - key: "high-stability"
          operator: "Exists"
          effect: "NoSchedule"
      containers:
        - name: controller
          readinessProbe:
            initialDelaySeconds: 5
          livenessProbe:
            initialDelaySeconds: 5
```

## Before/After Comparison

The following shows the generated k0s control plane StatefulSet with and without the `podTemplate` configuration from the example above:

```diff
--- Without podTemplate
+++ With podTemplate
diff --git a/sts.yaml b/sts.yaml
index f396a97e..8636b082 100644
--- a/sts.yaml
+++ b/sts.yaml
@@ -1,206 +1,215 @@
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   annotations:
+    skip-kyverno-policy-validation: "true"
   generation: 1
   labels:
     app: k0smotron
     cluster: docker-test
     cluster.x-k8s.io/cluster-name: docker-test
     cluster.x-k8s.io/control-plane: "true"
     component: cluster
   name: kmc-docker-test
   namespace: default
 spec:
   persistentVolumeClaimRetentionPolicy:
     whenDeleted: Retain
     whenScaled: Retain
   podManagementPolicy: OrderedReady
   replicas: 1
   revisionHistoryLimit: 10
   selector:
     matchLabels:
       app: k0smotron
       cluster: docker-test
       cluster.x-k8s.io/cluster-name: docker-test
       cluster.x-k8s.io/control-plane: "true"
       component: cluster
   serviceName: ""
   template:
     metadata:
+      annotations:
+        skip-kyverno-policy-validation: "true"
       labels:
         app: k0smotron
         cluster: docker-test
         cluster.x-k8s.io/cluster-name: docker-test
         cluster.x-k8s.io/control-plane: "true"
         component: cluster
     spec:
       affinity:
         podAntiAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:
           - podAffinityTerm:
               labelSelector:
                 matchLabels:
                   app: k0smotron
                   cluster: docker-test
               topologyKey: topology.kubernetes.io/zone
             weight: 100
           - podAffinityTerm:
               labelSelector:
                 matchLabels:
                   app: k0smotron
                   cluster: docker-test
               topologyKey: kubernetes.io/hostname
             weight: 50
       automountServiceAccountToken: false
       containers:
       - args:
         - /bin/sh
         - -c
         - /k0smotron-entrypoint.sh
         envFrom:
         - configMapRef:
             name: kmc-docker-test-config
         image: quay.io/k0sproject/k0s:v1.34.3-k0s.0
         imagePullPolicy: IfNotPresent
         livenessProbe:
           exec:
             command:
             - k0s
             - status
           failureThreshold: 10
-          initialDelaySeconds: 90
+          initialDelaySeconds: 5
           periodSeconds: 10
           successThreshold: 1
           timeoutSeconds: 1
         name: controller
         ports:
         - containerPort: 6443
           name: api
           protocol: TCP
         - containerPort: 8132
           name: konnectivity
           protocol: TCP
         readinessProbe:
           exec:
             command:
             - k0s
             - status
           failureThreshold: 15
-          initialDelaySeconds: 60
+          initialDelaySeconds: 5
           periodSeconds: 10
           successThreshold: 1
           timeoutSeconds: 1
         resources: {}
         terminationMessagePath: /dev/termination-log
         terminationMessagePolicy: File
         volumeMounts:
         - mountPath: /k0smotron-entrypoint.sh
           name: kmc-entrypoint-docker-test-config
           subPath: k0smotron-entrypoint.sh
         - mountPath: /var/lib/k0s
           name: kmc-docker-test
         - mountPath: /var/lib/k0s/manifests/k0s-telemetry
           name: kmc-docker-test-telemetry-config
           readOnly: true
       dnsPolicy: ClusterFirst
       initContainers:
       - command:
         - sh
         - -c
         - mkdir -p /var/lib/k0s/pki && rm -rf /var/lib/k0s/pki/server.* && cp /certs-init/*.*
           /var/lib/k0s/pki/
         image: quay.io/k0sproject/k0s:v1.34.3-k0s.0
         imagePullPolicy: IfNotPresent
         name: certs-init
         resources: {}
         terminationMessagePath: /dev/termination-log
         terminationMessagePolicy: File
         volumeMounts:
         - mountPath: /certs-init
           name: certs
         - mountPath: /var/lib/k0s
           name: kmc-docker-test
+      nodeSelector:
+        kubernetes.io/arch: arm64
       restartPolicy: Always
       schedulerName: default-scheduler
       securityContext: {}
       terminationGracePeriodSeconds: 30
+      tolerations:
+      - effect: NoSchedule
+        key: high-stability
+        operator: Exists
       volumes:
       - configMap:
           defaultMode: 484
           items:
           - key: k0smotron-entrypoint.sh
             path: k0smotron-entrypoint.sh
           name: kmc-entrypoint-docker-test-config
         name: kmc-entrypoint-docker-test-config
       - name: certs
         projected:
           defaultMode: 420
           sources:
           - secret:
               items:
               - key: tls.crt
                 path: ca.crt
               - key: tls.key
                 path: ca.key
               name: docker-test-ca
           - secret:
               items:
               - key: tls.crt
                 path: front-proxy-ca.crt
               - key: tls.key
                 path: front-proxy-ca.key
               name: docker-test-proxy
           - secret:
               items:
               - key: tls.crt
                 path: sa.pub
               - key: tls.key
                 path: sa.key
               name: docker-test-sa
           - secret:
               items:
               - key: tls.crt
                 path: etcd-ca.crt
               - key: tls.key
                 path: etcd-ca.key
               name: docker-test-etcd
           - secret:
               items:
               - key: tls.crt
                 path: apiserver-etcd-client.crt
               - key: tls.key
                 path: apiserver-etcd-client.key
               name: docker-test-apiserver-etcd-client
       - emptyDir: {}
         name: kmc-docker-test
       - configMap:
           defaultMode: 420
           name: kmc-docker-test-telemetry-config
         name: kmc-docker-test-telemetry-config
   updateStrategy:
     rollingUpdate:
       maxUnavailable: 1
       partition: 0
     type: RollingUpdate
```

## Use Cases

This enables several deployment scenarios without needing new API fields for each one:

- **Slow storage / constrained CPU**: tune readiness/liveness delays to avoid premature restarts
- **Multi-tenant clusters**: tolerations / affinity to run control planes on dedicated nodes
- **High availability**: topology spread constraints / affinity tuning
- **Performance optimization**: nodeSelector / runtimeClassName to target specific node pools/runtimes

## Backwards Compatibility

- `podTemplate` is **additive and optional**, so existing manifests continue to work unchanged.
- No breaking changes to existing APIs or behavior.

## Other

To avoid conflicts, I'm working directly on this PR's branch.


apedriza:
IMO embedding the pod template from the user can be good for total customization of the controlplane but also can start being a source of issues for configurations that are not set well so we should be very careful with a change like this. There are currently other fields in the spec doing the same (forwarded to the pod template) for example, `spec.image`, `spec.version`, `spec.mounts`.. and probably others. Probably we should think about some deprecation policy and eventually transition to the podtemplate approach

kahirokunn:
Yes, I completely agree with your concerns about the pod template approach.

Regarding how to handle this carefully, I see four possible approaches:

1. Ask users to take responsibility themselves and provide clear documentation
2. Implement a validating webhook to reject invalid configurations
3. Accept the configuration initially but surface errors in the status
4. Allow users to provide a podTemplate but selectively override it (whitelist approach) only for fields that would definitely break functionality

I believe option 2 (validating webhook) would be the best approach. This would allow us to gradually evolve the policy over time. By the way, since writing validating webhooks in Go can be tedious, I often use Kyverno to write policies instead. Kyverno makes this remarkably easy and straightforward.

I'm very interested in moving the podTemplate feature forward. Is there anything specific I should be doing in the short term to help advance this? I believe this feature can address a very wide range of requirements with minimal effort, while also being highly intuitive and familiar to users.

Looking forward to your thoughts.

jnummelin:
While I fully agree that users do need a way to customize the pods we create, I'm not entirely sure embedding a full podTemplate is the right way. I think it creates few problems for users.

So say we now depend (in go.mod) on kubernetes 1.32.x, the podTemplate we get is from 1.32. So say user is running on 1.35 cluster, if 1.35 removed any fields that were available on 1.32, we'd fail to create the pods. Also vice-versa, user couldn't use any newer features either since those would not be exposed in our CRDs.

Second concern is that this adds so much of things in CRDs which might create some challenges for the users.


Another approach would be to utilize something similar as [kubevirt](https://kubevirt.io/user-guide/cluster_admin/customize_components/) and others do:
```
  customizeComponents:
    patches:
    - resourceType: Deployment
      resourceName: virt-controller
      patch: '[{"op": "remove", "path": "/spec/template/spec/containers/0/livenessProbe"}]'
      type: json
    - resourceType: Deployment
      resourceName: virt-controller
      patch: '{"metadata":{"annotations":{"patch": "true"}}}'
      type: strategic
```

One added benefit of this approach is that it explicit about the oder of things, these provided patches are applied on top of what k0smotron creates. While in this PR we do document how things work, it's not really explicit for the user. And let's face it, we engineers rarely read docs. 😂 

@kahirokunn @apedriza @makhov WDYT?

kahirokunn:
@jnummelin Thanks for the suggestion on `customizeComponents`. I think it’s a solid direction.

Would it make sense to scope it so users can customize not only the k0s API server and etcd, but also Services and any other resources we generate? Having a single, consistent entry point to patch resources we own feels straightforward and user-friendly.

If everyone agrees with this direction, I’d be happy to help and can open a new PR to move it forward. Please let me know.

jnummelin:
> Would it make sense to scope it so users can customize not only the k0s API server and etcd, but also Services and any other resources we generate?

Yes, makes sense. I think the challenge is that how does the user target the specific resource. I mean it's not really easy for users to target by name since the names are generated by k0smotron. Although the names are predictable. 🤷 

I'd really like @apedriza and @makhov to chime in with the approach before we start the implementation.

apedriza:
> > Would it make sense to scope it so users can customize not only the k0s API server and etcd, but also Services and any other resources we generate?
> 
> Yes, makes sense. I think the challenge is that how does the user target the specific resource. I mean it's not really easy for users to target by name since the names are generated by k0smotron. Although the names are predictable. 🤷
> 
> I'd really like @apedriza and @makhov to chime in with the approach before we start the implementation.

we could use selectors, [same approach follow in CAPI ClusterClass](https://pkg.go.dev/sigs.k8s.io/cluster-api@v1.12.1/api/core/v1beta1#PatchSelector). So resource type + selector + patch should be enough to target the desired resource and operation

kahirokunn:
Thank you for the suggestion about using the CAPI ClusterClass approach with selectors!

While I think selectors are definitely part of the solution, I'm concerned that the CAPI ClusterClass approach may have some limitations when applied to k0smotron's use case.

In CAPI, the resource kinds that can be patched are structurally distinct, which makes selectors sufficient. However, in k0smotron's case, we generate multiple resources of the same kind. For example, if a user wants to patch a Service resource, which Service would be targeted? The kube-apiserver Service? The etcd Service? What about any additional Services we might generate in the future?

Without a way to distinguish between these resources beyond just their kind and labels, I think we could end up with ambiguity that makes the API unpredictable for users.

That's why I believe it's important to include a predictable `name` or some other unique identifier in the selector mechanism, so users can target specific resources reliably. Something like:

```yaml
  customizeComponents:
    patches:
    - resourceType: Deployment
      resourceName: {{ .Cluster.Name }}-kube-apiserver
      patch: '[{"op": "remove", "path": "/spec/template/spec/containers/0/livenessProbe"}]'
      type: json
    - resourceType: Deployment
      resourceName: {{ .Cluster.Name }}-etcd
      patch: '{"metadata":{"annotations":{"patch": "true"}}}'
      type: strategic
```

This would give us the flexibility of the patch-based approach while maintaining clarity about which exact resource is being modified.

What do you think about incorporating resource names alongside the selector approach? Would that address the concerns while still keeping the API intuitive?

apedriza:
> Thank you for the suggestion about using the CAPI ClusterClass approach with selectors!
> 
> While I think selectors are definitely part of the solution, I'm concerned that the CAPI ClusterClass approach may have some limitations when applied to k0smotron's use case.
> 
> In CAPI, the resource kinds that can be patched are structurally distinct, which makes selectors sufficient. However, in k0smotron's case, we generate multiple resources of the same kind. For example, if a user wants to patch a Service resource, which Service would be targeted? The kube-apiserver Service? The etcd Service? What about any additional Services we might generate in the future?
> 
> Without a way to distinguish between these resources beyond just their kind and labels, I think we could end up with ambiguity that makes the API unpredictable for users.
> 
> That's why I believe it's important to include a predictable `name` or some other unique identifier in the selector mechanism, so users can target specific resources reliably. Something like:
> 
> ```yaml
>   customizeComponents:
>     patches:
>     - resourceType: Deployment
>       resourceName: {{ .Cluster.Name }}-kube-apiserver
>       patch: '[{"op": "remove", "path": "/spec/template/spec/containers/0/livenessProbe"}]'
>       type: json
>     - resourceType: Deployment
>       resourceName: {{ .Cluster.Name }}-etcd
>       patch: '{"metadata":{"annotations":{"patch": "true"}}}'
>       type: strategic
> ```
> 
> This would give us the flexibility of the patch-based approach while maintaining clarity about which exact resource is being modified.
> 
> What do you think about incorporating resource names alongside the selector approach? Would that address the concerns while still keeping the API intuitive?

Yes, as you said, selectors do not allow target-specific resources unless we add proper labels identifying each resource created (which could be another option). But using selectors for targeting only one resource does not feel very useful from my point of view, seems use the resource name is what we need. 

If we follow the name for targeting resources, we need to add together with the changes proper documentation about all resources and its names created by k0smotron. Do we have a list of resources created by k0smotron?

kahirokunn:
@apedriza I put together a page listing the resources created by a k0smotron Cluster and their naming conventions, and opened **#1341** for it. I also added a small **draft doc for `customizeComponents`** in the same PR — mainly to make the discussion more concrete and help us align on the API/UX before implementation.
If you have a moment, could we continue the discussion and align on the approach in #1341?

apedriza:
@kahirokunn can we close this in favor of the patches approach?

kahirokunn:
yes 🙏

PR #1339, opened 2026-01-26
v1beta2

### LINKED ISSUE #1336: v1beta2 API
To implement Windows support we need to introduce new `spec.provisioner` struct to let end-user decide which user-data format to use, which effectively deprecates `spec.ignition` and moves it to `spec.provisioner.ignition`. 

That forces us to introduce new API version (probably `v1beta2`).

We definitely have some other fields that should be moved or changed. This issue is to collect those fields and to keep track the progress.

Proposed changes:
- `K0s[Controller|Worker]Config.spec.ignition`  -> `spec.provisioner.ignition`
- `K0s[Controller|Worker]Config.spec.customUserDataRef`  -> `spec.provisioner.customUserDataRef`
- `[pre|post]StartCommands` -> `[pre|post]K0sCommands`

makhov:
Fixes #1336


makhov:
@apedriza Good idea. I've changed the status fields

api/bootstrap/v1beta2/worker_bootstrap_webhook.go
apedriza, line 1: why to move webhooks inside `api` folder for both, bootstrap and controlplane providers?

api/bootstrap/v1beta1/k0s_worker_config_conversion.go
```diff
@@ -0,0 +1,118 @@
...
+	dst := dstRaw.(*v1beta2.K0sWorkerConfig)
+	dst.ObjectMeta = k.ObjectMeta
+	dst.Spec = k0sWorkerConfigV1beta1ToV1beta2Spec(k.Spec)
+	dst.Status = v1beta2.K0sWorkerConfigStatus{
+		Ready:          k.Status.Ready,
+		DataSecretName: k.Status.DataSecretName,
+		Conditions:     k.Status.Conditions,
+	}
+	if k.Status.DataSecretName != nil && *k.Status.DataSecretName != "" {
+		dst.Status.Initialization.DataSecretCreated = true
```
apedriza: same as the comment in the `k0scontrollerconfig` conversion file

api/bootstrap/v1beta2/k0s_types.go
```diff
@@ -0,0 +1,545 @@
...
+	Initialization StatusInitialization `json:"initialization,omitempty"`
+
+	// Conditions defines current service state of the K0sWorkerConfig.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// StatusInitialization represents the initialization status of the worker node
+type StatusInitialization struct {
+	DataSecretCreated bool `json:"dataSecretCreated,omitempty"`
```
apedriza: this should be a pointer to bool per CAPI contract

api/bootstrap/v1beta2/k0s_types.go
```diff
@@ -0,0 +1,545 @@
...
+}
+
+// K0sWorkerConfigStatus defines the observed state of K0sWorkerConfig
+type K0sWorkerConfigStatus struct {
+	// Ready indicates the Bootstrapdata field is ready to be consumed
+	Ready bool `json:"ready,omitempty"`
+
+	// DataSecretName is the name of the secret that stores the bootstrap data script.
+	// +optional
+	DataSecretName *string `json:"dataSecretName,omitempty"`
```
apedriza: `ready` and `datasecretname` should be removed per CAPI contract and change conversion accordingly

api/bootstrap/v1beta2/k0s_types.go
```diff
@@ -0,0 +1,545 @@
...
+
+// K0sControllerConfigStatus defines the observed state of K0sControllerConfig
+type K0sControllerConfigStatus struct {
+	// Ready indicates the Bootstrapdata field is ready to be consumed
+	Ready bool `json:"ready,omitempty"`
+
+	// DataSecretName is the name of the secret that stores the bootstrap data script.
+	// +optional
+	DataSecretName *string `json:"dataSecretName,omitempty"`
+
```
apedriza: `ready` and `datasecretname` should be removed per CAPI contract and change conversion accordingly

config/clusterapi/all/crd/patches/webhook_in_k0scontrollerconfig.yaml
apedriza, line 1: If using clusterctl I only install controlplane provider the conversion will not work I think because currently it is only set for the full-providers installation. I would locate the patches peer provider and then gather all of the in the `all` mode as we do with the crds

api/controlplane/v1beta2/k0smotron_types.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Hub marks K0smotronControlPlaneList as a conversion hub.
+func (*K0smotronControlPlaneList) Hub() {}
+
+// Initialization represents the initialization status of the control plane
+type Initialization struct {
+	// controlPlaneInitialized indicates whether the control plane is initialized
+	// +optional
+	ControlPlaneInitialized bool `json:"controlPlaneInitialized"`
```
apedriza: pointer to bool pero CAPI contract

apedriza (commented):
Should we also bump infra provider version just for simplicity/being aligned with the others? There are also changes in the RemoteCluster and RemoteMachine that should be done per CAPI contract so we could use v1beta2 to implement too

apedriza:
I also would suggest to remove the warnings in th webhook about +k0s.0 suffix and just reject the request starting from v1beta2 for simplicity in our code. We currently mutate the version in the k0scontrolplane and k0smotroncontrolplane controllers so we could remove that logic and expect alwys the proper k0s version format

config/clusterapi/all/crd/patches/webhook_in_k0scontrollerconfig.yaml
makhov: There is the same patch in separate providers yamls as well. When I try to directly reference them, I get 
```
`: security; file '/Users/amakhov/work/k0smotron/config/clusterapi/controlplane/patches/webhook_in_k0scontrolplanes.yaml' is not in or below '/Users/amakhov/work/k0smotron/config/clusterapi/all/crd'

```

makhov:
@apedriza I hope I addressed all the comments. I've created a new branch as we discussed to keep main clean.

Regarding suffix warnings, I suggest doing it in a separate PR. This one is quite big and complex already.

PR #1341, opened 2026-01-28
Add patches API for patching generated resources

kahirokunn:
## Description

Add a `patches` API that allows users to customize k0smotron-generated Kubernetes resources using standard patch types (JSON Patch, Strategic Merge Patch, Merge Patch).

Users can target specific resources by `Kind` and `app.kubernetes.io/component` label, then apply patches to modify things like resource limits, annotations, probe settings, and more — without forking or modifying the controller itself.

### Example

```yaml
kind: K0smotronControlPlane
metadata:
  name: docker-test-cp
  namespace: default
spec:
  version: v1.34.3-k0s.0
  persistence:
    type: emptyDir
  service:
    type: LoadBalancer
 patches:
  # Set resource requests/limits and probe timing on the control-plane pods
  - target:
      kind: StatefulSet
      component: control-plane
    patch:
      type: strategic
      content: |
        spec:
          template:
            spec:
              containers:
                - name: controller
                  resources:
                    requests:
                      memory: "256Mi"
                      cpu: "250m"
                    limits:
                      memory: "512Mi"
                      cpu: "500m"
                  readinessProbe:
                    initialDelaySeconds: 10
                  livenessProbe:
                    initialDelaySeconds: 10
  # Increase the etcd readiness probe initial delay
  - target:
      kind: StatefulSet
      component: etcd
    patch:
      type: json
      content: '[{"op":"add","path":"/spec/template/spec/containers/0/readinessProbe/initialDelaySeconds","value":15}]'
  # Add an annotation to the control-plane Service
  - target:
      kind: Service
      component: control-plane
    patch:
      type: merge
      content: |
        metadata:
          annotations:
            example.com/managed-by: "k0smotron-customized"
```

Relates to #1333

## Test plan

- Unit tests added in `internal/controller/util/patch_test.go`
- Manually tested


kahirokunn:
I applied the following yaml:

```yaml
kind: Cluster
metadata:
  name: docker-test
  namespace: default
spec:
  clusterNetwork:
    pods:
      cidrBlocks:
      - 192.168.0.0/16
    serviceDomain: cluster.local
    services:
      cidrBlocks:
      - 10.128.0.0/12
  controlPlaneRef:
    kind: K0smotronControlPlane
    name: docker-test-cp
    namespace: default
  infrastructureRef:
    kind: DockerCluster
    name: docker-test
    namespace: default
---
kind: K0smotronControlPlane # This is the config for the controlplane
metadata:
  name: docker-test-cp
  namespace: default
spec:
  version: v1.27.2-k0s.0
  persistence:
    type: emptyDir
  service:
    type: NodePort
---
kind: DockerCluster
metadata:
  name: docker-test
  namespace: default
  annotations:
    cluster.x-k8s.io/managed-by: k0smotron # This marks the base infra to be self managed. The value of the annotation is irrelevant, as long as there is a value.
spec: {}
  # More details of the DockerCluster can be set here
```

Then I ran the following commands:

```sh
$ kubectl krew install tree
$ kubectl tree K0smotronControlPlane docker-test-cp
W0128 10:57:58.693716   29375 warnings.go:70] v1 Endpoints is deprecated in v1.33+; use discovery.k8s.io/v1 EndpointSlice
NAMESPACE  NAME                                                      READY  REASON  AGE
default    K0smotronControlPlane/docker-test-cp                      -              13m
default    ├─Cluster/docker-test                                     -              13m
default    │ ├─ConfigMap/kmc-docker-test-config                      -              13m
default    │ ├─ConfigMap/kmc-docker-test-telemetry-config            -              13m
default    │ ├─ConfigMap/kmc-entrypoint-docker-test-config           -              13m
default    │ ├─Secret/docker-test-apiserver-etcd-client              -              13m
default    │ ├─Secret/docker-test-etcd-peer                          -              13m
default    │ ├─Secret/docker-test-etcd-server                        -              13m
default    │ ├─Secret/docker-test-kubeconfig                         -              11m
default    │ ├─Service/kmc-docker-test-etcd                          -              13m
default    │ │ └─EndpointSlice/kmc-docker-test-etcd-w5675            -              13m
default    │ ├─Service/kmc-docker-test-nodeport                      -              13m
default    │ │ └─EndpointSlice/kmc-docker-test-nodeport-hhgql        -              13m
default    │ ├─StatefulSet/kmc-docker-test                           -              13m
default    │ │ ├─ControllerRevision/kmc-docker-test-79688cf866       -              13m
default    │ │ └─Pod/kmc-docker-test-0                               True           13m
default    │ └─StatefulSet/kmc-docker-test-etcd                      -              13m
default    │   ├─ControllerRevision/kmc-docker-test-etcd-6b44c99f5c  -              13m
default    │   └─Pod/kmc-docker-test-etcd-0                          True           13m
default    ├─Secret/docker-test-ca                                   -              13m
default    ├─Secret/docker-test-etcd                                 -              13m
default    ├─Secret/docker-test-proxy                                -              13m
default    └─Secret/docker-test-sa                                   -              13m
```

This document is based on the above output.

kahirokunn:
However, I haven't yet verified Ingress mode, so I'm not very confident about that documentation. This is because I realized that readinessProbe/livenessProbe can become a bottleneck for subsequent launches.
Nevertheless, I believe it's possible to promote this feature even without detailed knowledge of it.

docs/generated-resources.md
```diff
@@ -0,0 +1,364 @@
...
+
+This page is written for **users who need to know the names and naming patterns of resources** created by K0smotron Clusters, such as when customizing deployments via `customizeComponents.patches`. It intentionally focuses on:
+
+- resource **names and name patterns**
+- **where** resources are created (management/external/workload cluster)
+- **when** resources are created (conditions / feature flags)
+
+It does **not** describe controller implementation details (for example, internal reconciliation logic), since those tend to change over time and are not needed for patching.
+
+## Naming Conventions Overview
```
apedriza, line 17: I think use names for targeting resources is good but we should ensure every resource created by k0smotron follows same schema: `kmc-<cluster-name>-<suffix>`. Found that entrypoint configmap does not follow it. Also, some names we currently set seems to be redundant like add `config` name in the configmaps created. Im thinking in allow the user target the resource for the patch with something like:

(if we want to avoid redundant suffix)
```yaml
resourceType: StatefulSets
purpose: etcd
```
With this, we should ensure every resource created has a descriptive suffix for the `purpose` field 

OR

(verbose option)
```yaml
resourceType: StatefulSets
name: kmc-mycluster-etc
```

In both cases we shoudl ensure that every resouce follow the schema `kmc-<cluster-name>-<suffix>`

docs/generated-resources.md
```diff
@@ -0,0 +1,364 @@
...
+kind: K0smotronControlPlane
+metadata:
+  name: docker-test-cp
+spec:
+  customizeComponents:
+    patches:
+    - resourceType: StatefulSet
+      resourceNameSuffix: ""  # Empty string for no suffix → kmc-<cluster-name>
+      patch: '[{"op": "replace", "path": "/spec/template/spec/containers/0/readinessProbe/initialDelaySeconds", "value": 10}]'
+      type: json
```
apedriza: we would be supporting merge, json and strategic patch types like kubectl does?

docs/generated-resources.md
```diff
@@ -0,0 +1,364 @@
+# Generated Resources and Naming Conventions
+
+This document describes all Kubernetes resources created and managed by a `k0smotron.io/v1beta1` Cluster (K0smotron Cluster) and their naming conventions. This information is essential when you need to customize these resources.
+
+When using Cluster API (CAPI), a `K0smotronControlPlane` resource creates a child `k0smotron.io/v1beta1` Cluster, which then produces the resources documented here.
```
apedriza: Patches should be supported also for k0smotron standalone. So I would restructured docs about the option of patching kmc resources either  using CAPI integration with k0smtoroncontrolplane or directly in standalone mode where user defines the patches in the Cluster resource directly

apedriza (commented):
I would split the documentation about resources created by k0smotron in one page and another page talking about the patches feature.

jnummelin:
I think for targeting, we definitely don't wanna support Go templates. They're gonna get messy, both for maintainers and for users.

I see two options:

#### Hardcoded "components"
Like what @apedriza also suggested. We'd maintain, and document, the list of type & component _things_ we create.
```
resourceType: StatefulSets
purpose: etcd
```

#### Dynamic "components" via labels

We'd need to introduce some common label for each resource we create and user could target that with the resource type. Effectively this is the same as the other proposal, the only difference is this is bit more "dynamic" as we do not have to hardcode those components.

So say we add this to e.g. etcd manifests:
```
labels:
  app.kubernetes.io/component: etcd
```
User could target that with:
```
resourceType: StatefulSets
component: etcd
```

docs/generated-resources.md
kahirokunn: Thank you 🙏 
I wrote the documentation using Dynamic “components” via labels.
How does this look?
https://github.com/k0sproject/k0smotron/pull/1341#issuecomment-3817706375

docs/generated-resources.md
kahirokunn: Yes, that's the plan. It looks like KubeVirt is doing the same.

https://github.com/kubevirt/kubevirt/blob/main/staging/src/kubevirt.io/api/core/v1/types.go#L2581-L2585

docs/generated-resources.md
kahirokunn: I've updated the documentation accordingly. What do you think?

kahirokunn:
I think the dynamic "components" via labels is a great idea. I've rewritten the documentation accordingly. Could everyone please take another look and re-review if possible?

kahirokunn:
> I would split the documentation about resources created by k0smotron in one page and another page talking about the patches feature.

This has also been addressed.

kahirokunn:
@apedriza 
If it's alright with you, let's align our understanding.
The following PR has been merged:
https://github.com/k0sproject/k0smotron/pull/1343

For the next step, I should just make minor adjustments to the documentation included in the current PR to match the merged content, correct?

apedriza:
> For the next step, I should just make minor adjustments to the documentation included in the current PR to match the merged content, correct?

Just to simplify things, I would create in the same PR the new patch feature using a selector + the documentation needed about it.

kahirokunn:
Thanks for the suggestion, @apedriza! I've gone ahead and implemented the new patch feature with the selector approach in this same PR, along with the updated documentation to match. I've also done various manual tests locally, and everything is working as expected so far. Would you be able to take a look when you get a chance? I'd really appreciate your feedback. Let me know if anything needs to be adjusted!

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -121,8 +121,49 @@ type ClusterSpec struct {
 	// Note: This metadata will have precedence over default labels/annotations on the Secret.
 	// +kubebuilder:validation:Optional
 	KubeconfigSecretMetadata SecretMetadata `json:"kubeconfigSecretMetadata,omitempty,omitzero"`
+	// CustomizeComponents defines patches to apply to generated resources (StatefulSet, Service, ConfigMap, etc.).
+	// Patches are applied after generation and before apply. Target resources are matched by Kind and app.kubernetes.io/component label.
```
apedriza: We can point documentation about generated resources here

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -121,8 +121,49 @@ type ClusterSpec struct {
...
+	// +kubebuilder:validation:Required
+	ResourceType string `json:"resourceType"`
+	// Component is the value of the app.kubernetes.io/component label on the target resource.
+	// +kubebuilder:validation:Required
+	Component string `json:"component"`
+	// Type is the patch type to apply.
+	// +kubebuilder:validation:Enum=json;strategic;merge
+	// +kubebuilder:validation:Required
+	Type PatchType `json:"type"`
+	// Patch is the patch content. The format depends on the Type field.
```
apedriza: We should be more transparent about what each type means (maybe pointing some docs to the standard) and what is the patch content expected on each one

internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -438,13 +438,20 @@ func (c *K0smotronController) computeStatus(ctx context.Context, cluster *cluste
 	err := c.Client.Get(ctx, types.NamespacedName{Name: cluster.Name, Namespace: cluster.Namespace}, &kmc)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			// The cluster is not yet created.
+			conditions.Set(kcp, metav1.Condition{
+				Type:    cpv1beta1.ChildClusterReconciledCondition,
+				Status:  metav1.ConditionFalse,
+				Reason:  cpv1beta1.ReconciliationInProgressReason,
+				Message: "Child k0smotron Cluster has not been created yet",
```
apedriza:
```suggestion
				Message: "k0smotron Cluster has not been created yet",
```

from my POV child cluster and k0smotron cluster is not the same and can be confused. Rather, the k0smotron Cluster resource creates a child cluster. There are more assumptions like this one that should be changes too this if we agree in the change

internal/controller/controlplane/k0smotron_controlplane_controller.go
apedriza, line 1: Is it necessary to introduce new conditions? I would keep this PR as atomic as possible, given the changes it adds

internal/controller/util/patch.go
```diff
@@ -0,0 +1,107 @@
...
+		switch p.Type {
+		case km.JSONPatchType:
+			currentData, err = applyJSONPatch(currentData, patchJSON)
+		case km.StrategicMergePatchType:
+			currentData, err = applyStrategicMergePatch(scheme, &gvks[0], currentData, patchJSON)
+		case km.MergePatchType:
+			currentData, err = applyMergePatch(currentData, patchJSON)
+		default:
+			return fmt.Errorf("unknown patch type %q for patch at index %d", p.Type, i)
+		}
```
apedriza, line 77: we should prevent this branch by adding some validation in the webhook. Anyways, returning this error is good.

docs/capi-advanced/components.md
```diff
@@ -0,0 +1,239 @@
+# Component customization
+
+This page explains how to customize Kubernetes resources generated by k0smotron.
+
+## Overview
+
+k0smotron creates the Kubernetes resources (for example `StatefulSet`, `Service`, and `ConfigMap`) documented in [Generated Resources](../generated-resources.md) through a `k0smotron.io/v1beta1` Cluster. You can customize those generated resources by defining patches.
+
+- **Standalone**: You create the `k0smotron.io/v1beta1` Cluster directly. See [Create a cluster](../cluster.md).
+- **Cluster API integration**: When k0smotron acts as a control plane provider (in-cluster), `K0smotronControlPlane` creates a child `k0smotron.io/v1beta1` Cluster, which then produces the same set of generated resources. See [k0smotron as Cluster API provider](../cluster-api.md).
```
apedriza:
```suggestion
- **Cluster API integration**: When k0smotron acts as a control plane provider (in-cluster), `K0smotronControlPlane` creates a `k0smotron.io/v1beta1` Cluster, which then produces the same set of generated resources. See [k0smotron as Cluster API provider](../cluster-api.md).
```

docs/capi-advanced/components.md
```diff
@@ -0,0 +1,239 @@
...
+In both modes, the resources that can be customized are the same, and you define patches in the same field (`.spec.customizeComponents.patches`); only the resource that has that spec differs (Cluster in standalone, K0smotronControlPlane in Cluster API).
+
+## Where to define patches
+
+Define patches in `spec.customizeComponents.patches` on the resource that owns the generated resources:
+
+| Mode | Resource | Field |
+| ------ | ---------- | -------- |
+| **Standalone** | `k0smotron.io/v1beta1` Cluster | `Cluster.spec.customizeComponents.patches` |
+| **Cluster API (control plane provider - in-cluster)** | `K0smotronControlPlane` | `K0smotronControlPlane.spec.customizeComponents.patches` |
```
apedriza: would be great point here to references the api for each resource

docs/capi-advanced/components.md
```diff
@@ -0,0 +1,239 @@
...
+
+This avoids string-based targeting via names and avoids template-based targeting. See the [Examples](#examples) above.
+
+## Patch types
+
+The patch format depends on `type`:
+
+- `json`: a JSON Patch array (RFC 6902)
+- `strategic`: a strategic merge patch object
+- `merge`: a JSON merge patch object (RFC 7386)
```
apedriza: some links for better explanation of them?

docs/capi-advanced/components.md
```diff
@@ -0,0 +1,239 @@
...
+- `etcd`: etcd resources (StatefulSet, Service, CronJob) when not using Kine
+- `cluster-config`: k0s configuration ConfigMap
+- `entrypoint`: entrypoint script ConfigMap
+- `telemetry`: telemetry configuration ConfigMap
+- `kubeconfig`: kubeconfig Secret
+- `monitoring`: monitoring ConfigMaps (when monitoring is enabled)
+- `ingress`: ingress-related resources, including konnectivity ConfigMaps (when ingress is enabled)
+- `tunneling`: FRP tunneling resources (when tunneling is enabled)
+
+If you need the full list of generated resources and their names, see [Generated Resources](../generated-resources.md).
```
apedriza: I would just point the docs about the naming per component and avoid another list here

docs/capi-advanced/components.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Notes and limitations
+
+- Patches are applied **in list order** after k0smotron generates each resource and before it is applied to the cluster. Multiple patches that match the same resource (same `resourceType` and `component`) are applied in sequence.
+- A patch is applied to all generated resources that match the given `resourceType` and `component` combination.
```
apedriza: Shouldn't be unique? I mean, is there any resource sharing type and component label value?

apedriza (commented):
Some comments here. I need to test it a bit

apedriza:
Also, please update the PR title according to the feature instead the docs

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
kahirokunn: Done. Added a link to the generated resources documentation in the godoc.

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
kahirokunn: Done. Added RFC/Kubernetes doc links and expected patch content format for each type in the Type and Patch field godocs.

internal/controller/controlplane/k0smotron_controlplane_controller.go
kahirokunn: Done. Applied the suggestion and updated all similar messages and comments.

internal/controller/util/patch.go
kahirokunn: Done. Added validateCustomizeComponents in ValidateClusterSpec to reject invalid patch types at admission time.

docs/capi-advanced/components.md
kahirokunn: Done. Added links to the API reference pages for both Cluster and K0smotronControlPlane.

docs/capi-advanced/components.md
kahirokunn: Done. Added links to RFC 6902, RFC 7386, and the Kubernetes strategic merge patch documentation.

docs/capi-advanced/components.md
kahirokunn: Done. Replaced the component list with a reference to the Generated Resources page.

docs/capi-advanced/components.md
kahirokunn: it's not always unique (e.g. component: monitoring is shared by two ConfigMaps). Added a note clarifying this behavior.

internal/controller/controlplane/k0smotron_controlplane_controller.go
kahirokunn: This has been separated into a PR.
https://github.com/k0sproject/k0smotron/pull/1364

kahirokunn:
@apedriza Thanks for your review 🙏 
I've addressed all the points you raised and rebased the code.
Could you please review it again?

docs/capi-advanced/components.md
kahirokunn: Done. Thank you 🙏

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -121,8 +121,59 @@ type ClusterSpec struct {
 	// Note: This metadata will have precedence over default labels/annotations on the Secret.
 	// +kubebuilder:validation:Optional
 	KubeconfigSecretMetadata SecretMetadata `json:"kubeconfigSecretMetadata,omitempty,omitzero"`
+	// CustomizeComponents defines patches to apply to generated resources (StatefulSet, Service, ConfigMap, etc.).
+	// Patches are applied after generation and before apply. Target resources are matched by Kind and app.kubernetes.io/component label.
+	// For the full list of generated resources and their component labels, see https://docs.k0smotron.io/stable/generated-resources/.
+	// +kubebuilder:validation:Optional
+	CustomizeComponents CustomizeComponents `json:"customizeComponents,omitempty"`
```
makhov: What is the purpose of `customizeComponents` object? Why not add `patches` directly here?

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -121,8 +121,59 @@ type ClusterSpec struct {
...
 
+// CustomizeComponents defines the customization of generated resources.
+type CustomizeComponents struct {
+	// Patches is a list of patches to apply to generated resources. Patches are applied in order.
+	// +kubebuilder:validation:Optional
+	Patches []ComponentPatch `json:"patches,omitempty"`
+}
+
+// ComponentPatch defines a patch to apply to a generated resource.
+type ComponentPatch struct {
```
makhov: I'd suggest having struct something like this:
```
type ComponentPatch struct {
	Selector MySelectorStruct `json:"selector"`
	Patch MyPatchStruct `json:"patch"`
}
```

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
```diff
@@ -121,8 +121,59 @@ type ClusterSpec struct {
...
+}
+
+// PatchType defines the type of patch to apply.
+// +kubebuilder:validation:Enum=json;strategic;merge
+type PatchType string
+
+const (
+	// JSONPatchType is RFC 6902 JSON Patch (array of operations).
+	// See https://datatracker.ietf.org/doc/html/rfc6902 for more details.
+	JSONPatchType PatchType = "json"
```
makhov: Do we need to support all the patch types? I mean, it's nice to have more features than less in general, but we need to test, maintain, and probably validate all of them

kahirokunn: Thanks for the thoughtful review! That's a fair concern.

All three patch types are covered by unit tests, including positive cases for each type, YAML input, error handling, and non-matching patches.

The implementation itself delegates directly to well-established libraries (jsonpatch, strategicpatch, and standard JSON merge patch from apimachinery), so the actual patch logic we maintain is just the
routing/matching in ApplyComponentPatches. The maintenance surface is quite small.

I think supporting all three is worth it because they serve different use cases:
- merge: simplest, good for most cases (adding labels, annotations, changing scalar fields)
- strategic: needed when you want Kubernetes-aware array merge semantics (e.g. appending a container or volume without replacing the whole list)
- json: needed for precise operations like removing a specific field or array element

Dropping any of them would leave a real gap in what users can express. Happy to add more tests if you'd like though, let me know!

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
kahirokunn: Great suggestion, done! Restructured `ComponentPatch` into `Target` + `Patch` sub-structs as you suggested.

Before:
```yaml
patches:
  - resourceType: StatefulSet
    component: etcd
    type: strategic
    patch: |
      ...
```

After:
```yaml
patches:
  - target:
      kind: StatefulSet
      component: etcd
    patch:
      type: strategic
      content: |
        ...
```

Much cleaner separation of concerns. Thanks for the guidance!

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
kahirokunn: Thank you, done. Renamed customizeComponents to patches and promoted it directly into ClusterSpec.

internal/controller/util/patch.go
```diff
@@ -0,0 +1,107 @@
...
+func applyStrategicMergePatch(scheme *runtime.Scheme, gvk *schema.GroupVersionKind, original, patchData []byte) ([]byte, error) {
+	empty, err := scheme.New(*gvk)
+	if err != nil {
+		return nil, fmt.Errorf("create empty struct for strategic merge: %w", err)
+	}
+	return strategicpatch.StrategicMergePatch(original, patchData, empty)
+}
+
+func applyMergePatch(original, patchData []byte) ([]byte, error) {
+	return jsonpatch.MergeMergePatches(original, patchData)
```
apedriza: can we use `MergePatch` instead? In this case we have a document (generated resource) and the patch

internal/controller/k0smotron.io/k0smotroncluster_monitoring_config.go
```diff
@@ -78,6 +79,10 @@ func (scope *kmcScope) reconcileMonitoringCM(ctx context.Context, kmc *km.Cluste
 		return err
 	}
 
+	if err := kcontrollerutil.ApplyComponentPatches(scope.client.Scheme(), &cm, kmc.Spec.Patches); err != nil {
```
apedriza: nit: I'm thinking in cases where we might forget to use this method for apply patches. Maybe we should encapsulate the creation of resources within a wrapper which apply patches first. Same issue we have right now for owner references when using external cluster. I think using a proper linter to not allow calling the controller-runtime client in this package might help. Nothing to block the changes, just thinking aloud

apedriza (commented):
would be great to add a dummy patch in some of the existing e2e flavors using `k0smotroncontrolplane` and a small check in the spec about everything is good. Just extend a bit some of the existing e2e specs, maybe `remote_hcp_test.go` is good for it atm.

apedriza:
Another point is if we should push this together with v1beta2..? adding new fields does not need to be in a new api version but maybe this is a quite big change in the k0smotron API

makhov:
I think this PR should target `v1beta2-api-version` branch and the changes done in v1beta2 api version

kahirokunn:
@makhov Is the v1beta2-api-version branch already in a state where it currently works? 👀

makhov:
Depends on what do you mean by "works". We're collecting all the work on new types there to make main clean in case we need to release some fixes before this work is done. Once it's done we'll merge it to main and make release. But all tests passes, k0smotron actually works in that branch, etc. 

Although adding a new field doesn't require a version change, it seems logical to put it to the new version since we have it anyway. And new version means new big minor (or maybe even major, we haven't decided yet) release, and this feature is also big and deserve to be a part of a big release.

internal/controller/util/patch.go
kahirokunn: Thank you. I fixed it.

kahirokunn:
Understood. I rebase on the v1beta2-api-version branch, so I'll run it through CI.

internal/controller/k0smotron.io/k0smotroncluster_monitoring_config.go
kahirokunn: Alright. Since I was at it, I did that too.

kahirokunn: CleanShot 2026-03-05 at 14 09 07@2x
I actually implemented handling for patches that would violate the rule in testing, and since it's showing as a lint error, this Go lint rule is definitely working. 👍

kahirokunn:
Hi @makhov @apedriza,

Could you take a look at merging this PR soon? Every time there's a force push or a big change on the `v1beta2-api-version` branch, this PR breaks and needs to be fixed again.

I've tested this locally with v1beta2 and it works fine. Happy to address any feedback!

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
apedriza, line 1: patches feature should be only supported for `v1beta2` no?

kahirokunn: Both have been tested and confirmed to work, so I think it's fine to provide them both.
I believe that would also simplify maintenance. What do you think?

.golangci.yml
```diff
@@ -8,10 +8,19 @@ run:
...
     - modernize
   settings:
+    forbidigo:
+      analyze-types: false
+      forbid:
+        - pattern: "^scope\\.client\\.(Create|Patch|Update)$"
+          msg: >-
+            Direct scope.client mutation call detected.
+            Use scope.reconcileResource() to ensure user-defined component patches are applied.
+            If this call is intentionally exempt, add //nolint:forbidigo with a justification comment.
```
apedriza, line 23: 👍

api/k0smotron.io/v1beta1/k0smotroncluster_types.go
apedriza: mm well not a strong opinion against any of the options but I think the less maintenance the better in general  and we are targeting this to v1beta2 branch so I thought the idea was to only include the feature for v1beta2 version. Including such a feature for a deprecated version does not feel very useful IMO. @makhov, any thoughts on this?

kahirokunn: That makes sense. agreed. I've removed the patches API from v1beta1 and restricted it to v1beta2 only. Thanks for the input!

kahirokunn:
@apedriza Hi! 👋 Thank you for the feedback, I've addressed all the review comments, including restricting the Patches API to v1beta2 only.

I've also tested everything locally (unit tests, build, and full e2e with make kind-capi-k0smotron) and confirmed it's all working as expected.

On my end, I went through another round of conflict resolution today due to the divergence with `v1beta2-api-version` branch 😢 
I'd really appreciate a re-review when you get a chance, I'm hoping we can get this across the finish
line and break out of the rebase cycle 🙏

Thanks again for your time and thorough reviews! 🙇

apedriza (approved):
LGTM, @kahirokunn thanks for the changes and patience 🙏

makhov (approved):

kahirokunn:
@makhov @apedriza Congratulations on the v2.0.0 release!
Just wanted to confirm. Will this commit be cherry-picked into some previous version? Thank you 🙏

PR #1343, opened 2026-01-30
Add `app.kubernetes.io/component` label

kahirokunn:
The label `app.kubernetes.io/component` is supported by a large number of tools by default.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -53,6 +53,10 @@ linters:
         - name: redefines-builtin-id
   exclusions:
     generated: strict
+    rules:
+      - linters: [revive]
+        text: "var-naming: avoid meaningless package names"
+        path: "internal/controller/util/"
     paths:
       # Ignore issues in generated files.
       - zz_generated.*\.go$


internal/controller/util/util.go
```diff
@@ -26,6 +26,7 @@ func LabelsForK0smotronCluster(kmc *km.Cluster) map[string]string {
 		labels[k] = v
 	}
 	labels["component"] = "cluster"
+	labels["app.kubernetes.io/component"] = "cluster"
```
apedriza: This func is used by multiple resources, which are not a "cluster" component, so it can be misleading to set this component here.

If the idea is to identify each component uniquely, we should ensure each of them has a unique component label. Now most of them are labeled with `app.kubernetes.io/component: cluster`. How can I then only patch one instead of all the components sharing the same label?

kahirokunn: I tried to add labels["app.kubernetes.io/component"] = "cluster" to all resources that have labels["component"] = "cluster", but does that mean there's actually a mistake in the existing implementation of labels["component"] = "cluster" itself?

apedriza: Found these two PRs related:
- https://github.com/k0sproject/k0smotron/pull/593
- https://github.com/k0sproject/k0smotron/pull/137

I would try to keep a solution for both cases to not break the fixes and follow a proper approach. 

for https://github.com/k0sproject/k0smotron/pull/593 we should keep a distinction between pods for controlplane and etcd, which could be done by the proposal of `app.kubernetes.io/component` label, for example, `control-plane` and `etcd` labels for each one. That way, we would be solving the same case that the PR wants to cover

On the other hand, https://github.com/k0sproject/k0smotron/pull/137 basically adds labels `app=k0smotron` and `cluster=<clustername>`, which I think are good in order to group components to the k0smotron cluster.

So probably some refactor for add base labels app=k0smotron and cluster=<clustername> for every component + the specific `app.kubernetes.io/component` label for each one covers all cases.

kahirokunn: Thanks for the detailed analysis and linking the related PRs!

I've tried to implement the approach you suggested.
Could you please check if this aligns with what you had in mind?

kahirokunn:
@apedriza Agreed, that’s more precise. I’ve updated the labels: the entrypoint ConfigMap now uses app.kubernetes.io/component=entrypoint and the cluster config ConfigMap uses app.kubernetes.io/component=cluster-config. Thanks for the suggestion.

apedriza (approved):
Thanks! 

Proper documentation should be added. It can be in another PR

kahirokunn:
Thank you😊
Yes, I'll create the documentation in a separate pull request. 😊
Let's merge🎉

apedriza:
@kahirokunn, it seems there are some issues in the k0smtoron upgrade test. Can you take a look? Retry does not works so it seems not be a flaky test. Changes in statefulset selector is forbidden so we need to think in another strategy

kahirokunn:
To achieve an upgrade rather than a recreation, it seems the only way is to separate the resource's metadata.labels and selector. 👀

apedriza:
I guess this is the easiest way we can follow at the moment, assuming a rollout for the controlplanes will be done once the template hash changes. 

I understand that by design, a replicaset selector is immutable but this makes me think what we can do in the future to deal with a needed change in the selector and how to switch to a new statefulset recreation, migrating the old state to the existing one. No idea atm, being honest.

kahirokunn:
Totally agree. For now, this approach looks good enough, so how about we go ahead and merge this and move forward? We can revisit the selector migration strategy in the future if needed. 👍

apedriza (approved):

PR #1346, opened 2026-02-04
fix: support all Ignition spec versions during provisioning

apedriza:
Ignition config was always generated for 3.4 version spec. Some variant versions for the butane config used to generate the ignition config may generate other ignition spec versions ([see map table](https://coreos.github.io/butane/specs/#butane-specifications-and-ignition-specifications)). 

With this change, we make use of [`ignition.config.merge`](https://coreos.github.io/ignition/configuration-v3_0/) feature to merge basic butane config generated for machine bootstrapping + the additional butane config the user can set and generate the proper ignition spec version. 

For additional context, merging butane configs [is not implemented](https://github.com/coreos/butane/issues/118) by butane package so we rely on the merge feature of the ignition config once basic and additional butane configs has been transpiled.

Also added an e2e example using Fedora CoreOS for machine nodes

### DIFF
--- a/internal/controller/bootstrap/controlplane_bootstrap_controller.go
+++ b/internal/controller/bootstrap/controlplane_bootstrap_controller.go
@@ -397,7 +397,7 @@ func (c *ControlPlaneController) generateBootstrapDataForController(ctx context.
 	files = append(files, resolvedFiles...)
 
 	if scope.currentKCPVersion.LessThan(minVersionForETCDMemberCRD) {
-		files = append(files, genShutdownServiceFiles(scope.Config.Spec.K0sInstallDir)...)
+		files = append(files, genShutdownServiceFiles(getShutdownFilesDir(scope.Config.Spec.WorkingDir), scope.Config.Spec.K0sInstallDir)...)
 	}
 
 	commands, commandsMap, err := c.genK0sCommands(scope, installCmd)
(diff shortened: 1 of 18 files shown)


internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -824,7 +824,7 @@ if [ $IS_LEAVING = "true" ]; then
 fi
 `, k0sPath),
 		}, {
-			Path:        "/k0s/k0sleave.service",
+			Path:        "/etc/k0s/k0sleave.service",
```
makhov: WDYT, does it make sense to use `spec.workingDir` here?

apedriza: absolutely, changed for using `spec.workingDir`, thanks!

internal/controller/bootstrap/controlplane_bootstrap_controller.go
```diff
@@ -967,6 +968,14 @@ func getFirstRunningMachineWithLatestVersion(machines collections.Machines) *clu
 	return res[0]
 }
 
+func getShutdownFilesDir(configuredWorkingDir string) string {
+	if configuredWorkingDir == "" {
+		// default to '/k0s' if no working dir is configured.
+		return "/k0s"
```
makhov: In the previous version it was changed it to `/etc/k0s` (and I believe for a reason). Is it ok to switch it back?

makhov (approved):
LGTM, just left a small question

internal/controller/bootstrap/controlplane_bootstrap_controller.go
apedriza: I changed it to `/etc/k0s` (assuming was OK no matters the distro) because for a  Fedora machine I cannot use `/k0s` but now that is configurable I use the `workingDir` field in the e2e for Fedora and keep the `/k0s` as the default for the shutdown files

makhov: Then maybe it makes sense to change it to `/etc/k0s` to make it work on Fedora by default?

apedriza: and keep the possibility to customize the shutdown files location with the `workingDir` field?

makhov: Yeah, of course. I mean, we can change the default location to cover more cases by default, but still make it configurable

apedriza: Alright, changed as discussed

PR #1349, opened 2026-02-09
fix: get workload cluster client from proper kubeconfig

### LINKED ISSUE #726: Use connection tunneling internally
Currently, we use "normal" kubeconfig in k0smotron for some features like dynamic config updating and providerID controller. 
But the workload cluster can have `externalAddress`, which is not accessible from the management one. In this case, we can use a kubeconfig for tunneling (if it's enabled)

https://docs.k0smotron.io/stable/capi-controlplane-bootstrap/?h=tunneling#client-connection-tunneling

apedriza:
fix #726


apedriza:
superseed by #1433

PR #1357, opened 2026-02-16
Bump K8s version references in docs and tests to v1.34.3

kahirokunn:
The k0s versions referenced in documentation and tests were significantly outdated (v1.27.x).
Updated them to v1.34.3 to better reflect current usage.


kahirokunn:
@apedriza Would you at least update the documentation and tests to use a version that has upstream support? 👍

PR #1394, opened 2026-03-11
feat: unify bootstrap resources into `K0sConfig`

apedriza:
### Changes

- Remove in v1beta2 usage of `K0sControllerConfig` and `K0sWorkerConfig` and unify them into `K0sConfig`. This makes things easier in terms of UX. Now, the user will use the same resource type to declare controller or worker configs, no extra fields needed for it.
- Rename field `k0s` where the user defines the k0s configuration file to `clusterConfig`, this way is more verbose the value of the field. For example, `K0sControllerConfig.spec.k0s` -> `K0sConfig.spec.clusterconfig`
- Removed conversion. Since there is no new version of the existing resources, but replacing them, the "conversion" for backward compatibility is done in the controller by mimicking the resources in the controller logic

### Fix

- Add conversion webhook in `K0sControlplaneTemplate` CRD


api/bootstrap/v1beta2/k0sconfig_template_types.go
```diff
@@ -0,0 +1,58 @@
...
+package v1beta2
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+func init() {
+	SchemeBuilder.Register(&K0sConfigTemplate{}, &K0sConfigTemplateList{})
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:deprecatedversion
```
makhov: I think it shouldn't be here, right?

makhov:
As I said before, it's not that easy. 

`clusterConfig` and `tunneling` fields are only meaningful for controller nodes. Also `args` means k0s controller flags in one context and k0s worker flags in another. 

And we already have K0sConfigSpec which supposed to keep common fields from both objects, we just don't use it in K0sWorkerConfig for some reason. 

Also, I'm not sure that creating v1beta1 shadow objects is a good idea in general. Sounds fragile.

PR #1408, opened 2026-03-23
Add clustercache for workload clusters connections

apedriza:


makhov:
there are still a couple of `remote.NewClusterClient` usages left in k0smotron ClusterReconciler.

apedriza:
> there are still a couple of `remote.NewClusterClient` usages left in k0smotron ClusterReconciler.

👍  will check it out

internal/controller/util/wc_client.go
```diff
@@ -17,33 +17,41 @@ package util
...
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
 	"slices"
 	"time"
 
+	"github.com/cloudflare/cfssl/log"
```
makhov: It's probably the wrong logger

makhov (commented):
`internal/controller/k0smotron.io/k0smotroncluster_configmap.go:175` still uses `remote.NewClusterClient`

internal/controller/util/wc_client.go
apedriza: oh yes

apedriza:
> `internal/controller/k0smotron.io/k0smotroncluster_configmap.go:175` still uses `remote.NewClusterClient`

we cannot use clustercache pkg for k0smotron standalone as it only works for CAPI clusters

makhov (approved):

PR #1417, opened 2026-04-09
Add status conditions for JoinTokenRequest resource

apedriza:
Deprecate `status.reconciliationStatus` and use conditions instead to report reconcile status. 

JointTokenRequest controller mainly generates a secret with the join token so added a condition `JoinTokenSecretCreated` describing if it is created or not.


internal/controller/k0smotron.io/jointokenrequest_controller.go
```diff
@@ -72,34 +75,44 @@ func (r *JoinTokenRequestReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-		r.updateStatus(ctx, jtr, "Failed update JoinTokenRequest with cluster UID label")
-		return ctrl.Result{Requeue: true, RequeueAfter: time.Minute}, err
+		reconcileFailureReason = "Failed to add finalizer to cluster for JoinTokenRequest resource removal"
 	}
 
 	logger.Info("Reconciling")
 	pod, err := util.FindStatefulSetPod(ctx, r.ClientSet, km.GetStatefulSetName(jtr.Spec.ClusterRef.Name), jtr.Spec.ClusterRef.Namespace)
 	if err != nil {
-		r.updateStatus(ctx, jtr, "Failed finding pods in statefulset")
+		reconcileFailureReason = "Failed finding pods in statefulset"
```
makhov: If I recall correctly, Reason should regex, basically OneCamelCaseWord, spaces wouldn't match it

apedriza: right, fixed!

makhov (approved):

PR #1418, opened 2026-04-09
Align RemoteMachine and RemoteCluster status for CAPI v1beta2

apedriza:
Machine and Cluster controllers expect conditions in RemoteMachine and RemoteCluster to surface state into Machine and Cluster resources


internal/controller/infrastructure/remote_machine_controller.go
```diff
@@ -213,6 +152,17 @@ func (r *RemoteMachineController) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, fmt.Errorf("failed to parse bootstrap data: %w", err)
 	}
 
+	// Reserve a machine from the pool if pool is specified in the RemoteMachine spec, and populate the RemoteMachine spec with the
+	// details of the reserved machine needed for provisioning. This needs to be done before fetching the ssh key because the ssh
+	// key reference is part of the pool machine spec.
+	if rm.Spec.Pool != "" {
```
makhov: Does this mean k0smotron will try to reserve pooled machine even it has one (e.g. on deletion)

apedriza: added a condition in the `if` to check if the machine is being deleted

In a follow up I would like to take a deeper look whether a bit of refactoring makes sense in this controller

makhov (approved):

PR #1420, opened 2026-04-09
Add --watch-filter and --namespace flags to support multiple k0smotron instances

### LINKED ISSUE #1419: Multiple k0smotron instances in the same cluster cause continuous reconcile loops
When running more than one k0smotron instance in the same cluster, all instances watch resources cluster-wide with no isolation between them. Each instance's reconcile mutations trigger watch events on the other instances, causing a continuous reconcile storm that wastes CPU and hammers the API server. There is currently no way to scope an instance to a subset of resources.

This also causes CPU spikes and makes management cluster to suffer.

makhov:
Fixes #1419 

`--watch-filter` flag scopes the informer's LIST/WATCH calls to only resources matching the given `cluster.x-k8s.io/watch-filter=<filter>` label selector. Example usage: label managed resources per instance (e.g. `cluster.x-k8s.io/watch-filter=instance-foo`) and start each controller with `--watch-filter=instance-foo`.

`--namespace` specifies that the controller watches.


apedriza:
Is this something that should be explained in the installation docs?

apedriza:
I see [this in CAPI docs](https://cluster-api.sigs.k8s.io/developer/core/support-multiple-instances#:~:text=Providers%20MUST%20support%20the%20%2D%2Dnamespace,filter%20flag%20in%20their%20controllers.) about expected flags in the providers running multiple instances for same provider. Seems to be required

makhov:
Thanks! Fixed

apedriza (approved):

PR #1422, opened 2026-04-09
Fix ssh host key mismatch on VM recreation

### LINKED ISSUE #1421: ssh host key mismatch on VM recreation with same IP
rig persists known hosts keys by default which causes reconcile error when VM was recreated with the same IP:
```
failureMessage: 'failed to connect to host: client connect: retry: abort condition
    reached after 1 attempts: operation can not be completed: ssh: handshake failed:
    host key mismatch: knownhosts: key mismatch'
```

makhov:
Fixes #1421


apedriza (commented):
we could remove `SSH_KNOWN_HOSTS` env in deployment then, no?

apedriza (approved):

PR #1428, opened 2026-04-16
fix postcommand forward when provisioning in controlplane nodes

apedriza:


makhov (approved):
We probably need to create a release-1.10 branch, backport and release this fix...

PR #1436, opened 2026-04-21
feat: Don't move onto draining next node while current drain is in progress

wikoion:


apedriza:
@wikoion did you experience more than one drain at the same time? Can you describe your case? Control plane controller covers one-by-one deletions

wikoion:
> @wikoion did you experience more than one drain at the same time? Can you describe your case? Control plane controller covers one-by-one deletions

No, sorry. I thought this might be the cause of an issue we're having with cp node rollouts, but turned out to be something else.

PR #1446, opened 2026-05-04
Add convert tool

apedriza:
Best-effort tool for convert manifests with k0smotron resources from v1beta1 to v1beta2 by using implemented `Convetible` inferface. Removed after v1beta1 support is gone

### DIFF
--- a/docs/update/migrate-v1beta2-api.md
+++ b/docs/update/migrate-v1beta2-api.md
@@ -2,6 +2,59 @@
 
 This guide explains changes introduced in new `v1beta2` and how to migrate manifests from current `v1beta1` to the new k0smotron API. For more details about each Api Group, please check current custom resource references.
 
+## Migration tool
+
+k0smotron ships an experimental CLI tool that automates the manifest conversion described in this guide.
+
+### Download pre-built binary
+
+Pre-built binaries are published with every release for the following platforms:
+
+| Platform | Binary |
+|---|---|
+| Linux x86-64 | `convert-v1beta1-to-v1beta2-linux-amd64` |
+| Linux ARM 64-bit | `convert-v1beta1-to-v1beta2-linux-arm64` |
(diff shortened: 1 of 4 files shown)


docs/update/migrate-v1beta2-api.md
```diff
@@ -2,6 +2,59 @@
...
+
+k0smotron ships an experimental CLI tool that automates the manifest conversion described in this guide.
+
+### Download pre-built binary
+
+Pre-built binaries are published with every release for the following platforms:
+
+| Platform | Binary |
+|---|---|
+| Linux x86-64 | `convert_linux_amd64` |
```
makhov: Nit: we use dashes everywhere, I'd prefer to keep it consistent. 
And does it make sense to add `v1beta1-to-v1beta2` to name or something like this?

hack/convert/v1beta2.go
```diff
@@ -0,0 +1,204 @@
...
+		infrav1beta1.AddToScheme,
+		infrav1beta2.AddToScheme,
+	} {
+		if err := add(scheme); err != nil {
+			panic(err)
+		}
+	}
+}
+
+func main() {
```
makhov, line 64: Not really sure how smart this command should be, I think stdin/file to stdout is good enough. The only question, does it work with complex multi-object yaml files?

makhov (commented):
Great idea!

docs/update/migrate-v1beta2-api.md
apedriza: Changed!

hack/convert/v1beta2.go
apedriza: yes, I think current stdin or file as input is enough for now. We can iterate over this if something else is required later. I've tested with full cluster deployment manifest with multiple resources and seems to work well

makhov (approved):

PR #1450, opened 2026-05-08
Kubeconfig reconcile fix

makhov:
Currently we recreate kubeconfig on every reconcile request

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_kubeconfig.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_kubeconfig.go
@@ -18,10 +18,7 @@ package k0smotronio
 
 import (
 	"context"
-	"crypto/x509"
-	"encoding/pem"
 	"fmt"
-	"time"
 
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


internal/controller/k0smotron.io/k0smotroncluster_kubeconfig.go
```diff
@@ -143,3 +164,49 @@ func rewriteKubeconfigValues(kubeconfigYAML string, kmc *km.Cluster) (string, er
 
 	return string(configBytes), nil
 }
+
+// kubeconfigStillValid reports whether the secret already holds a usable admin kubeconfig:
+// the client certificate is well above the renewal threshold and the server URL matches
+// the desired one when an Ingress is configured. Returning true lets the caller skip the
+// pod exec and SSA write, which is what prevents the reconcile-loop storm.
+func kubeconfigStillValid(secret *v1.Secret, kmc *km.Cluster) bool {
```
apedriza: I wonder whether we can use the logic we have for vms cps https://github.com/k0sproject/k0smotron/blob/13c19f25b6c8d5adb3d3b9a349177b3a54eb4199/internal/controller/controlplane/k0s_controlplane_controller.go#L243-L248 when checking cert rotation, and the same threshold var

makhov: Thanks! Switched to the kubeconfig.NeedsClientCertRotation helper

apedriza (approved):
lgtm, the only thing is that we are using different renewal thresholds for hcp and vms. 6 months vs 30 days. Not sure if those should be same or not

PR #1451, opened 2026-05-08
Cache cluster client

makhov:
Cache cluster client to avoid creating a new one on each reconcile request. Saves memory and cpu during high reconcile loads


apedriza (commented):
did you check https://github.com/k0sproject/k0smotron/pull/1408/? I think using the CAPI controller is a good way to externalize cache management. Also, I'm not sure how much overlap there is between both PRs

makhov:
Sorry, totally forgot about that PR.

PR #1452, opened 2026-05-08
Refactor reconcile machine logic

apedriza:
Refactor of machines reconcile logic related to scale up/down process. Also, added more integration tests covering more scaling scenarios

### DIFF
--- a/internal/controller/controlplane/helper.go
+++ b/internal/controller/controlplane/helper.go
@@ -38,7 +38,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/ptr"
 	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
-	"sigs.k8s.io/cluster-api/controllers/external"
 	"sigs.k8s.io/cluster-api/util/collections"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -64,27 +63,6 @@ const (
 var (
 	errMachineWithoutK0sConfigAnnotation = fmt.Errorf("k0s config annotation not found on machine")
 )
-
-func (c *K0sController) createMachine(ctx context.Context, name string, cluster *clusterv1.Cluster, kcp *cpv1beta2.K0sControlPlane, infraRef clusterv1.ContractVersionedObjectReference, failureDomain string) (*clusterv1.Machine, error) {
-	machine, err := c.generateMachine(ctx, name, cluster, kcp, infraRef, failureDomain)
-	if err != nil {
-		return nil, fmt.Errorf("error generating machine: %w", err)
-	}
-	_ = ctrl.SetControllerReference(kcp, machine, c.Client.Scheme())
-
-	err = c.Client.Patch(ctx, machine, client.Apply, &client.PatchOptions{
-		FieldManager: "k0smotron",
-	})
-	if err != nil {
-		return machine, err
-	}
-
-	// Remove the annotation tracking that a remediation is in progress.
-	// A remediation is completed when the replacement machine has been created above.
-	delete(kcp.Annotations, cpv1beta2.RemediationInProgressAnnotation)
-
-	return machine, nil
-}
 
 func (c *K0sController) deleteMachine(ctx context.Context, name string, kcp *cpv1beta2.K0sControlPlane) error {
 	machine := &clusterv1.Machine{
@@ -156,6 +134,7 @@ func (c *K0sController) generateMachine(_ context.Context, name string, cluster 
 			InfrastructureRef: infraRef,
 		},
 	}
+	_ = ctrl.SetControllerReference(kcp, machine, c.Client.Scheme())
 
 	return machine, nil
 }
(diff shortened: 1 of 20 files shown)


internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -619,12 +368,10 @@ func (c *K0sController) createBootstrapConfig(ctx context.Context, name string,
...
 			Annotations: kcp.Spec.MachineTemplate.ObjectMeta.Annotations,
 			OwnerReferences: []metav1.OwnerReference{{
-				APIVersion:         machine.APIVersion,
-				Kind:               machine.Kind,
-				Name:               machine.GetName(),
-				UID:                machine.GetUID(),
-				BlockOwnerDeletion: ptr.To(true),
-				Controller:         ptr.To(true),
+				APIVersion: cpv1beta2.GroupVersion.String(),
+				Kind:       "K0sControlPlane",
```
makhov, line 368: Why do we change the owner here? CAPI convention is bootstrap config owned by Machine, isn't it?

internal/controller/controlplane/helper.go
```diff
@@ -316,40 +295,18 @@ func (c *K0sController) generateMachineFromTemplate(ctx context.Context, name st
...
 	machineK0sConfig, err := getMachineK0sConfig(machine)
 	if err != nil {
 		// TODO: Remove this fallback logic in a future release.
 		if errors.Is(err, errMachineWithoutK0sConfigAnnotation) {
-			return deprecatedIsK0sConfigChanged(&bootstrapConfig, kcp, machine)
+			return !deprecatedIsK0sConfigChanged(bootstrapConfig, kcp, machine)
 		}
 
-		return false
+		return true
```
makhov: Hm, it returns true on error?

internal/controller/controlplane/scale.go
```diff
@@ -0,0 +1,371 @@
...
+	logger := log.FromContext(ctx)
+
+	err := c.reconcileUnhealthyMachines(ctx, scope)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	// Before machines scaling, we need to make sure that all deleted machines are fully deleted and their k0s resources cleaned
+	// up. This is needed to ensure a graceful deletion of the control plane machines when the user deletes them directly, without
+	// relying on the k0smotron controller.
+	err = c.recocileDeletedMachines(ctx, scope)
```
makhov: A typo

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -1011,6 +735,83 @@ func (c *K0sController) createFRPToken(ctx context.Context, cluster *clusterv1.C
...
+	}
+
+	uBootstrapConfig, err := external.GetObjectFromContractVersionedRef(ctx, c.Client, m.Spec.Bootstrap.ConfigRef, m.Namespace)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return machineState{}, fmt.Errorf("failed to retrieve controller config for machine object %s: %w", m.Name, err)
+		}
+	}
+	bootstrapConfig := &bootstrapv2.K0sControllerConfig{}
+	err = runtime.DefaultUnstructuredConverter.FromUnstructured(uBootstrapConfig.Object, bootstrapConfig)
```
makhov: This probably will panic on NotFound error.

internal/controller/controlplane/scale.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+// isNeededScaleUp checks if it's needed to scale up the control plane based on the current state of the cluster and the desired number of replicas.
+func isNeededScaleUp(scope *controlplane) bool {
+	if isNeededApplyDeleteFirstStrategy(scope) {
+		return false
+	}
+
+	potentialMachinesCount := scope.activeMachines.Len() + 1
+	maximunAllowedMachines := int(scope.kcp.Spec.Replicas) + 1
```
makhov: A typo

internal/controller/controlplane/scale.go
```diff
@@ -0,0 +1,371 @@
...
+	// If we already have the maximum allowed machines, we cannot scale up anymore until some machines are deleted.
+	if potentialMachinesCount > maximunAllowedMachines {
+		return false
+	}
+
+	// Scale up control plane machines if there are less up to date machines than the desired replicas.
+	return scope.upToDateMachines.Len() < int(scope.kcp.Spec.Replicas)
+}
+
+// isNeededDeleteBeforeScaleUp checks if it's needed to delete a machine before scaling up the control plane, based on the UpdateStrategy and the current state of the cluster.
```
makhov: isNeededApplyDeleteFirstStrategy ...

internal/controller/controlplane/remediation.go
```diff
@@ -85,10 +79,10 @@ func (c *K0sController) reconcileUnhealthyMachines(ctx context.Context, cluster
 	}()
 	// Ensure that the cluster remains available during and after the remediation process. The remediation must not
 	// compromise the cluster's ability to serve workloads or cause disruption to the control plane's functionality.
-	if ptr.Deref(kcp.Status.Initialization.ControlPlaneInitialized, false) {
+	if ptr.Deref(scope.kcp.Status.Initialization.ControlPlaneInitialized, false) {
 		// The cluster MUST have more than one replica, because this is the smallest cluster size that allows any etcd failure tolerance.
-		if !(machines.Len() > 1) {
-			log.Info("A control plane machine needs remediation, but the number of current replicas is less or equal to 1. Skipping remediation", "replicas", machines.Len())
+		if !(scope.activeMachines.Len() > 1) {
```
makhov: Maybe simplify it to `if scope.activeMachines.Len() <= 1` ?

internal/controller/controlplane/k0s_controlplane_controller.go
apedriza: k0scontrolplane is a temporal owner in the bootstrap config to keep cascade deletion now that the machine is created after it. The machine controller then [updates the owner ref in the bootstrap config](https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/controllers/machine/machine_controller_phases.go#L122-L127) when the machine is created. Same approach used by kubeadmcontrolplane.

Sometimes there are cases where the machine reconciliation does not found the expected bootstrao config because it is not created yet (just a timing thing but can be a bit confusing) so I inverted the creation order. This way I think makes more sense where bootstrap config does not wait for the machine, it just creates its stuff

internal/controller/controlplane/helper.go
apedriza: fixed!

internal/controller/controlplane/scale.go
apedriza: fixed!

internal/controller/controlplane/k0s_controlplane_controller.go
apedriza: fixed! It follows our current approach, where if inframachine or bootrapconfig are not found, the reconcile logic assumes machines is up to date.

internal/controller/controlplane/scale.go
apedriza: fixed

internal/controller/controlplane/remediation.go
apedriza: updated

internal/controller/controlplane/scale.go
apedriza: fixed

internal/controller/controlplane/k0s_controlplane_controller.go
makhov: I think the same for `uInfraMachine` - it might panic inside `isInfraMachineUpToDate` if nil

apedriza: fixed those cases too! thanks

apedriza:
@makhov can you have a look again please?

makhov (approved):

PR #1460, opened 2026-05-29
JoinTokenRequest in the same k0smotron Cluster namespace

### LINKED ISSUE #1456: JoinTokenRequest in the same namespace as k0smotron cluster
Currently is it allowed deploy a `JoinTokenRequest` in a different namesapce than the one used by the k0smtoron cluster. This makes impossible to use owner references for cascade deletion and instead, use a "manual deletion" of the JTR associated to a cluster when this is gone. We could follow same aporach than CAPI follows which is put everything related to the parent resource (in this case k0smotoron cluster) over the same namespace.

apedriza:
fix #1456 

Deprecate support of cross-namespace references for k0smtoron cluster in JoinTokenRequest.

For v1beta1 cases, behavior keeps the same where manual deletion of JoinTokenRequest resources is done by using the `k0smotron.io/cluster-uid`. 

For v1beta2, owner references are used for gc so no need to include `k0smotron.io/cluster-uid` label.


internal/controller/k0smotron.io/jointokenrequest_controller.go
```diff
@@ -236,6 +277,27 @@ func (r *JoinTokenRequestReconciler) updateStatus(jtr *km.JoinTokenRequest, fail
...
+			return false
+		}
+	}
+
+	refs = append(refs, metav1.OwnerReference{
+		APIVersion: km.GroupVersion.String(),
+		Kind:       kmc.GroupVersionKind().Kind,
+		Name:       kmc.GetName(),
+		UID:        kmc.GetUID(),
+		Controller: ptr.To(true),
```
makhov: Linter with Go 1.26 will show warning :(

makhov (approved):

internal/controller/k0smotron.io/jointokenrequest_controller.go
apedriza: fixed!

makhov:
LGTM!

PR #1463, opened 2026-06-01
Switch to rig v0.21.8

makhov:


makhov:
The linter fails with a bunch of errors unrelated to the PR. I'll fix them after review to avoid the noise

apedriza (approved):

PR #1469, opened 2026-06-04
Change k0smotron version including v1beta2 to v2.0.0

apedriza:
As agreed, next k0smotron version for v1beta2 will be v2.0.0


makhov:
Should we bump metadata.yaml as well in this PR?

makhov (approved):

PR #1471, opened 2026-06-05
Adapt metadata v1beta2 for aws e2e

apedriza:
This PR fixes aws e2e configuration, making it work with the new v1beta2 version and disables Windows test temporarily, as further changes need to be done to make it pass

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -116,10 +116,11 @@ jobs:
     runs-on: oracle-vm-16cpu-64gb-x86-64
     strategy:
       fail-fast: false
+      max-parallel: 1
       matrix:
         e2e-suite:
           - TestIgnitionProvisioning
-          - TestWindowsProvisioning
+          #- TestWindowsProvisioning
     env:
       SSH_KEY_NAME: ${{ secrets.SSH_KEY_NAME }}
       SSH_PUBLIC_KEY: ${{ secrets.SSH_PUBLIC_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-nightly.yml
```diff
@@ -116,10 +116,11 @@ jobs:
     runs-on: oracle-vm-16cpu-64gb-x86-64
     strategy:
       fail-fast: false
+      max-parallel: 1
```
apedriza, line 119: more than one test in parallel will hit limits about resources in aws

.github/workflows/e2e-nightly.yml
```diff
@@ -116,10 +116,11 @@ jobs:
...
       matrix:
         e2e-suite:
           - TestIgnitionProvisioning
-          - TestWindowsProvisioning
+          #- TestWindowsProvisioning
```
makhov, line 123: Is it some debug leftover?

apedriza: It was to unblock CI a bit and disable windows test temporarily as I couldnt run it. Seems manifest needs some changes like the api version for the worker config to v1beta2, new ssh key pair for the region and a another ami. So it was intentional. I prefer deal with only windows stuff isolated to not mix e2e config

makhov: Ok, got it. I think it should not delay the new release. We can fix it later

makhov (approved):

PR #1475, opened 2026-06-11
Helm Chart

makhov:
CRDs are installed as a subchart. The chart supports only installing into `k0smotron-system` namespace. To install it to other ns it needs to be regenerated `make helm-chart CRD_NAMESPACE=<custom_namespace>

For now, the chart versioning follows the k0smotron versioning because the chart is autogenerated


.github/workflows/release.yml
```diff
@@ -221,3 +221,61 @@ jobs:
...
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | helm registry login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Log in to Quay registry
+        run: echo "${{ secrets.QUAY_PASSWORD }}" | helm registry login quay.io -u ${{ secrets.QUAY_USER }} --password-stdin
+
+      - name: Push chart to ghcr.io and quay.io
+        run: |
+          helm push "${CHART_TGZ}" oci://ghcr.io/k0sproject/charts
+          # We push chart only to GitHub for now
+          # helm push "${CHART_TGZ}" oci://quay.io/k0sproject/charts
```
apedriza: why not to push it there too?

PROJECT
apedriza, line 1: is this dile needed?

dist/chart/charts/k0smotron-crds/Chart.yaml
apedriza, line 1: could this be splited in 4 charts? standalone, only capi controlplane prov, only capi bootstrap prov, only capi infra prov. That way we allow user decide which crds to install by using `--enable-controller` manager flag

apedriza (commented):
When executin `make helm-chart` the chart dir gets modified so seems its content is not up-to-date. Maybe we should include `make helm-chart` target into the `make generate-all`

.github/workflows/release.yml
makhov: I just wanted to test it first, since quay registry requires some setup

PROJECT
makhov: It's actually a standard file for kubebuilder's controllers, used by helm plugin, also allows us to use `kubebuilder create api / create webhook` instead of doing all the work manually

dist/chart/charts/k0smotron-crds/Chart.yaml
makhov: I added `--enable-controller` support to the values

.github/workflows/release.yml
makhov: it's probably should `quay.io/k0sproject/k0smotron-chart`

PR #1477, opened 2026-06-11
Unexpected rollout when upgrading to v2.0.0

apedriza:


internal/controller/controlplane/helper.go
```diff
@@ -351,15 +354,35 @@ func hasControllerConfigChanged(bootstrapConfigs map[string]bootstrapv2.K0sContr
 
 		return false
 	}
+	withV1Beta2Defaults(machineK0sConfig)
```
makhov, line 357: I think the approach in the #1479 a bit cleaner since it reuses the existing conversion funcs. I suspect, the defaults approach misses the field renaming, so we may accidentally skip or swap the things like ignition to cloud-init

apedriza: yes, I agree. I was about close this one in favor on yours. But Im testing that PR and Im also falling in a rollout, let me double check it

PR #1479, opened 2026-06-12
Fix unexpected machine rollout due to wrong version unmarshalling

makhov:


apedriza:
I think `downloadURL` also needs to be converted to the default at conversion as in v1beta2 the field has a kubebuilder fault too

makhov:
I copied your approach with `withV1Beta2Defaults`. I think it should be better now

apedriza (approved):
LGTM

PR #1487, opened 2026-06-16
Fix: do not run clustercache controller in standalone mode

### LINKED ISSUE #1486: standalone installation does not work with v2.0.0
### What happened?

The [instructions for installing](https://docs.k0smotron.io/v2.0.0/install/#k0smotron-standalone) don't work for the [2.0.0 standalone installation resources](https://docs.k0smotron.io/v2.0.0/install-standalone.yaml).

The k0smotron-controller-manager deployment gets stuck crashlooping. Logs indicate it's failing to read clusters.cluster.x-k8s.io, where that CRD is not included in the standalone resources and the k0smotron-controller-manager service account does not have permission to read the resource.

### Involved Areas

_No response_

Standalone installation should work without relying on clusters.cluster.x-k8s.io

### Steps to reproduce

`kubectl apply --server-side=true -f https://docs.k0smotron.io/v2.0.0/install-standalone.yaml`

### k0smotron version

2.0.0

### k0s version

any

Workaround kustomization.yaml example:
```yaml
resources:
- https://docs.k0smotron.io/v2.0.0/install-standalone.yaml # does not include CAPI CRDs.
# HACK: standalone k0smotron is still reading CAPI CRDs, and fails without
- https://raw.githubusercontent.com/kubernetes-sigs/cluster-api/refs/tags/v1.13.2/config/crd/bases/cluster.x-k8s.io_clusters.yaml
patches:
# HACK: also add permissions for manager to read the extra added CRD
- target:
    kind: ClusterRole
    name: k0smotron-manager-role
  patch: |-
    - op: add
      path: /rules/-
      value:
        apiGroups:
        - cluster.x-k8s.io
        resources:
        - clusters
        - clusters/status
        verbs:
        - get
        - list
        - watch
```

apedriza:
fix #1486

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -267,23 +267,26 @@ func main() {
 
 	ctx := ctrl.SetupSignalHandler()
 
-	clusterCache, err := clustercache.SetupWithManager(ctx, mgr, clustercache.Options{
-		SecretClient: secretCachingClient,
-		Cache:        clustercache.CacheOptions{},
-		Client: clustercache.ClientOptions{
-			UserAgent: "k0smotron",
-			Cache: clustercache.ClientCacheOptions{
-				DisableFor: []client.Object{
-					// Don't cache ConfigMaps & Secrets.
-					&corev1.ConfigMap{},
-					&corev1.Secret{},
+	var clusterCache clustercache.ClusterCache
+	if runCAPIControllers {
+		clusterCache, err = clustercache.SetupWithManager(ctx, mgr, clustercache.Options{
+			SecretClient: secretCachingClient,
+			Cache:        clustercache.CacheOptions{},
+			Client: clustercache.ClientOptions{
+				UserAgent: "k0smotron",
+				Cache: clustercache.ClientCacheOptions{
+					DisableFor: []client.Object{
+						// Don't cache ConfigMaps & Secrets.
+						&corev1.ConfigMap{},
+						&corev1.Secret{},
+					},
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -267,23 +267,26 @@ func main() {
...
-		SecretClient: secretCachingClient,
-		Cache:        clustercache.CacheOptions{},
-		Client: clustercache.ClientOptions{
-			UserAgent: "k0smotron",
-			Cache: clustercache.ClientCacheOptions{
-				DisableFor: []client.Object{
-					// Don't cache ConfigMaps & Secrets.
-					&corev1.ConfigMap{},
-					&corev1.Secret{},
+	var clusterCache clustercache.ClusterCache
```
twz123, line 270: Is it okay to pass a nil cluster cache to K0s(motron)Controller around lines 325 and following? That branch isn't guarded by `runCAPIControllers`, just by `isControllerEnabled(controlPlaneController)`.

twz123 (approved):
Left one question that wasn't directly clear to me from having a quick look at the code.

cmd/main.go
apedriza: Every controller (only CAPI integration related) that can use cluster cache is under the `runCAPIControllers` equals true and in that case clustercache cannot be nil

twz123: Right! I overlooked the nested if statement :grimacing:

PR #1488, opened 2026-06-16
Override konnectivity agent image if possible

### LINKED ISSUE #1484: Ingress setup ignores `images.repository` for the konnectivity agent image
### What happened?

When deploying clusters with ingress enabled, the konnectivity agent image is hardcoded to `quay.io/k0sproject/apiserver-network-proxy-agent:v0.33.0` (see: https://github.com/k0sproject/k0smotron/blob/v2.0.0/internal/controller/k0smotron.io/k0smotroncluster_ingress.go#L285). As a result, the image is not rewritten to use the configured global registry (`images.repository`), causing image pulls to fail in air-gapped environments where access to quay.io is unavailable.

### Involved Areas

Cluster API integration

The Konnectivity agent image should honor the global `images.repository` setting when ingress is enabled, or provide a dedicated configuration option to override the image registry.

### Steps to reproduce

1. Deploy a CAPI cluster using K0smotronControlPlane in an air-gapped environment with a global image registry configured (with `images.repository`, `k0s` image override, `etcd` image override, etc.) and ingress enabled. One way to reproduce this is by using k0rdent and one of the default cluster templates that enables ingress and configures a global registry, such as: https://github.com/k0rdent/kcm/blob/main/templates/cluster/openstack-hosted-cp/templates/k0smotroncontrolplane.yaml
2. After the cluster is provisioned, observe that the konnectivity agent image is still pulled from `quay.io/k0sproject/apiserver-network-proxy-agent:v0.33.0` instead of being rewritten to use the configured `images.repository` mirror.

### k0smotron version

v2.0.0

### k0s version

v1.35.4+k0s.0

Related to: https://github.com/k0rdent/k0rdent-enterprise/issues/360

apedriza:
fix #1484

### DIFF
--- a/internal/controller/k0smotron.io/k0smotroncluster_ingress.go
+++ b/internal/controller/k0smotron.io/k0smotroncluster_ingress.go
@@ -19,10 +19,12 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	km "github.com/k0sproject/k0smotron/v2/api/k0smotron.io/v1beta2"
 	kcontrollerutil "github.com/k0sproject/k0smotron/v2/internal/controller/util"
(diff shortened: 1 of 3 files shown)


internal/controller/k0smotron.io/k0smotroncluster_ingress.go
```diff
@@ -92,6 +93,21 @@ func (scope *kmcScope) reconcileIngress(ctx context.Context, kmc *km.Cluster) er
 	return nil
 }
 
+func (scope *kmcScope) getKonnectivityAgentImageWithOverrides(kmc *km.Cluster) string {
```
twz123: Oof, this is reproducing a very brittle part of the k0s config logic. Would be good if this wouldn't be necessary here. You probably want to have a look at spec.images.repository, as well, and implement the repository override. Moreover, the default pull policy doesn't seem to be used at all in the konnectivity manifests.

twz123 (approved):
Maybe add the repository override?

apedriza:
> Maybe add the repository override?

I've replicated k0s logic for repo override and also take into account if `default_pull_policy` is set. Thanks!

PR #1496, opened 2026-06-23
Fix bootstrap command to run shell logic with an interpreter

### LINKED ISSUE #1494: RemoteMachine: `k0s install worker` reported as exit status 1 in per-command mode (succeeds on host); Provision not idempotent
### Summary

A `RemoteMachine` worker never provisions in the default per-command provisioning path: the controller reports `k0s install worker` as `Process exited with status 1`, even though the command **actually succeeds on the host** (it writes the systemd unit). Because the controller then treats provisioning as failed and `Provision` is not idempotent, every retry aborts with `Init already exists` and the machine can never recover. Setting `commandsAsScript: true` works around both problems.

### Environment

- k0smotron **v2.0.1** (bootstrap + control-plane + infrastructure)
- Cluster API v1.13.2
- k0s **v1.34.8+k0s.0**, pre-installed on the host (`K0sWorkerConfig.preInstalledK0s: true`)
- Worker: bare-metal Ubuntu 22.04.5, joined over SSH via `RemoteMachine` (`useSudo: true`, `user: ubuntu`)
- Hosted control plane via `K0smotronControlPlane` (single replica)

### What happens

Infrastructure controller log:

```
running command  ... "command": "/usr/local/bin/k0s install worker --token-file /etc/k0s.token --labels=k0smotron.io/machine-name=gpu-a-sd-188 --kubelet-extra-args=\"--hostname-override=gpu-a-sd-188\""
DEBUG [ssh] host:22: executing `sudo -- /usr/local/bin/k0s install worker --token-file /etc/k0s.token --labels=k0smotron.io/machine-name=gpu-a-sd-188 --kubelet-extra-args="--hostname-override=gpu-a-sd-188"`
failed to run command ... "output": "", "error": "command failed: client exec: ssh session wait: Process exited with status 1"
```

`RemoteMachine`: `BootstrapExecSucceeded=False` (InternalError), `failureReason=ProvisionFailed`.

But on the host the install **succeeded** — `/etc/systemd/system/k0sworker.service` is written on each attempt. The byte-identical command from a normal SSH session returns rc=0:

```
$ ssh ubuntu@host 'sudo -- /usr/local/bin/k0s install worker --token-file /etc/k0s.token \
    --labels=k0smotron.io/machine-name=gpu-a-sd-188 \
    --kubelet-extra-args="--hostname-override=gpu-a-sd-188"'; echo rc=$?
rc=0          # unit created
```

So in the default per-command path (`SSHProvisioner.Provision` → `connection.ExecOutput(cmd, exec.Sudo(...))` in `internal/controller/infrastructure/ssh_provisioner.go`), `k0s install worker` is reported as exit status 1 even though it writes the unit and would otherwise succeed. (Suspect: the per-command `ExecOutput` + `exec.Sudo` path mishandles either the exit status or the `--kubelet-extra-args="..."` quoting; the single-script path below does not.)

### Secondary problem — Provision is not idempotent

Because the unit was written but the attempt is treated as failed, every retry re-runs `k0s install worker` and aborts (to **stderr**, so the controller's `output` is empty):

```
Error: failed to install worker service: Init already exists: /etc/systemd/system/k0sworker.service
```

Once any attempt has run, the `RemoteMachine` can only recover if the host is reset to a pristine state (no `k0sworker.service`) **and** the `Machine` is recreated. A plain retry never succeeds.

### Workaround

`RemoteMachine.spec.commandsAsScript: true` (+ `workingDir: /root` so the generated `k0s_install.sh` path is absolute under sudo) makes the controller write the runcmd to a single `#!/bin/sh -e` script and run it once. With that, `k0s install worker` reports its true rc=0 and the worker joins.

### Expected

The per-command path should report the true exit code of `k0s install worker` (rc=0 on success), and/or `Provision` should be idempotent (tolerate an already-installed `k0sworker.service` on retry, e.g. skip/`k0s install` only when not present).

### Reproduce

1. Pristine host (no `k0sworker.service`), k0s binary pre-installed.
2. `RemoteMachine` with `useSudo: true`, default `commandsAsScript` (false); `K0sWorkerConfig` with `preInstalledK0s: true` and `--kubelet-extra-args="--hostname-override=<name>"`.
3. Observe `BootstrapExecSucceeded=False` / `Process exited with status 1` while `/etc/systemd/system/k0sworker.service` is in fact created on the host.
4. Set `commandsAsScript: true`, reset the host pristine, recreate the `Machine` → it now provisions successfully.

apedriza:
fix #1494

### DIFF
--- a/internal/controller/infrastructure/ssh_provisioner.go
+++ b/internal/controller/infrastructure/ssh_provisioner.go
@@ -27,6 +27,7 @@ import (
 	"regexp"
 	"strings"
 
+	"al.essio.dev/pkg/shellescape"
 	"github.com/go-logr/logr"
 	"github.com/k0sproject/rig"
 	"github.com/k0sproject/rig/exec"
@@ -138,7 +139,8 @@ func (p *SSHProvisioner) Provision(ctx context.Context) error {
 		// Run commands
 		for _, cmd := range p.cloudInit.Commands {
 			p.log.Info("running command", "command", cmd)
-			output, err := connection.ExecOutput(cmd, execOpts...)
+			cmdWithShell := "sh -c " + shellescape.Quote(cmd)
+			output, err := connection.ExecOutput(cmdWithShell, execOpts...)
 			if err != nil {
 				p.log.Error(err, "failed to run command", "command", cmd, "output", output)
 				return fmt.Errorf("failed to run command: %w", err)


jnummelin (approved):

makhov:
I'm not fully sure, but looks like this also fixes #1500, so probably should be backported to v1.10 release branch

PR #1506, opened 2026-06-29
Run `check-capi-remote-machine` inttest with sudo and non-root user

apedriza:


inttest/capi-remote-machine/capi_remote_machine_test.go
```diff
@@ -338,7 +339,8 @@ metadata:
 spec:
   address: {{ .Address }}
   port: 22
-  user: root
+  user: test
+  useSudo: true
```
twz123, line 343: Just wondering: Do we support doas as well? 🙈

twz123 (approved):

inttest/capi-remote-machine/capi_remote_machine_test.go
apedriza: only sudo for escalation

Issue #1492, opened 2026-06-19
Ingress mode: worker bootstrap deadlock — kubernetes Service missing publishNotReadyAddresses causes empty EndpointSlice on NotReady nodes


valentynkhalin:
## Summary

  When a hosted cluster is deployed with `ingress.enabled: true`, worker nodes enter a
  deadlock during bootstrap and never reach Ready state. The root cause is that the
  `kubernetes` Service in the child cluster (which routes API traffic through the local
  haproxy DaemonSet) does not set `publishNotReadyAddresses: true`. On a NotReady worker,
  the endpoint controller excludes the haproxy pod from the EndpointSlice — so all
  bootstrap components that need the Kubernetes API (calico, kube-proxy, konnectivity)
  hit `10.96.0.1:443` which is unreachable in ingress topology, and the node never
  becomes Ready.

  ## Steps to reproduce

  1. Deploy a k0smotron hosted cluster with ingress enabled:
     ```yaml
     ingress:
       enabled: true
       apiHost: api.<name>.<ip>.sslip.io
       konnectivityHost: konnectivity.<name>.<ip>.sslip.io
  2. Wait for a worker VM to register with the child cluster.
  3. On the child cluster, inspect the kubernetes Service Endpoints:
  kubectl get endpointslices -n default
  kubectl get endpoints kubernetes -n default -o yaml
  4. Observe calico-node install-cni entering CrashLoopBackOff:
  kubectl get pods -n kube-system -l k8s-app=calico-node
  kubectl logs -n kube-system <calico-pod> -c install-cni
  # dial tcp 10.96.0.1:443: connect: connection refused
  5. Node stays NotReady indefinitely.

  Root cause

  The ingress manifests ConfigMap (generateIngressManifestsConfigMap) deploys:

  3_kube-service.yaml — the kubernetes Service with selector
  app: k0smotron-haproxy:
  spec:
    selector:
      app: k0smotron-haproxy
    # publishNotReadyAddresses is absent

  The haproxy DaemonSet (2_haproxy-ds.yaml) correctly tolerates not-ready and
  uninitialized nodes, so the haproxy pod IS scheduled on the NotReady worker.
  However, without publishNotReadyAddresses: true on the Service, the endpoint
  controller excludes all NotReady pods — including the haproxy pod — from the
  EndpointSlice. The EndpointSlice for kubernetes is empty.

  A static Endpoints object (0_temp-kubernetes-ep.yaml) with placeholder IP
[...]

PR #1507, opened 2026-06-30
Ingress test with calico

makhov:
Test ingress solutions for both supported CNIs. Calico does call kubernetes default service


apedriza:
Let me understand calico case. The cni not reaching k8s api is a temporary issue until node becomes ready itself? 

So this differs from the case when the cloud provider is external and CCM needs manual intervention?

e2e/data/infrastructure-docker/main/cluster-template-ingress-calico/cluster-with-ingress-calico.yaml
apedriza, line 1: What about reusing the same ingress manifest for both cases and configure the clusterconfig with a kustomize patch? so each one declares its own config but reuse the rest. kustomize should do its stuff internally in the capi framework

makhov: Done, thanks!

makhov:
> Let me understand calico case. The cni not reaching k8s api is a temporary issue until node becomes ready itself?

Yes. I haven't been able to reproduce #1492, we need more details about the setup and configuration

apedriza (approved):

PR #1509, opened 2026-07-01
Namespace-scoped count selector

### LINKED ISSUE #1508: K0smotronControlPlane replica count not namespace-scoped
K0smotronControlPlane status counts control-plane pods cluster-wide using only the label `cluster.x-k8s.io/cluster-name=<name>`, without namespace scoping. Two clusters sharing a name in different namespaces get inflated replica counts.

makhov:
Fixes #1508


apedriza:
Aa a future PR I think worth including e2es where more than one cluster is deployed, for both, hcp and vms setups in order to catch earlier this cases. A basic test spec check should catch this cases just waiting for ready replicas or smtg like that. Will take a look

apedriza (approved):

PR #1514, opened 2026-07-08
Windows ingress support

makhov:
Switching to traefik which supports windows


internal/controller/k0smotron.io/k0smotroncluster_ingress.go
```diff
@@ -35,39 +39,80 @@ func (scope *kmcScope) reconcileIngress(ctx context.Context, kmc *km.Cluster) er
...
+func upsertIngressManifestVolumes(kmc *km.Cluster) {
 	ingressVolume := corev1.Volume{
 		Name: kmc.GetIngressManifestsConfigMapName(),
 		VolumeSource: corev1.VolumeSource{
-			ConfigMap: &corev1.ConfigMapVolumeSource{
-				LocalObjectReference: corev1.LocalObjectReference{
-					Name: kmc.GetIngressManifestsConfigMapName(),
-				},
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: kmc.GetIngressManifestsConfigMapName(),
```
apedriza:
```suggestion
				SecretName: kmc.GetIngressManifestsSecret(),
```
Now that it is stored as a secret

Anyway, this name is used in several places not related to configmaps/secrets. I would use something mroe general like `GetIngressManifestsResourceName` or something like that

internal/controller/k0smotron.io/k0smotroncluster_ingress.go
```diff
@@ -35,39 +39,80 @@ func (scope *kmcScope) reconcileIngress(ctx context.Context, kmc *km.Cluster) er
...
+
+	proxyCert, proxyKey, caCert, err := scope.loadIngressCertMaterial(ctx, kmc)
 	if err != nil {
-		return fmt.Errorf("failed to generate ingress manifests configmap: %w", err)
+		return fmt.Errorf("failed to load ingress cert material: %w", err)
 	}
-	_ = kcontrollerutil.SetExternalOwnerReference(kmc, &configMap, scope.client.Scheme(), scope.externalOwner)
-	err = scope.reconcileResource(ctx, kmc, &configMap)
+
+	configMap, err := scope.generateIngressManifestsConfigMap(kmc, proxyCert, proxyKey, caCert)
```
apedriza: Why not create the secret directly instead of creating it from a non-stored ConfigMap?

internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -224,6 +224,13 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{Requeue: true, RequeueAfter: time.Minute}, err
 	}
 
+	if kmc.Spec.Ingress != nil {
+		err := kmcScope.ensureHAProxyCerts(ctx, kmc)
+		if err != nil {
+			return ctrl.Result{Requeue: true, RequeueAfter: time.Minute}, fmt.Errorf("error generating ingress certificates: %w", err)
+		}
+	}
+
```
apedriza: Can we move this within `reconcileIngress`? I would encapsulate all the old logic support/considerations deeper, otherwise can be a bit confusing from a high-level POV and rename cert from `ingress-haproxy` to `ingress-proxy`, just in case proxy implementation is changed in the future

apedriza (commented):
has the `TestWindowsIngressProvisioning` been executed succesfully? We could manually trigger before merge it as currently will not run until it is on main

internal/controller/k0smotron.io/k0smotroncluster_controller.go
makhov: Done

internal/controller/k0smotron.io/k0smotroncluster_ingress.go
makhov: Right, fixed!

internal/controller/k0smotron.io/k0smotroncluster_ingress.go
makhov: changed

apedriza (commented):
We should also update current docs talking about haproxy DS

apedriza (commented):
Can we force the aws workflow? running this test from local I see `IMDS token unavailable, cannot auto-detect host internal IP`

makhov:
@apedriza I've fixed the test, it passes now: https://github.com/k0sproject/k0smotron/actions/runs/32474284401

PR #1515, opened 2026-07-08
Fix reconcile loop when machine is already updated for inplace updates

apedriza:
After recent changes in machine reconciliation, the controller is always requeuing a reconciliation based on machine.spec.version, which is not updated when inplace updates are used. 

With these changes, inplace update logic also updates the machines.spec.version after update is made by autopilot. Also, fixing a bug for only use InPlace update if only the k0s version changes


inttest/capi-docker-machine-template-update/capi_docker_machine_template_update_test.go
```diff
@@ -308,9 +308,6 @@ spec:
   replicas: 3
   version: v1.30.1+k0s.0
   k0sConfigSpec:
-    postStartCommands:
```
apedriza, line 311: Now that the bug about not recreating machines when k0scontrolplane spec changes is fixed, the `postStartCommands` addition will trigger a recreation of the machine. This test is about InPlace updates so removing that update to not trigger a recreation but just a k0s version inplace update.

Probably this test and `check-capi-docker-machine-template-update-recreate` are a bit redundant compared them with `TestWorkloadClusterUpgrade` e2e

makhov (approved):

PR #1525, opened 2026-07-25
Fix K0sWorkerConfig ignoring Machine's spec.version

### LINKED ISSUE #1524: K0sWorkerConfig doesn't respect Machine spec.version
### What happened?

We upgraded recently to k0smotron v2.0.4 and ran into a bug where K0sWorkerConfig without version always install the latest k0s version. This is problematic especially when using a ClusterClass since the k0s version is reconciled from Machine objects.

I didn't read the commit in detail, but the issue might be coming from [this change](https://github.com/k0sproject/k0smotron/commit/f1e352029dadccab25dd40ab581cb369ca6ccfa5#diff-f26f5429a4b202f0cb3685ad5134a990b698d3071ba25f7e81b24e13acc866ddL140-L143).

### Involved Areas

Cluster API integration

K0sWorkerConfig should use the Machine spec.Version and fallback to latest when not set.

### Steps to reproduce

1. Creating a ClusterClass using k0smotron
```yaml
# Will only show the relevant K0sWorkerConfigTemplate here
---
kind: K0sWorkerConfigTemplate
metadata:
  name: worker
spec:
  template:
    spec:
      useSystemHostname: true
```

2. Creating a cluster using the ClusterClass
```yaml
apiVersion: cluster.x-k8s.io/v1beta2
kind: Cluster
metadata:
  name: test-cluster
spec:
  clusterNetwork:
    pods:
      cidrBlocks:
        - 192.168.0.0/16
    serviceDomain: cluster.local
    services:
      cidrBlocks:
        - 10.128.0.0/12
  topology:
    classRef:
      name: clusterclass
    version: v1.35.3
    controlPlane:
      replicas: 3
    workers:
      machineDeployments:
        - class: worker
          name: worker
          replicas: 3
```
3. Workers will attempt to join the cluster using latest k0s version instead of v1.35.3 resulting in Machines being stuck

### k0smotron version

v2.0.4

### k0s version

v1.35.3

We haven't upgraded our manifests yet to v1beta2. So I can only confirm that this behaviour happens on v1beta1, however I suspect this issue will be present when using v1beta2.

pujitha24:
Motivation:
When a K0sWorkerConfig has no Spec.Version set, the worker bootstrap
controller always installs the latest k0s release instead of falling
back to the owning Machine's (or MachinePool's) Kubernetes version.
This is especially problematic with ClusterClass, where the k0s
version is set on the Machine via the Cluster topology and
K0sWorkerConfigTemplate intentionally omits Spec.Version. Workers then
join with an unexpected k0s version, which can get stuck instead of
joining the cluster with the version the topology requested.

The fallback was previously implemented by converting the config
owner to a Machine and reading its Spec.Version, but that code was
removed in a refactor (f1e352029dadccab25dd40ab581cb369ca6ccfa5)
without a replacement.

Approach:
Restore the fallback using configOwner.KubernetesVersion(), a helper
already available on CAPI's bsutil.ConfigOwner that reads spec.version
for a Machine and spec.template.spec.version for a MachinePool. This
covers both owner kinds, unlike the original manual Machine
conversion. The fallback runs before the existing "-k0s." to "+k0s."
normalization and default-suffix logic, so a version picked up from
the owner is normalized the same way an explicitly set version is.
An explicitly set Spec.Version is never overwritten.

Validation:
Added an envtest-tagged regression test,
TestReconcileWorkerConfigVersionFallsBackToMachineVersion, which
creates a Machine with spec.version "v1.30.5" and a K0sWorkerConfig
with no version owned by it, reconciles to completion, and asserts
the generated bootstrap secret's install script contains
"K0S_VERSION=v1.30.5+k0s.0" rather than installing k0s with no
version pin (which would resolve to latest). Confirmed the test fails
without the fix (falls back to installing latest) and passes with it.

Ran:
  KUBEBUILDER_ASSETS="$(./bin/setup-envtest use 1.32.0 --bin-dir "$(pwd)/bin" -p path)" \
    go test ./internal/controller/bootstrap/... -tags envtest
  go test ./internal/controller/bootstrap/...
  go build ./...
  go vet ./internal/controller/bootstrap/...
All passed.

Fixes #1524

### DIFF
--- a/internal/controller/bootstrap/worker_bootstrap_controller.go
+++ b/internal/controller/bootstrap/worker_bootstrap_controller.go
@@ -130,6 +130,12 @@ func (r *Controller) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.
 
 	log = log.WithValues("kind", configOwner.GetKind(), "version", configOwner.GetResourceVersion(), "name", configOwner.GetName())
 
+	// If the K0sWorkerConfig does not have a version set, fall back to the owner's
+	// (Machine or MachinePool) Kubernetes version, e.g. as set by a ClusterClass topology.
+	if config.Spec.Version == "" {
+		config.Spec.Version = configOwner.KubernetesVersion()
+	}
+
 	// If the version does not contain the k0s suffix, append it.
 	if config.Spec.Version != "" {
 		// When machine is created by CAPI, for example by using a clusterclass template, the version
--- a/internal/controller/bootstrap/worker_bootstrap_controller_env_test.go
+++ b/internal/controller/bootstrap/worker_bootstrap_controller_env_test.go
@@ -481,6 +481,107 @@ func TestReconcileControlPlaneNotReady(t *testing.T) {
 	}, 10*time.Second, 100*time.Millisecond)
 }
 
+func TestReconcileWorkerConfigVersionFallsBackToMachineVersion(t *testing.T) {
+	ns, err := testEnv.CreateNamespace(ctx, "test-reconcile-workerconfig-version-fallback")
+	require.NoError(t, err)
+
+	cluster, kcp, _ := createClusterWithControlPlane(ns.GetName())
+	require.NoError(t, testEnv.Create(ctx, cluster))
+	require.NoError(t, testEnv.Create(ctx, kcp))
+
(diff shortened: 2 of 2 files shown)


apedriza (approved):
LGTM

PR #1529, opened 2026-07-30
deprecate version field in k0s config resources

apedriza:
link a k0sworkerconfig with a fixed version does not allow a mahcine deployment version using cluster topology. When a new version in cluster.spec.topology.version is detected, machines controller tries to roll out the machine but the machine deployment is still using  the version from its k0sworkerconfig. The k0s version should be set from the machine resources instead, that way the new machine will still use the old k0sworkerconfig but getting the new version set by the machine owner controller

### DIFF
--- a/api/bootstrap/v1beta2/k0s_types.go
+++ b/api/bootstrap/v1beta2/k0s_types.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/k0sproject/k0smotron/v2/internal/provisioner"
 )
@@ -135,6 +136,7 @@ type K0sWorkerConfigSpec struct {
 	// Make sure the version is compatible with the k0s version running on the control plane.
 	// For reference see the Kubernetes version skew policy: https://kubernetes.io/docs/setup/release/version-skew-policy/
 	// +kubebuilder:validation:Optional
+	// Deprecated: Use Machine version instead.
 	Version string `json:"version,omitempty"`
 
 	// UseSystemHostname specifies whether to use the system hostname for the kubernetes node name.
@@ -282,6 +284,7 @@ type K0sControllerConfigSpec struct {
 	// Make sure the version is compatible with the k0s version running on the control plane.
 	// For reference see the Kubernetes version skew policy: https://kubernetes.io/docs/setup/release/version-skew-policy/
 	// +kubebuilder:validation:Optional
+	// Deprecated: Use Machine version instead.
 	Version string `json:"version,omitempty"`
 
 	*K0sConfigSpec `json:",inline"`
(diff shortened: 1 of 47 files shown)


api/bootstrap/v1beta2/k0s_types.go
```diff
@@ -621,9 +626,16 @@ func (cs *K0sWorkerConfigSpec) validateFiles(pathPrefix *field.Path) field.Error
...
 
-func (cs *K0sWorkerConfigSpec) validateVersion(pathPrefix *field.Path) field.ErrorList {
+const deprecatedK0sConfigVersionField = "K0sWorkerConfig.spec.version is deprecated and will be removed in a future release. It currently takes precedence over the Machine's version field."
+
+func (cs *K0sWorkerConfigSpec) validateVersion(pathPrefix *field.Path) (admission.Warnings, field.ErrorList) {
 	var allErrs field.ErrorList
 
+	var warnings admission.Warnings
+	if cs.Version != "" {
+		warnings = append(warnings, deprecatedK0sConfigVersionField)
```
jnummelin, line 638: Should the below `validateWindows` remove the version validation entirely? Or at least skip it if it is not set?

api/bootstrap/v1beta2/k0s_types.go
```diff
@@ -282,6 +284,7 @@ type K0sControllerConfigSpec struct {
 	// Make sure the version is compatible with the k0s version running on the control plane.
 	// For reference see the Kubernetes version skew policy: https://kubernetes.io/docs/setup/release/version-skew-policy/
 	// +kubebuilder:validation:Optional
+	// Deprecated: Use Machine version instead.
```
jnummelin, line 289: Should `K0sControllerConfig` webhook validate the version too and emit a warning?

api/bootstrap/v1beta2/k0s_types.go
```diff
@@ -621,9 +626,16 @@ func (cs *K0sWorkerConfigSpec) validateFiles(pathPrefix *field.Path) field.Error
 	return allErrs
 }
 
-func (cs *K0sWorkerConfigSpec) validateVersion(pathPrefix *field.Path) field.ErrorList {
+const deprecatedK0sConfigVersionField = "K0sWorkerConfig.spec.version is deprecated and will be removed in a future release. It currently takes precedence over the Machine's version field."
```
jnummelin, line 631: Do we have testcases for the precedence logic?

jnummelin (changes requested):
Couple things IMO worth of fixing or at least looking into. Also make sure we do not hit any nil pointers in bootstrap cloud-init generations if we still directly use the spec.version somewhere.

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -389,7 +389,6 @@ func (c *K0sController) createBootstrapConfig(ctx context.Context, name string,
 			}},
 		},
 		Spec: bootstrapv2.K0sControllerConfigSpec{
-			Version:       kcp.Spec.Version,
```
apedriza, line 392: We don't set k0scontrollerconfig.spec.version anymore. Use the version from its owner, i.e, k0scontrolplane.

api/bootstrap/v1beta2/k0s_types.go
apedriza: That resource must be created only by the control plane controller so we don't need admission validation as we control how we create it. [See](https://github.com/k0sproject/k0smotron/pull/1529/changes#r4080634126)

api/bootstrap/v1beta2/k0s_types.go
apedriza: Changed! the webhook now only validates spec.version when set, since the version may come from the Machine. The reconciler checks the resolved version so when platform Windows is set, it's required and must be >= v1.34.2+k0s.0, otherwise we report and error about using a version which dooesnt support windows

api/bootstrap/v1beta2/k0s_types.go
apedriza: Added tests for the block in the reconciler about how version is resolved

apedriza:
> Also make sure we do not hit any nil pointers in bootstrap cloud-init generations if we still directly use the spec.version somewhere.

now we use the resolved version and not the one set in the resoruce (if set). The version is an string which can accept a zero value in linux which uses the default k0s version in the installation command and a resolved version with zero value in windows reports an error

PR #1535, opened 2026-08-03
Fix duplicate Machine creation when cache lags after scale-up

### LINKED ISSUE #1534: K0sControlPlane creates duplicate Machines when cached state lags after scale-up
## What happened?

A `K0sControlPlane` configured with `spec.replicas: 1` created six control-plane Machines during initial provisioning. Five of them progressed to running EC2 instances before the controller observed the actual Machine count and started scaling back down to one.

The desired replica count remained 1 throughout the incident.

Sanitized controller timeline:

```text
12:25:08 active=0 upToDate=0 desired=1 -> Scaling up
12:25:08 active=0 upToDate=0 desired=1 -> Scaling up
12:25:08 active=0 upToDate=0 desired=1 -> Scaling up
12:25:18 active=0 upToDate=0 desired=1 -> Scaling up
12:25:28 active=0 upToDate=0 desired=1 -> Scaling up
12:25:38 active=0 upToDate=0 desired=1 -> Scaling up

Later:
active=6 upToDate=6 desired=1 -> Scaling down
```

Each reconciliation generated a different random Machine name, so every retry created another valid Machine.

This appears to be a read-after-write/cache-observation issue in the scale-up path:

1. `reconcileMachines` determines that scale-up is required from `scope.activeMachines`.
2. `scaleUp` creates the infrastructure object, bootstrap object, and Machine.
3. `reconcileMachines` calls `retrieveControlPlaneState` immediately after the write.
4. `retrieveControlPlaneState` lists Machines through `c.Client`, which is the manager's cached client.
5. While the informer cache has not observed the new Machine, the refreshed scope still contains zero active Machines.
6. The controller requeues and creates another randomly named Machine.

Relevant code:

- https://github.com/k0sproject/k0smotron/blob/v2.0.4/internal/controller/controlplane/scale.go#L43-L99
- https://github.com/k0sproject/k0smotron/blob/v2.0.4/internal/controller/controlplane/scale.go#L204-L249
- https://github.com/k0sproject/k0smotron/blob/v2.0.4/internal/controller/controlplane/k0s_controlplane_controller.go#L745-L807

The controller deployment had two replicas with leader election enabled. Only the leader emitted these reconciliation logs, so this was not caused by both manager Pods acting as leaders.

The default `--concurrency=5` may widen the observation window, but reducing concurrency alone does not appear sufficient because the explicit 10-second retries also continued seeing `active=0`.

## Involved Areas

Cluster API integration

## What did you expect to happen?

Exactly one control-plane Machine should be created when `spec.replicas: 1`.

A successful scale-up operation should be idempotent across subsequent reconciliations, even when the controller-runtime cache has not observed the newly created Machine yet.

## Steps to reproduce

1. Install Cluster API and k0smotron v2.0.4.
2. Run the k0smotron control-plane controller with its default concurrency and leader election enabled.
3. Create a workload Cluster and a `K0sControlPlane` with:
   ```yaml
   spec:
     replicas: 1
     updateStrategy: InPlace
   ```
4. Use an infrastructure provider where infrastructure readiness and Machine creation generate several reconciliation events.
5. Observe the k0smotron control-plane controller logs and count the owned Machines.

The issue is most visible when the Machine informer/cache is delayed relative to API writes. A deterministic regression test could use a reader that continues returning the pre-create Machine list for several reconciliations.

## k0smotron version

v2.0.4 (`b1ccce1197d0248e7ca9e7d0f9b383dbe36fbaab`)

## k0s version

v1.33.2+k0s.0

## Anything else we need to know?

Cluster API version: v1.11.4  
Infrastructure provider: CAPA  
Machine update strategy: `InPlace`

Potential fixes:

- Track an expected/pending Machine creation per K0sControlPlane and block further scale-up until the informer observes it or an authoritative API read confirms it does not exist.
- Perform an uncached authoritative count immediately before creating another Machine.
- Avoid relying only on randomly named resources for retry idempotency.
- Add a regression test that simulates cache lag after a successful Machine write.

The current `main` branch was also checked and does not appear to contain a fix for this scale-up path.

liuxu623:
## What this PR does

Fixes #1534

`K0sControlPlane` scale-up creates Machines with randomly generated names. When the controller-runtime informer cache lags behind the API server, reconciliations following a successful creation still see the stale Machine list (`active=0`), generate a new random name, and create duplicate Machines — six Machines were observed for `spec.replicas: 1`.

This PR adds an **authoritative pre-create check**: right before creating a Machine, an uncached List via the new `APIReader` verifies the API server doesn't know any Machines the cache hasn't observed yet, and requeues if it does. Reading directly from the API server (rather than tracking process-local state) means the guard also covers leader changes, controller restarts, and writes that time out on the client but are committed by the API server. The check only runs when a scale-up is actually being considered, so it is not on the hot path.

## Testing

Added `TestReconcileMachinesScaleUpIdempotentWithStaleCache`, which simulates permanent cache lag with a client whose Machine lists never return new Machines, and asserts exactly one Machine is created across repeated reconciliations — including after a simulated leader change (a fresh controller instance with no process-local state).

### DIFF
--- a/internal/controller/controlplane/scale.go
+++ b/internal/controller/controlplane/scale.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/utils/ptr"
 	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	"sigs.k8s.io/cluster-api/util/collections"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/failuredomains"
 	"sigs.k8s.io/cluster-api/util/patch"
@@ -67,6 +68,22 @@ func (c *K0sController) reconcileMachines(ctx context.Context, scope *controlpla
 
 	switch {
 	case isNeededScaleUp(scope):
+		// Authoritative guard before creating a Machine: the informer cache can lag behind the
+		// API server, so a scale-up decision based on the cached scope may undercount Machines
+		// and, because Machines get random names, create duplicates (see issue #1534). Read
+		// directly from the API server and never create a new Machine while it knows Machines
+		// the cache has not observed yet.
+		stale, err := c.isMachineViewStale(ctx, scope)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("error verifying machine view against the API server: %w", err)
+		}
+		if stale {
+			// The cache usually catches up within milliseconds, so requeue quickly instead of
+			// stalling the scale-up. Same duration Cluster API uses for stale-cache requeues.
+			logger.Info("Cached machine view is stale compared to the API server, requeuing before scaling up")
+			return ctrl.Result{RequeueAfter: 100 * time.Millisecond}, nil
+		}
+
 		logger.Info("Scaling up control plane")
 		if err := c.scaleUp(ctx, scope); err != nil {
 			return ctrl.Result{}, fmt.Errorf("error scaling up control plane: %w", err)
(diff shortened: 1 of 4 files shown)


internal/controller/controlplane/scale.go
```diff
@@ -69,6 +82,19 @@ func (c *K0sController) reconcileMachines(ctx context.Context, scope *controlpla
 
 	switch {
 	case isNeededScaleUp(scope):
+		// Final authoritative guard before creating a Machine: the in-memory expectations
+		// checked above can be lost on leader changes, controller restarts, or writes that
+		// timed out on the client but were committed by the API server. Never create a new
+		// Machine while the API server knows Machines the cache has not observed yet.
+		stale, err := c.isMachineViewStale(ctx, scope)
```
apedriza, line 76: is this not the only check we should run before scale opt? checking cache state is up to date with etcd..? Why adding only this is not enough?

apedriza (commented):
In general we should make sure every reconcile event is processed only if cache is up to date.  So adding a more general approach for reconciled resources as CAPI does I think is the way to go in future, probably making use of the resource version of the resource

internal/controller/controlplane/scale.go
liuxu623: Good point — collapsed it to just the authoritative pre-create check (isMachineViewStale). The in-memory expectations were a strict subset of what the uncached read already covers, and the read only runs when a scale-up is actually being considered (not on every requeue), so the "fast path" wasn't buying much.

liuxu623:
> In general we should make sure every reconcile event is processed only if cache is up to date. So adding a more general approach for reconciled resources as CAPI does I think is the way to go in future, probably making use of the resource version of the resource

Agreed — CAPI landed exactly that in v1.14 as the "stale controller mitigation": a consistencyStore in the public util/controller package that records the resourceVersion of each written child object and defers reconciles until the informer cache catches up. Adopting it requires CAPI v1.14 + k8s.io v0.36 libs (it relies on the new official resourceversion.CompareResourceVersion), so it's one for after we bump deps.

Note it complements rather than replaces the check in this PR: the consistencyStore is in-memory, so it can't cover writes that time out on the client but are committed by the API server, which the authoritative pre-create read here does.

apedriza:
@liuxu623 can you sign the commit? DCO check is failing

liuxu623:
> @liuxu623 can you sign the commit? DCO check is failing

done

apedriza (approved):

PR #1542, opened 2026-08-10
Add e2e checks for each upgrade strategy

### LINKED ISSUE #1541: Include testing logic per update strategy
Currently, we are running same test spec for every update strategy which is good in terms of no code-duplications but we are not adding custom checks depending on the strategy. 

For instance, inplace upgrade should check there is no mahcine recreated, RecreateDeleteFirst should check there are no more machines than desired..etc. Probably some check injection would be great in order to avoid if-elses depending on the strategy

junnhwan:
**What this PR does / why we need it:**

Adds strategy-specific e2e checks for control plane upgrades in the workload cluster upgrade spec:

- `InPlace` (standalone): verifies an autopilot Plan appears in the workload cluster and that no control plane Machine carries the CAPI in-place update annotations during the upgrade.
- `InPlaceCAPI` (runtime extension): verifies at least one control plane Machine carries the CAPI in-place update annotations during the upgrade.
- Both in-place flavors: verifies no control plane Machine is recreated.
- `Recreate` / `RecreateDeleteFirst`: verifies the final control plane Machines are all new (no UID overlap with the pre-upgrade set) and the final count matches the desired replicas.
- `RecreateDeleteFirst`: verifies the control plane Machine count never exceeds the desired replicas while the upgrade is in progress.

The checks are injected into the shared upgrade helper via `DuringUpgradeCheck` / `PostUpgradeCheck` hooks, keeping the spec shared across flavors.

**Which issue(s) this PR fixes:**

Fixes: https://github.com/k0sproject/k0smotron/issues/1541

**Test plan:**

- `gofmt` and `git diff --check` pass.
- `go test -tags e2e -c ./e2e/` compiles the e2e test binary.
- Full e2e runs require kind/Docker and were not executed locally; they run in the project e2e CI.

### DIFF
--- a/e2e/util/controlplane.go
+++ b/e2e/util/controlplane.go
@@ -83,11 +83,34 @@ type UpgradeControlPlaneAndWaitForUpgradeInput struct {
 	KubernetesUpgradeVersion         string
 	WaitForKubeProxyUpgradeInterval  Interval
 	WaitForControlPlaneReadyInterval Interval
+
+	// DuringUpgradeCheck is an optional strategy-specific check that runs in the
+	// background while the upgrade is in progress. It returns a cleanup function
+	// that is invoked after the upgrade completes; the returned error is checked then.
+	DuringUpgradeCheck func(ctx context.Context, input UpgradeControlPlaneAndWaitForUpgradeInput) (func() error, error)
+
+	// PostUpgradeCheck is an optional strategy-specific check invoked after the
+	// control plane reports ready but before the helper returns.
+	PostUpgradeCheck func(ctx context.Context, input UpgradeControlPlaneAndWaitForUpgradeInput) error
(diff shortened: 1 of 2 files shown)


apedriza:
I think e2e test is facing a condition race when the check runs faster than the delete + recreation of the new plan. It is expected that the plan get preserved even if it is completed. We should take this in the test spec

junnhwan:
Thanks for the review. I updated the standalone in-place check so a completed plan from the previous upgrade is treated as an expected intermediate state: the check keeps polling until the new plan for the target upgrade version appears, instead of failing on the version mismatch. The e2e package compiles with \go test -tags e2e -c ./e2e/\.

apedriza (approved):

PR #1545, opened 2026-08-11
Fix stale machine read on cache lags

makhov:
Followup for #1535

Remove the staleness instead of detecting it


apedriza (approved):
As a side note, I plan to write some pre-reconcile logic to verify resource version in cache  >= resource version of latest scale/write operation

PR #1562, opened 2026-08-18
Certs rotation

makhov:


api/k0smotron.io/v1beta2/k0smotroncluster_webhook.go
```diff
@@ -115,6 +127,41 @@ func (c ClusterValidator) validatePatches(patches []ComponentPatch) error {
...
+// Cluster would let an invalid K0smotronControlPlane be admitted and then have
+// its child Cluster Create rejected by this same rule inside the controller,
+// leaving the hosted control plane permanently unprovisioned with the reason
+// visible only in controller logs.
+func validateCertificates(kcs *ClusterSpec) error {
+	if kcs.Certificates == nil {
+		return nil
+	}
+
+	if kcs.Certificates.Duration != nil && kcs.Certificates.Duration.Duration <= 0 {
```
apedriza, line 200: can `RenewBefore` be negative? I guess not transitively (with checks kcs.Certificates.Duration.Duration <= 0 check and kcs.Certificates.RenewBefore.Duration >= kcs.Certificates.Duration.Duration) but maybe it wouldn't hurt to make it visible

internal/controller/controlplane/k0s_controlplane_controller.go
apedriza, line 1: can we trigger a rollout if certs are expired? for example, kubeadmcontrolplane expose an option to rollout when cert are expired. We could offer something similar..?

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -679,6 +698,13 @@ func (c *K0sController) reconcileDelete(ctx context.Context, controlplane *contr
 
 	clearAvailabilityFailures(&c.availabilityFailures, controlplane.kcp)
 
+	// Drop this cluster's certificate metric series first, before any
+	// error-prone step below can return early: a deleted cluster must stop
+	// reporting a certificate that will never be renewed. Resetting early is
+	// safe - if deletion is somehow abandoned, the next successful reconcile
+	// re-emits the series.
+	metrics.ResetCluster(controlplane.cluster.Namespace, controlplane.cluster.Name)
```
apedriza: We are adding metrics too. Should we do this in a separate PR and evaluate which metrics are worth adding?

internal/controller/controlplane/reconcile_certificate_conditions_test.go
apedriza, line 1: should we move this to certificates_test.go instead? the file name and test location looks a bit out of the current naming approach

internal/metrics/certificates.go
apedriza, line 1: We are adding metrics too. Should we do this in a separate PR and evaluate which metrics are worth adding?

PR #1564, opened 2026-08-20
feat: bootstrap file owner/encoding/append

s3rj1k:
Adds three fields to bootstrap files (mirrors `kubeadm`): `owner`, `encoding` and `append`, `encoding` only takes base64.

cloud-init and Ignition use all three. Powershell cannot set an owner, so it
rejects it. RemoteMachine cannot append, so it rejects that.

Additional bug fixes: directories are made 0755 instead of copying the file mode, powershell no longer breaks content that is not UTF-8, and paths are escaped in the CloudInitVars block.


makhov (approved):

makhov (approved):

s3rj1k:
Thanks!

PR #1567, opened 2026-08-24
fix: report the control plane as unavailable when it cannot be reached

s3rj1k:
Fixes a control plane that cannot be reached still reporting itself as available.

### DIFF
--- a/internal/controller/controlplane/inplace.go
+++ b/internal/controller/controlplane/inplace.go
@@ -45,13 +45,21 @@ import (
 )
 
 func (c *K0sController) reconcileInplaceK0sVersionUpdate(ctx context.Context, scope *controlplane) (ctrl.Result, error) {
+	controlplaneRequiresUpdate := scope.hasMachinesWithOnlyVersionOutdated && scope.kcp.Spec.UpdateStrategy == cpv1beta2.UpdateInPlace
+
 	if !conditions.IsTrue(scope.kcp, cpv1beta2.ControlPlaneAvailableCondition) {
 		// If the control plane is not available, we cannot proceed with the in-place update, as access to the
 		// workload cluster is required to manage the autopilot plan.
+
+		// Falling through would replace machines an in place update should upgrade
+		// where they are. Only the rollout is held, never a count that does not match.
+		countMatches := scope.activeMachines.Len() == int(scope.kcp.Spec.Replicas)
+		if controlplaneRequiresUpdate && countMatches {
+			return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+
 		return ctrl.Result{}, nil
 	}
-
-	controlplaneRequiresUpdate := scope.hasMachinesWithOnlyVersionOutdated && scope.kcp.Spec.UpdateStrategy == cpv1beta2.UpdateInPlace
 
 	logger := log.FromContext(ctx).WithValues("version", scope.kcp.Spec.Version)
 
--- a/internal/controller/controlplane/k0s_controlplane_controller.go
+++ b/internal/controller/controlplane/k0s_controlplane_controller.go
@@ -147,8 +147,6 @@ func (c *K0sController) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 
 	if err := c.Get(ctx, req.NamespacedName, kcp); err != nil {
 		if apierrors.IsNotFound(err) {
-			// The count cannot be dropped here, since the key is the UID and this
-			// object was never read. reconcileDelete covers the ordinary path.
 			return ctrl.Result{}, nil
 		}
 		log.Error(err, "Failed to get K0sControlPlane")
--- a/internal/controller/controlplane/k0s_controlplane_controller.go
+++ b/internal/controller/controlplane/k0s_controlplane_controller.go
@@ -24,7 +24,9 @@ import (
 	"sync"
 	"time"
 
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 
 	"github.com/google/uuid"
 	autopilot "github.com/k0sproject/k0s/pkg/apis/autopilot/v1beta2"
@@ -51,6 +53,7 @@ import (
 	"sigs.k8s.io/cluster-api/util/collections"
 	"sigs.k8s.io/cluster-api/util/kubeconfig"
 	"sigs.k8s.io/cluster-api/util/patch"
+	"sigs.k8s.io/cluster-api/util/predicates"
 	"sigs.k8s.io/cluster-api/util/secret"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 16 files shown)


internal/controller/controlplane/status.go
```diff
@@ -182,6 +183,46 @@ func versionMatches(machine *clusterv1.Machine, ver string) bool {
 	return vKCP.Equal(vMachine)
 }
 
+// truncateMessage keeps a condition message within what the CRD accepts, since
+// a longer one would make every status patch fail validation.
```
makhov: which CRD has this limit? I haven't heard about it before

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -862,16 +865,44 @@ func (c *K0sController) calculateMachineState(ctx context.Context, kcp *cpv1beta
...
-		Complete(c)
+		Owns(&clusterv1.Machine{})
+
+	// Nothing schedules a reconcile while the control plane reports available, so
+	// without this a cluster whose connection drops is only noticed by chance.
+
+	// Connect and disconnect events need no option. WatchForProbeFailure exists to
+	// drive staleness grace periods, which this provider does not have.
+	if c.ClusterCache != nil {
+		builder = builder.WatchesRawSource(c.ClusterCache.GetClusterSource("k0scontrolplane", clusterToK0sControlPlane))
```
makhov: Will it work for [сlient connection tunneling](https://docs.k0smotron.io/stable/capi-controlplane-bootstrap/#client-connection-tunneling)?

internal/controller/controlplane/status.go
```diff
@@ -205,6 +265,17 @@ func (c *K0sController) computeAvailability(ctx context.Context, controlplane *c
...
+
+		// While coming up the contract fallback says this better than any
+		// internal error would.
+		if !initialized(controlplane.kcp) {
+			return
+		}
+
+		// The cache probe only asks for the API root, which a server with
+		// broken storage still answers, so this read decides on its own.
+		setNotAvailable(controlplane.kcp, cpv1beta2.ControlPlaneNotAvailableReason, err.Error())
```
makhov: single ping failure will set cluster not available. I think we need to have some retries here or some logic like readiness/liveness probes have

internal/controller/controlplane/status.go
s3rj1k: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1693

It should have been this limit `MaxLength=32768`, and the idea was to just truncate to sane amount, comment is wrong here, got incorrect after automatic summarization and rewrite, thanks!

s3rj1k: on a second thought, I just might go with `MaxLength=32768`

makhov (approved):

makhov:
LGTM, but I'd like @apedriza to have a look before merging.

internal/controller/controlplane/inplace.go
```diff
@@ -45,14 +45,22 @@ import (
...
 	if !conditions.IsTrue(scope.kcp, cpv1beta2.ControlPlaneAvailableCondition) {
 		// If the control plane is not available, we cannot proceed with the in-place update, as access to the
 		// workload cluster is required to manage the autopilot plan.
+
+		// Falling through would replace machines an in place update should upgrade
+		// where they are. Only the rollout is held, never a count that does not match.
+		countMatches := scope.activeMachines.Len() == int(scope.kcp.Spec.Replicas)
+		if controlplaneRequiresUpdate && countMatches {
+			return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+		}
```
apedriza:
```suggestion
		desiredReplicasCount := scope.activeMachines.Len() == int(scope.kcp.Spec.Replicas)
		if controlplaneRequiresUpdate && desiredReplicasCount {
		    // Desired replica count reached, requeue waiting for controlplane availability
			return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
		}
```
Found the comment a bit confusing, do you find this block clearer?

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -138,6 +147,8 @@ func (c *K0sController) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 
 	if err := c.Get(ctx, req.NamespacedName, kcp); err != nil {
 		if apierrors.IsNotFound(err) {
+			// The count cannot be dropped here, since the key is the UID and this
+			// object was never read. reconcileDelete covers the ordinary path.
```
apedriza:
```suggestion
```
Seems it is missing a bit more context about what "the count" means. I guess you refer `availabilityFailures` but it's confusing at first glance. I would just remove it

internal/controller/controlplane/status.go
```diff
@@ -182,6 +185,136 @@ func versionMatches(machine *clusterv1.Machine, ver string) bool {
...
+	if state.LastProbeSuccessTime.IsZero() && state.ConsecutiveFailures < clusterCacheFailureThreshold {
+		return false
+	}
+
+	return time.Since(state.LastProbeSuccessTime) > clusterCacheConnectionGracePeriod
+}
+
+// availabilityFailureFloor is how many failed reads have to be seen by this process
+// before an outage is reported, whatever the persisted anchor says about the time.
+const availabilityFailureFloor = 2
```
apedriza:
```suggestion
const availabilityFailureThreshold = 2
```
for consistency with other limit consts

internal/controller/controlplane/status.go
```diff
@@ -191,6 +324,25 @@ func (c *K0sController) computeAvailability(ctx context.Context, controlplane *c
 	client, err := kutil.GetControllerRuntimeClient(ctx, c.Client, c.ClusterCache, controlplane.kcp, client.ObjectKeyFromObject(controlplane.cluster))
 	if err != nil {
 		logger.Info("Failed to get cluster client", "error", err)
+
+		// While coming up the contract fallback says this better than any
+		// internal error would.
+		if !initialized(controlplane.kcp.Status.Initialization.ControlPlaneInitialized) {
+			return
+		}
```
apedriza, line 353: I think we could just assume controlplane is not available at the beginning of the func if it is not initialized. Ideally, any try of workload connection should be done once we now it is initialized

internal/controller/controlplane/status.go
```diff
@@ -222,18 +388,53 @@ func (c *K0sController) computeAvailability(ctx context.Context, controlplane *c
...
+	}
+
 	if !conditions.IsTrue(kcp, string(cpv1beta2.ControlPlaneAvailableCondition)) {
-		return true
+		return availabilityRetryInterval
 	}
 
-	if *kcp.Status.UpToDateReplicas != kcp.Spec.Replicas {
-		return true
+	if ptr.Deref(kcp.Status.UpToDateReplicas, 0) != kcp.Spec.Replicas {
```
apedriza: requeue decision based on replica count is made in the scale logic. I wonder if we should move the current block here of focus this func to requeue based on Availability. If so, I would rename the func otherwise, I would centralize requeue based on replica status in one single point

internal/controller/controlplane/k0smotron_controlplane_controller.go
```diff
@@ -653,12 +666,15 @@ func (c *K0smotronController) computeAvailability(ctx context.Context, cluster *
...
-			Status:  metav1.ConditionFalse,
-			Reason:  "ClusterClientCreationFailed",
-			Message: err.Error(),
-		})
+
+		// While coming up the contract fallback says this better than any internal
+		// error would, and the condition defaults to Unknown at the epoch here.
+		if !initialized(kcp.Status.Initialization.ControlPlaneInitialized) {
+			return
+		}
```
apedriza: same as in the other controller, controlplane initialized should determine if we compute availability

apedriza (approved):
LGTM

We can keep iterating over availability logic. Probably we can inspect deeper the workload cluster to surface its state clearer and rely on this inspection for availability.

PR #1572, opened 2026-08-31
Pin the CAPI version in tests

makhov:
Pin the CAPI version to the same version as the clusterctl binary and go.mod.  FIxes test failures like https://github.com/k0sproject/k0smotron/actions/runs/33166903414/job/99404441053


apedriza (approved):
I think we should need to find another way to test using CAPI latest. Approved as I understand it unblocks a well known issue but it can be good to face this cases ealier using latest

PR #1580, opened 2026-09-02
Fix unexpected controlplane rollout when etcd image is upgraded

apedriza:
An upgrade in etcd image should not trigger an upgrade for the controlplane pods. This bug is because the hash used to decide if the k0smotron cluster requires an update includes the etcd image reference.

### DIFF
--- a/api/k0smotron.io/v1beta2/k0smotroncluster_types.go
+++ b/api/k0smotron.io/v1beta2/k0smotroncluster_types.go
@@ -99,6 +99,10 @@ const (
 
 	// NotFoundReason surfaces when a resource is not found.
 	NotFoundReason = "NotFound"
+
+	// AnnotationKeyClusterSpecHash is the annotation key used to store the hash of the desired cluster specification. This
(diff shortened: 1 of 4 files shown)


makhov:
Is it the right thing to do? I mean we restarting storage, so control plane should reconnect to a new one anyway. And this cause some time anyway as well, I don't think it's possible to do fully gracefully.

apedriza:
> Is it the right thing to do? I mean we restarting storage, so control plane should reconnect to a new one anyway. And this cause some time anyway as well, I don't think it's possible to do fully gracefully.

Well, I think both should be decoupled. Even while etcd is rolling out and pods need to reconnect, the apiserver can keep full API availability so it doesn't need to restart just because etcd's storage layer is being rolled out. And probably rolling out both statefulsets at the same time seems riskier than doing it separately. Also, things like defrag config changes are triggering rolling out controlplane but not for etcd 😮

makhov:
Yeah, the last one is definitely a bug

makhov (approved):

PR #1581, opened 2026-09-02
Reject etcd upgrades that skip a minor version in the Cluster webhook

### LINKED ISSUE #1579: validate etcd version upgrades

Upgrading an etcd cluster skipping minor version [is not supported by etcd](https://etcd.io/docs/v3.6/upgrades/upgrading-etcd/#:~:text=Minor%20upgrade%3A%20Upgrading%20one%20minor%20version%20at%20a%20time%20(e.g.%203.5%20%2D%203.6).%20Upgrades%20that%20skip%20a%20minor%20version%20are%20not%20supported%20and%20will%20likely%20fail.%20Update%20to%20the%20most%20recent%20patch%20version%20before%20upgrading%20to%20the%20next%20minor%20version.). Recently faced in #1578 

We should validate etcd version upgrades within our admission webhook

_No response_

### Additional context

_No response_

pujitha24:
Motivation:
etcd only supports upgrading one minor version at a time; skipping a
minor version (e.g. 3.5 straight to 3.7) leaves etcd unable to read
its own data on startup and crash-loops. This was observed after a
k0smotron-hosted control plane was upgraded across an etcd minor-version
gap, per the report below.

Approach:
Add a check to ClusterValidator.ValidateClusterSpecUpdate, in
api/k0smotron.io/v1beta2/k0smotroncluster_webhook.go, that runs
whenever Storage.Etcd.Image changes on an etcd-backed cluster. It
extracts the version tag from the old and new etcd image references
and, using the same github.com/k0sproject/version helper already used
for the analogous k0s minor-version-skew check in
K0sControlPlaneValidator, rejects the update if the new minor version
is more than one greater than the old one. If either image tag can't
be parsed as a version (custom image, non-semver tag), the check is
skipped rather than blocking the update, since Storage.Etcd.Image is a
free-form image reference. This is a validating-webhook check only: it
prevents users from applying a bad spec update via k0smotron, it does
not touch etcd or a running control plane.

K0smotronControlPlaneValidator.ValidateUpdate delegates to this same
function, so hosted control planes managed through K0smotronControlPlane
are covered as well as the Cluster CRD directly.

Validation:
- go build ./api/... passes.
- go test ./api/k0smotron.io/v1beta2/... passes, including new test
  TestValidateEtcdVersionUpgrade covering: same version, one-minor
  upgrade, patch upgrade, downgrade (not blocked), skipping a minor
  version (rejected), and unparsable/missing tags (skipped).
- make test (full unit test suite) passes.
- golangci-lint run --config .golangci.yml ./api/k0smotron.io/v1beta2/...
  reports 0 issues.
- No CRD/type files changed, so no codegen diff is expected from
  make generate-all.
- Not run: envtest/e2e/smoke tests requiring a live cluster; this
  change is pure admission-webhook logic validated at the unit level.

Report: https://github.com/k0sproject/k0smotron/issues/1579

Fixes #1579

### DIFF
--- a/api/k0smotron.io/v1beta2/k0smotroncluster_webhook.go
+++ b/api/k0smotron.io/v1beta2/k0smotroncluster_webhook.go
@@ -79,44 +79,48 @@ func (c ClusterValidator) ValidateClusterSpecUpdate(oldKCS, kcs *ClusterSpec) (w
 
 	if (kcs.Storage.Type == "" || kcs.Storage.Type == StorageTypeEtcd) &&
 		oldKCS.Storage.Etcd.Image != kcs.Storage.Etcd.Image {
-		if err := validateEtcdVersionUpgrade(oldKCS.Storage.Etcd.Image, kcs.Storage.Etcd.Image); err != nil {
+		warn, err := validateEtcdVersionUpgrade(oldKCS.Storage.Etcd.Image, kcs.Storage.Etcd.Image)
+		warnings = append(warnings, warn...)
+		if err != nil {
 			return warnings, err
 		}
 	}
 
-	return c.ValidateClusterSpec(kcs)
+	specWarnings, err := c.ValidateClusterSpec(kcs)
+	warnings = append(warnings, specWarnings...)
+	return warnings, err
 }
 
 // validateEtcdVersionUpgrade rejects etcd upgrades that skip a minor version. etcd only
 // supports upgrading one minor version at a time; skipping one is likely to leave etcd
 // unable to start on its existing data (see https://etcd.io/docs/v3.6/upgrades/upgrading-etcd/).
 // If either image's tag cannot be parsed as a version (e.g. a custom image or a non-semver
-// tag), the check is skipped rather than blocking the update.
-func validateEtcdVersionUpgrade(oldImage, newImage string) error {
-	oldV, err := etcdImageVersion(oldImage)
+// tag), the check is skipped and a warning is returned instead of blocking the update.
+func validateEtcdVersionUpgrade(oldImage, newImage string) (admission.Warnings, error) {
+	oldV, err := imageVersion(oldImage)
 	if err != nil {
-		return nil
+		return admission.Warnings{fmt.Sprintf("could not determine etcd version from image %q, skipping etcd upgrade path validation", oldImage)}, nil
 	}
-	newV, err := etcdImageVersion(newImage)
+	newV, err := imageVersion(newImage)
 	if err != nil {
-		return nil
+		return admission.Warnings{fmt.Sprintf("could not determine etcd version from image %q, skipping etcd upgrade path validation", newImage)}, nil
 	}
 
 	if newV.LessThanOrEqual(oldV) {
-		return nil
+		return nil, nil
 	}
 
 	if newV.Core().Segments()[1]-oldV.Core().Segments()[1] > 1 {
-		return fmt.Errorf("upgrading etcd image from %q to %q is not allowed: etcd does not support skipping a minor "+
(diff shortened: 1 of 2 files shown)


api/k0smotron.io/v1beta2/k0smotroncluster_webhook.go
```diff
@@ -76,9 +77,59 @@ func (c ClusterValidator) ValidateClusterSpecUpdate(oldKCS, kcs *ClusterSpec) (w
...
+		return fmt.Errorf("upgrading etcd image from %q to %q is not allowed: etcd does not support skipping a minor "+
+			"version when upgrading, please upgrade to an intermediate version first", oldImage, newImage)
+	}
+
+	return nil
+}
+
+// etcdImageVersion extracts the version from an etcd container image reference, e.g.
+// "quay.io/k0sproject/etcd:v3.7.1" returns the version "v3.7.1".
+func etcdImageVersion(image string) (*version.Version, error) {
```
makhov: Nothing in the func is etcd related, so it can use more general name

api/k0smotron.io/v1beta2/k0smotroncluster_webhook_test.go
```diff
@@ -58,3 +58,64 @@ func TestClusterValidator_validateVersionSuffix(t *testing.T) {
...
+			newImage:  "quay.io/k0sproject/etcd:v3.7.1",
+			wantError: true,
+		},
+		{
+			name:     "unparsable old tag skips the check",
+			oldImage: "quay.io/k0sproject/etcd:latest",
+			newImage: "quay.io/k0sproject/etcd:v3.7.1",
+		},
+		{
+			name:     "unparsable new tag skips the check",
```
makhov: Maybe it makes sense to add warnings on skip cases

api/k0smotron.io/v1beta2/k0smotroncluster_webhook.go
```diff
@@ -76,7 +77,61 @@ func (c ClusterValidator) ValidateClusterSpecUpdate(oldKCS, kcs *ClusterSpec) (w
...
 
-	return c.ValidateClusterSpec(kcs)
+	if (kcs.Storage.Type == "" || kcs.Storage.Type == StorageTypeEtcd) &&
+		oldKCS.Storage.Etcd.Image != kcs.Storage.Etcd.Image {
+		warn, err := validateEtcdVersionUpgrade(oldKCS.Storage.Etcd.Image, kcs.Storage.Etcd.Image)
+		warnings = append(warnings, warn...)
+		if err != nil {
+			return warnings, err
+		}
+	}
```
apedriza, line 87: this is part of the cluster spec so we can include this logic within `validateClusterSpec` func

pujitha24:
@makhov renamed `etcdImageVersion` to `imageVersion` and added warnings on the skip paths in 064b24f6 — thanks for both notes.

@apedriza `ValidateClusterSpec` only ever sees the new spec — it's also called from `ValidateCreate`, where there's no old spec to diff against — so this check (which needs to compare old vs new to catch a skipped minor version) can't move there as-is without changing its signature and touching all three call sites. Did you have a specific restructuring in mind, e.g. an optional old-spec param, or did you mean something else by "part of the cluster spec"?

apedriza:
> @apedriza `ValidateClusterSpec` only ever sees the new spec — it's also called from `ValidateCreate`, where there's no old spec to diff against — so this check (which needs to compare old vs new to catch a skipped minor version) can't move there as-is without changing its signature and touching all three call sites. Did you have a specific restructuring in mind, e.g. an optional old-spec param, or did you mean something else by "part of the cluster spec"?

Ok, then we can keep it as it is

apedriza (approved):
LGTM

PR #1583, opened 2026-09-02
Trust the cluster CA instead of skipping TLS verification in detectJoinHost

### LINKED ISSUE #1582: InsecureSkipVerify: true in detectJoinHost, confirm intentional?
### What happened?

`detectJoinHost` in `internal/controller/bootstrap/controlplane_bootstrap_controller.go` (line 674) builds an HTTP client with TLS certificate verification unconditionally disabled:

```go
transport.TLSClientConfig = &tls.Config{
    // Since we are using self-signed certificates, we need to skip the verification
    InsecureSkipVerify: true,
}
```

This client probes `https://<control-plane-endpoint>:<port>/v1beta1/ca` on a newly-provisioned machine to detect the join host, falling back to the first controller's IP if that probe fails. There is no opt-in flag gating this, it always skips verification.

The comment explains the immediate reason: at bootstrap time the machine's certificate is self-signed and there is no CA yet to validate against. That is a real constraint, not obviously a mistake. What is unclear is whether the network path this probe travels, before cluster trust is established, is considered safe by design, or whether this is an accepted gap that should eventually be closed with something like TOFU style fingerprint pinning.

### Involved Areas

Cluster API integration

Looking for maintainer confirmation on one of two things. Either this is a known, accepted tradeoff given the network model CAPI infra providers set up around these machines, in which case this can be closed as working as intended, or it is worth tracking as a hardening item if that network path could be reachable by something untrusted during the join window.

### Steps to reproduce

This is a static code observation, not a runtime reproduction.
1. Open `internal/controller/bootstrap/controlplane_bootstrap_controller.go`.
2. Look at `detectJoinHost` (line 674).
3. Note `InsecureSkipVerify: true` is unconditional, with no flag to opt out.

### k0smotron version

N/A, code level observation against current main, not tied to a specific release.

### k0s version

N/A, not specific to a k0s version, the affected code is entirely within k0smotron's own bootstrap controller.

Not attempting a fix here since resolving it properly needs infra and network model context this repo alone does not capture. Happy to help implement whatever direction maintainers confirm, for example TOFU pinning, if useful.

pujitha24:
Motivation:
detectJoinHost (internal/controller/bootstrap/controlplane_bootstrap_controller.go)
probes a newly-provisioned control-plane machine's k0s join API
(https://<control-plane-endpoint>:<port>/v1beta1/ca) to pick the host used to
build the join token, falling back to the first controller's IP if the probe
fails. The HTTP client set InsecureSkipVerify: true unconditionally, with no
way to opt out, so the probe never actually validated the server it was
talking to.

detectJoinHost is only called from genControlPlaneJoinFiles, which is used
when joining additional control-plane nodes to an already-initialized
cluster. By that point the function has already fetched the cluster CA via
getCerts (the same CA k0smotron writes to /var/lib/k0s/pki on every node),
and unconditionally dereferences ca.KeyPair.Cert a few lines later to build
the join token. k0s signs the join API's serving certificate (k0s-api) with
this same cluster CA, so it is the correct trust anchor for this probe.

This is a hardening change, not a fix for a demonstrated exploit: the probe
runs from the management cluster to a freshly-provisioned node, typically
over the private network CAPI infrastructure providers set up for
bootstrapping, which already limits who can sit on that path. Skipping
verification still meant the probe could not detect a mismatched or spoofed
endpoint, which is what this closes.

Approach:
- detectJoinHost now takes the cluster CA (*secret.Certificate) and builds
  an x509.CertPool from it, using RootCAs instead of InsecureSkipVerify.
  If the CA is missing or unparsable, it returns an error immediately
  instead of silently trusting an unverified endpoint. This does not
  introduce a new panic risk: the existing code already dereferences
  ca.KeyPair.Cert unconditionally later in the same call chain, so a nil
  KeyPair was already fatal to this code path before this change.
- The single call site in genControlPlaneJoinFiles passes the ca value it
  already had in scope.

Validation:
- go build ./... and go vet ./internal/controller/bootstrap/... pass.
- gofmt -l reports no formatting issues.
- golangci-lint run --config .golangci.yml ./internal/controller/bootstrap/...
  reports 0 issues.
- make test and make envtest (envtest run with real etcd/kube-apiserver
  binaries) both pass across the repo. Note: make envtest's bootstrap
  controller scenarios only exercise the single-machine
  genInitialControlPlaneFiles path, not genControlPlaneJoinFiles, so this
  does not directly exercise detectJoinHost; it only confirms no regression
  elsewhere in the package.
- Added TestControlPlaneController_detectJoinHost, a targeted unit test
  using an httptest.NewTLSServer: a CA pool containing the server's own
  certificate succeeds and returns the control-plane endpoint; a CA pool
  containing an unrelated certificate causes the probe to fail closed and
  fall back to the first controller's address; a nil CA returns an error.
  I confirmed this test is a genuine regression check by temporarily
  reverting just the TLS behavior back to InsecureSkipVerify (keeping the
  new function signature): the "falls back" subtest failed as expected
  (the probe wrongly succeeded against the wrong CA), then passed again
  once the fix was restored.

Report: https://github.com/k0sproject/k0smotron/issues/1582

Fixes #1582

### DIFF
--- a/internal/controller/bootstrap/controlplane_bootstrap_controller.go
+++ b/internal/controller/bootstrap/controlplane_bootstrap_controller.go
@@ -19,6 +19,7 @@ package bootstrap
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"net/http"
@@ -464,7 +465,7 @@ func (c *ControlPlaneController) genControlPlaneJoinFiles(ctx context.Context, s
 		return nil, err
 	}
 
-	host, err := c.detectJoinHost(ctx, scope, firstControllerMachine)
+	host, err := c.detectJoinHost(ctx, scope, firstControllerMachine, ca)
 	if err != nil {
 		log.Error(err, "Failed to detect join controller host")
 		return nil, err
@@ -671,11 +672,17 @@ func mergeControllerExtraArgs(scope *ControllerScope) []string {
 	return mergeExtraArgs(scope.installArgs, scope.ConfigOwner, scope.WorkerEnabled, scope.Config.Spec.UseSystemHostname)
 }
 
-func (c *ControlPlaneController) detectJoinHost(ctx context.Context, scope *ControllerScope, firstControllerMachine *clusterv1.Machine) (string, error) {
+func (c *ControlPlaneController) detectJoinHost(ctx context.Context, scope *ControllerScope, firstControllerMachine *clusterv1.Machine, ca *secret.Certificate) (string, error) {
+	caCertPool := x509.NewCertPool()
+	if ca == nil || ca.KeyPair == nil || !caCertPool.AppendCertsFromPEM(ca.KeyPair.Cert) {
+		return "", errors.New("failed to load cluster CA certificate for join host detection")
+	}
+
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 	transport.TLSClientConfig = &tls.Config{
-		// Since we are using self-signed certificates, we need to skip the verification
-		InsecureSkipVerify: true,
+		// The join API serves a certificate signed by the cluster CA, so we
+		// trust it explicitly instead of skipping verification.
+		RootCAs: caCertPool,
 	}
 	httpClient := &http.Client{
 		Transport: transport,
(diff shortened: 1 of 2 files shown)


apedriza (approved):
LGTM

PR #1586, opened 2026-09-04
Probe Cluster API v1beta2 to gate CAPI controllers, not v1beta1

### LINKED ISSUE #1585: Still referencing ClusterAPI resources when running in standalone mode
### What happened?

I am installing **v2.0.4** in standalone mode via: https://docs.k0smotron.io/v2.0.4/install-standalone.yaml. `controller-manager` is in CrashLoop Backoff with Error message: 

```2026-09-03T21:09:42Z    ERROR   setup   problem running manager {"error": "failed to wait for clustercache caches to sync kind source: *v1beta2.Cluster: timed out waiting for cache to be synced for Kind *v1beta2.Cluster"}```

My cluster had some outdated ClusterAPI CRDs (v1beta1) installed. However, I don't thing this should impact when running in standalone mode

This was supposedly fixed in **v2.0.1**: https://github.com/k0sproject/k0smotron/issues/1486, so re-reporting

### Involved Areas

_No response_

Cluster runs fully in standalone mode

### Steps to reproduce

1. Install v1beta1 ClusterAPI CRDs in cluster
2. Install k0smotron in standalone mode via https://docs.k0smotron.io/v2.0.4/install-standalone.yaml

### k0smotron version

v2.0.4

### k0s version

NA

_No response_

pujitha24:
Motivation:
cmd/main.go decides whether to enable the CAPI controllers (bootstrap,
control-plane, infrastructure) and start the clustercache by probing
via discovery whether "cluster.x-k8s.io/v1beta1" resources are served.
That probe predates the switch of this manager's clusterv1 import to
"sigs.k8s.io/cluster-api/api/core/v1beta2" and was never updated: every
controller, the scheme registration, and clustercache.SetupWithManager
now use v1beta2 types exclusively.

When a cluster has stale/partial Cluster API remnants that still serve
v1beta1 but not v1beta2 for the "cluster.x-k8s.io" group (e.g. an old
CRD installed by another tool, with no CAPI actually running), the
v1beta1 probe returns true, the CAPI controllers and clustercache start,
clustercache tries to watch "*v1beta2.Cluster", the API server doesn't
serve that kind, and the cache never syncs. The manager then crash-loops
with "failed to wait for clustercache caches to sync kind source:
*v1beta2.Cluster: timed out waiting for cache to be synced". This is
not limited to standalone installs; any install hitting that CRD state
hits the same crash loop, but standalone users (who never intend to run
CAPI controllers at all) are the ones most likely to have only such
stray CRDs and no working CAPI otherwise.

Approach:
Change the discovery probe to check "cluster.x-k8s.io/v1beta2" instead
of "v1beta1", matching the API version the manager's scheme and
clustercache actually watch. Extract the probe into a small
"shouldRunCAPIControllers" helper so it is unit-testable, and add a
table-driven test in cmd/main_test.go using a fake discovery client
covering: no CAPI resources installed, only-v1beta1 installed
(reproduces the reported crash-loop precondition), and v1beta2
installed.

Cluster API v1.12.x (this repo's sigs.k8s.io/cluster-api dependency)
serves both v1beta1 (via conversion, not storage) and v1beta2 for the
core Cluster CRD, so any properly installed current CAPI environment
continues to pass the new probe unaffected.

Validation:
- go build ./... and make build (the CI build target) both pass.
- make test (repo's CI unittest command: go test ./api/... ./cmd/...
  ./internal/... -run '' -coverprofile cover.out) passes for all
  packages, including the new cmd/main_test.go table:
  TestShouldRunCAPIControllers/only_outdated_v1beta1_cluster-api_CRDs_installed
  fails against the old v1beta1 probe (that function did not exist
  before this change) and passes after this change, directly
  reproducing and proving the fix for the reported precondition.
- golangci-lint run --config .golangci.yml ./cmd/... reports 0 issues.
- Not run: envtest and the e2e/smoke-test suites, since they require a
  live cluster/docker and do not exercise cmd/main.go's discovery
  gating logic; the change is fully covered by the new unit test.

Report: https://github.com/k0sproject/k0smotron/issues/1585

Fixes #1585

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/go-logr/logr"
 	"github.com/spf13/pflag"
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -253,19 +254,13 @@ func main() {
 	}
 
 	// Absent CAPI Machine kind means we are not running in a CAPI environment, so we don't need to run CAPI controllers.
-	var runCAPIControllers bool
 	discoveryClient, err := discovery.NewDiscoveryClientForConfig(restConfig)
 	if err != nil {
 		setupLog.Error(err, "unable to create discovery client")
 		os.Exit(1)
 	}
 
-	resources, err := discoveryClient.ServerResourcesForGroupVersion(schema.GroupVersion{Group: "cluster.x-k8s.io", Version: "v1beta1"}.String())
-	if err == nil && len(resources.APIResources) > 0 {
-		runCAPIControllers = true
-	} else {
-		mgr.GetLogger().Info("Cluster API v1beta1 not installed, skipping cluster-api controllers setup")
-	}
+	runCAPIControllers := shouldRunCAPIControllers(discoveryClient, mgr.GetLogger())
 
 	//+kubebuilder:scaffold:builder
 	ctrlOptions := capictrl.Options{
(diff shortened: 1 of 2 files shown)


apedriza:
Unfortunately, this solution does not solve the root issue, which is enabling controllers (in this case, the clustercache CAPI controller) for standalone that should be running. I will take care of the fix, as it probably requires design decisions about what we expect depending on the way k0smotron is installed and what was installed previously (CAPI controllers). 

Thanks anyways @pujitha24

pujitha24:
Fair — the probe fix stops the specific v1beta1/v1beta2 mismatch crash-loop, but you're right that it doesn't touch the bigger question of whether CAPI controllers should be running at all for a given install type. That's more of a design call than what I scoped here, so it makes sense for you to take it. Thanks for looking at it.

PR #1587, opened 2026-09-04
Add Pull Request template

apedriza:
Automated AI PRs are comming more a more. Besides add a proper pull request template this PR inteds to "force" or help to keep ai-generated PR human friendly.


makhov:
I'm not sure if it's even possible, but I would love to see something like "this PR fixes the root cause, not just adds a workaround"

ncopa (approved):

PR #1592, opened 2026-09-07
Fix k0smotron manager setup based on prior cluster env

apedriza:
## Description

Taking into account that k0smotron enables certain controllers based on the k0smotron deployment flags and also checking if CAPI is installed before, we can improve the logic that checks if CAPI is installed without expect CAPI resources using a determined version and just try to discover CAPI "cluster.x-k8s.io" API group is present.

Also, fix the explicit installation of standlaone k0smotron when CAPI is previously installed by not running CAPI clustercache controller.

## Checklist

- [ ] The PR title clearly describes the change
- [ ] The description explains the *why*, not the *what* (that's already visible in the code)
- [ ] I have tested the changes locally / on a test cluster
- [ ] The changes do not break backward compatibility (or it's called out and justified if they do)
- [ ] I have checked that no credentials, tokens, or sensitive data are included

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"crypto/md5"
 	"crypto/tls"
 	"flag"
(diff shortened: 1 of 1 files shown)


apedriza:
duplicated

PR #1594, opened 2026-09-09
Add tests for ReconcileDynamicConfig and RandomString

magic-peach:
## Description

internal/util had zero test coverage across all three of its files. ReconcileDynamicConfig in particular strips storage and network.controlPlaneLoadBalancing from the incoming config before building the merge patch, since those are meant to stay under local k0s config control, not be pushed centrally. That's exactly the kind of behavior that's easy to break silently in a later refactor without a test catching it, since a merge patch simply omitting a field means "leave it alone," not "delete it" - a distinction worth pinning down explicitly. Also added coverage for RandomString's length and charset guarantees and for ReconcileDynamicConfig's patch error path.

## Checklist

## Related issues

None - self-directed test coverage improvement, no existing issue.


internal/util/random_test.go
```diff
@@ -0,0 +1,35 @@
...
+//nolint:revive
+package util
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRandomString(t *testing.T) {
```
makhov: This test tests nothing, we don't need it

internal/util/dynamic_config_test.go
```diff
@@ -0,0 +1,111 @@
...
+	cplbEnabled, _, err := unstructured.NestedBool(got.Object, "spec", "network", "controlPlaneLoadBalancing", "enabled")
+	require.NoError(t, err)
+	require.True(t, cplbEnabled, "controlPlaneLoadBalancing should be left untouched, not overwritten with the incoming value")
+
+	podCIDR, _, err := unstructured.NestedString(got.Object, "spec", "network", "podCIDR")
+	require.NoError(t, err)
+	require.Equal(t, "10.245.0.0/16", podCIDR, "fields other than the stripped ones should still be applied")
+}
+
+func TestReconcileDynamicConfigPatchError(t *testing.T) {
```
makhov: This test tests nothing

makhov (changes requested):

magic-peach:
@makhov I made the changes, please have a look when you get time

PR #1595, opened 2026-09-09
Assorted unrelated fixes across the bootstrap and control-plane providers

s3rj1k:
Assorted unrelated fixes across the bootstrap and control-plane providers:

- A tunneled control plane no longer reports itself unavailable when it is spelled `--enable-worker=true` or `--single`.
- Waiting for the first controller machine no longer reads as success, which used to publish bootstrap data with a null value that was never regenerated.
- The join target is picked with a real ordering, so the controller a machine joins is no longer decided by map iteration order.
- An Ignition file with no permissions falls back to 0644 instead of failing the whole render, which blocked ingress-enabled workers.
- The pending `UpdateMachine` hook is set after the objects it applies to, as the control-plane contract requires.
- `RemediationFailed` is a declared constant instead of an inline string literal.
- The v1beta2 `downloadURL` note names `preK0sCommands`, which is the field that actually exists in v1beta2.
- The tunneling guide names the kubeconfig secrets the controller actually creates.
- The generated haproxy files carry explicit modes, so a TLS private key no longer lands world-readable through a renderer's fallback.
- A failed bootstrap secret write reports `DataSecretAvailable` as false rather than true.


s3rj1k:
Split into PRs

https://github.com/k0sproject/k0smotron/pull/1595
https://github.com/k0sproject/k0smotron/pull/1600
https://github.com/k0sproject/k0smotron/pull/1601
https://github.com/k0sproject/k0smotron/pull/1602
https://github.com/k0sproject/k0smotron/pull/1603
https://github.com/k0sproject/k0smotron/pull/1604
https://github.com/k0sproject/k0smotron/pull/1605
https://github.com/k0sproject/k0smotron/pull/1606
https://github.com/k0sproject/k0smotron/pull/1607
https://github.com/k0sproject/k0smotron/pull/1608

PR #1597, opened 2026-09-10
fix: preserve pooled machine cleanup commands on deletion (#1596)

### LINKED ISSUE #1596: RemoteMachineTemplate/pooled RemoteMachine: customCleanUpCommands set on PooledRemoteMachine is not applied to the claimed machine
### What happened?

When a `RemoteMachine` is claimed from a pool through a `MachineDeployment` or `KubeadmControlPlane`, the `customCleanUpCommands` configured on the `PooledRemoteMachine` are not available during deletion.

The controller logs:

```text
k0smotron is not the bootstrap provider and no cleanup commands specified, skipping cleanup
```

No cleanup is executed on the host. As a result, kubelet remains running and files such as `/etc/kubernetes/kubelet.conf`, `/etc/kubernetes/pki`, and CNI state are left behind.

When the same host is reclaimed after scaling back up, `kubeadm join` repeatedly fails during preflight checks because the previous cluster state was not removed. Manual execution of `kubeadm reset` is required to recover the host.

There also appears to be inconsistent propagation of pool values. The claimed `RemoteMachine` spec shows `user: root` and no `sshKeyRef`, although the SSH connection succeeds using the actual user and key configured on the pool entry. This suggests that some pool values are used at connection time without being persisted on the claimed object, while `customCleanUpCommands` is not propagated at all.

The cleanup commands cannot be configured on `RemoteMachineTemplate`, because `spec.template.spec` only supports `pool` and `provisionJob`. Therefore, configuring cleanup commands directly on the template is not a workaround.

### Involved areas

Cluster API integration, pooled `RemoteMachine` lifecycle, and cleanup handling.

When a `RemoteMachine` is claimed from a `PooledRemoteMachine`, all relevant values from the pool entry should be applied to the claimed machine, including `customCleanUpCommands`.

When the machine is deleted, those commands should run on the host so it is clean and reusable.

This is also consistent with the documentation in [`[docs/capi-remote.md](https://github.com/k0sproject/k0smotron/blob/main/docs/capi-remote.md)`](https://github.com/k0sproject/k0smotron/blob/main/docs/capi-remote.md):

> When CAPI controller creates a `RemoteMachine` from template object, k0smotron will pick one of the `PooledRemoteMachine` objects and use its values for the `RemoteMachine` object.

### Steps to reproduce

1. Create a pooled remote machine and a remote machine template:

```yaml
kind: PooledRemoteMachine
metadata:
  name: worker-host-1
spec:
  pool: workers
  machine:
    address: <IP>
    user: capi
    useSudo: true
    customCleanUpCommands:
      - kubeadm reset -f
      - rm -rf /etc/kubernetes /var/lib/kubelet /run/kubeadm /etc/cni/net.d
    sshKeyRef:
      name: remotemachine-key
---
kind: RemoteMachineTemplate
metadata:
  name: worker-template
spec:
  template:
    spec:
      pool: workers
```

2. Apply a `MachineDeployment` that references `worker-template` with `replicas: 1`.

   The machine provisions and joins successfully using CABPK/kubeadm as the bootstrap provider, rather than `K0sWorkerConfig`.

3. Scale the `MachineDeployment` to `replicas: 0`.

   The `Machine` and `RemoteMachine` are deleted, and `PooledRemoteMachine.status.reserved` changes to `false`. No cleanup command runs on the host. The kubelet remains active and `/etc/kubernetes` is still populated.

4. Scale the `MachineDeployment` back to `replicas: 1`.

   The same host is reclaimed, but `kubeadm join` repeatedly fails during preflight checks because state from the previous cluster remains. The host must be reset manually with `kubeadm reset`.

### k0smotron version

`v2.1.1`

### k0s version

N/A. This cluster uses CABPK/kubeadm as the bootstrap provider, not k0s.

- Cluster API version: `v1.14.2`.
- Installed using `clusterctl init --infrastructure k0sproject-k0smotron`.
- The installation includes CAPI's default `bootstrap-kubeadm` and `control-plane-kubeadm` providers alongside k0smotron's infrastructure provider.
- This does not use `K0sWorkerConfig` or

david6983:
## Description

When CAPI claims a `RemoteMachine` from a `PooledRemoteMachine`, the claimed machine inherits the pool’s connection and cleanup configuration. During deletion, the controller skipped this synchronization because the machine had a deletion timestamp. As a result, cleanup commands configured on the pool were not available to non-k0s bootstrap providers, leaving hosts dirty and causing kubeadm preflight failures when those hosts were reclaimed.

This change keeps the claimed machine synchronized with its pool throughout its lifecycle, including deletion, and reconciles claimed machines when their pool entry changes. Regression tests cover cleanup command propagation and pool-to-machine reconciliation.

## Checklist

- [x] The description explains the *why*, not just the *what*
- [x] I have added or updated tests covering the change
- [ ] I have updated relevant documentation
- [x] The changes do not break backward compatibility

## Related issues

Closes #1596

### DIFF
--- a/internal/controller/infrastructure/remote_machine_controller.go
+++ b/internal/controller/infrastructure/remote_machine_controller.go
@@ -26,6 +26,7 @@ import (
 	"gopkg.in/yaml.v3"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/retry"
@@ -44,7 +45,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 // ErrPooledMachineNotFound is returned when a RemoteMachine references a pool
@@ -178,9 +181,9 @@ func (r *RemoteMachineController) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, fmt.Errorf("failed to parse bootstrap data: %w", err)
 	}
 
-	if rm.Spec.Pool != "" && rm.ObjectMeta.DeletionTimestamp.IsZero() {
+	if rm.Spec.Pool != "" {
 		err := r.reservePooledMachineAndPopulateRemoteMachine(ctx, rm)
-		if err != nil {
+		if err != nil && rm.ObjectMeta.DeletionTimestamp.IsZero() {
 			log.Error(err, "Error reserving PooledMachine")
 			return ctrl.Result{Requeue: true}, err
 		}
@@ -331,6 +334,18 @@ func (r *RemoteMachineController) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	return ctrl.Result{}, nil
+}
+
+func pooledRemoteMachineToRemoteMachine(_ context.Context, obj client.Object) []reconcile.Request {
+	pooledMachine, ok := obj.(*infrastructure.PooledRemoteMachine)
+	if !ok || !pooledMachine.Status.Reserved || pooledMachine.Status.MachineRef.Name == "" {
+		return nil
+	}
+
+	return []reconcile.Request{{NamespacedName: types.NamespacedName{
+		Name:      pooledMachine.Status.MachineRef.Name,
+		Namespace: pooledMachine.Status.MachineRef.Namespace,
+	}}}
 }
 
 // mergedMap copies src over dst and allocates dst when it is nil, since copying
@@ -525,5 +540,6 @@ func (r *RemoteMachineController) SetupWithManager(mgr ctrl.Manager, opts contro
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(opts).
 		For(&infrastructure.RemoteMachine{}).
+		Watches(&infrastructure.PooledRemoteMachine{}, handler.EnqueueRequestsFromMapFunc(pooledRemoteMachineToRemoteMachine)).
 		Complete(r)
 }
--- a/internal/controller/infrastructure/remote_machine_controller_test.go
+++ b/internal/controller/infrastructure/remote_machine_controller_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	infrastructure "github.com/k0sproject/k0smotron/v2/api/infrastructure/v1beta2"
@@ -95,6 +96,10 @@ func TestReservePooledMachineCopiesMetadataOntoBareRemoteMachine(t *testing.T) {
 				Address: "10.0.0.1",
 				Port:    22,
 				User:    "root",
+				CleanUpCommands: []string{
+					"kubeadm reset -f",
+					"rm -rf /etc/kubernetes",
+				},
 			},
 		},
 	}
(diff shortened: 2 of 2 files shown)


apedriza:
Waiting clarification for https://github.com/k0sproject/k0smotron/issues/1596#issuecomment-5617559680

internal/controller/infrastructure/remote_machine_controller.go
```diff
@@ -178,9 +181,9 @@ func (r *RemoteMachineController) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, fmt.Errorf("failed to parse bootstrap data: %w", err)
 	}
 
-	if rm.Spec.Pool != "" && rm.ObjectMeta.DeletionTimestamp.IsZero() {
+	if rm.Spec.Pool != "" {
 		err := r.reservePooledMachineAndPopulateRemoteMachine(ctx, rm)
-		if err != nil {
+		if err != nil && rm.ObjectMeta.DeletionTimestamp.IsZero() {
```
apedriza, line 186: now that we watch poolremotemachine, I would add a reconcile method for it and within it encapsulate reservation and spec propagation. Now that `reservePooledMachineAndPopulateRemoteMachine` can be hit due to a spec change in poolremotemachine I think the name does not have sense anymore.

On the other hand, should we just return err if this reconciliation I mention above fails instead check deletiontimestamp?

internal/controller/infrastructure/remote_machine_controller.go
```diff
@@ -525,5 +540,6 @@ func (r *RemoteMachineController) SetupWithManager(mgr ctrl.Manager, opts contro
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(opts).
 		For(&infrastructure.RemoteMachine{}).
+		Watches(&infrastructure.PooledRemoteMachine{}, handler.EnqueueRequestsFromMapFunc(pooledRemoteMachineToRemoteMachine)).
```
apedriza, line 543: This is good but also have implications. The user will reconcile new value from the poolremotemachine into the remotemachine so we should either document that changing address port key.. etc might cause issues when trying to clean up the machine

internal/controller/infrastructure/remote_machine_controller.go
david6983: Yes I agree that now the pool watch can trigger this logic after a `PooledRemoteMachine` spec change. How about we rename the helper like `reservePooledMachineAndPopulateRemoteMachine` and we treat it as reconciliation rather than only initial reservation.

For error handling, I’m considering returning reconciliation errors normally instead of ignoring them during deletion. The one case thing I’m not sure about is when the pool entry has already been deleted: returning the error could leave the `RemoteMachine` finalizer stuck, while the persisted `RemoteMachine` spec may still contain enough information to perform cleanup. Would it make sense to return all errors except `ErrPooledMachineNotFound` during deletion?

And yes the changing address, port, user, or SSH key on a claimed pool entry can affect cleanup. We can add a documentation warning for that behavior, what do you think?

internal/controller/infrastructure/remote_machine_controller.go
david6983: Oh yes you are right. if I understand correctly watching `PooledRemoteMachine` changes means updates to the address, port, user, SSH key, or other connection fields will also be propagated to the claimed `RemoteMachine`. And this could affect cleanup if the machine is already in use.

Maybe we can add a warning to the CAPI remote-machine documentation explaining this and recommending that connection settings are not changed while a pool entry is claimed?

apedriza: Yes! I think this should be the proper placer to locate that warning https://docs.k0smotron.io/stable/capi-remote/#using-remotemachines-in-machinetemplates-of-higher-level-objects

internal/controller/infrastructure/remote_machine_controller.go
apedriza: > How about we rename the helper like reservePooledMachineAndPopulateRemoteMachine and we treat it as reconciliation rather than only initial reservation.

Lets create a new `reconcileFromPool` or something similar when the remotemachine belongs to a pool, wdyt?

>  Would it make sense to return all errors except ErrPooledMachineNotFound during deletion?

Yes. Let's not block the deletion process, instead let's log an error if the pool is not found

apedriza:
@david6983 Great discussion! Would you take care of the changes we talked about?

david6983:
> @david6983 Great discussion! Would you take care of the changes we talked about?

Yes working on it

internal/controller/infrastructure/remote_machine_controller.go
david6983: Noted, I will update the documentation

david6983:
@apedriza just pushed the new changes, looking forward for your review

david6983:
@apedriza I implemented the snapshot behavior by removing the PooledRemoteMachine watch and using the values persisted on RemoteMachine during cleanup. I also added a validating webhook that rejects spec changes while a pool entry is reserved, while still allowing metadata and status updates. 

Is this the intended scope, or would you prefer the webhook to be handled separately?

apedriza (changes requested):
this is looking good! We still need to include entries for the`ValidatingWebhookConfiguration` for the `clusterapi/all/webhook/manifests.yaml` file. This file is used for the installation where all the controllers are enabled.

You can generate it just executing `make generate-all`

david6983:
> this is looking good! We still need to include entries for the`ValidatingWebhookConfiguration` for the `clusterapi/all/webhook/manifests.yaml` file. This file is used for the installation where all the controllers are enabled.
> 
> You can generate it just executing `make generate-all`

@apedriza done 👍

apedriza:
can you squash your commits and sign the result commit? Also, the linter is failing in CI. Can you have a look?

david6983:
> can you squash your commits and sign the result commit? Also, the linter is failing in CI. Can you have a look?

@apedriza all good, just need to approve workflows and I will double check if it pass

apedriza (approved):
LGTM! thanks @david6983

PR #1602, opened 2026-09-10
Pick the join target with the shared oldest first ordering

s3rj1k:
The comparator required both a lower version and an earlier timestamp, which is not an ordering, so machines of the same version kept the random order of the map they came from. The version key served no purpose, so it goes and Filter and Oldest from the collections package do the work.


apedriza (approved):
lgtm

PR #1604, opened 2026-09-10
Ensure an in place update trigger completes before another starts

s3rj1k:
The Machine was patched with the pending UpdateMachine hook before the objects it applies to were marked, so a restart in between left the version bumped and the hook stranded. The latch now goes on first and the hook last, and a Machine carrying the latch without the hook has its trigger finished before a new one starts.

### DIFF
--- a/internal/controller/controlplane/inplace.go
+++ b/internal/controller/controlplane/inplace.go
@@ -345,21 +345,9 @@ func (c *K0sController) updateMachineVersion(ctx context.Context, machine *clust
 	return nil
 }
 
+// triggerCAPIInplaceVersionUpdate marks a Machine and the objects it points at as being
+// updated in place. The pending hook has to come after the other two are marked.
 func triggerCAPIInplaceVersionUpdate(ctx context.Context, c client.Client, desiredVersion string, desiredMachine *clusterv1.Machine, desiredInfraMachine *unstructured.Unstructured, desiredBootstrapConfig *bootstrapv2.K0sControllerConfig) error {
-	if _, ok := desiredMachine.Annotations[clusterv1.UpdateInProgressAnnotation]; !ok {
-		orig := desiredMachine.DeepCopy()
-		desiredMachine.Spec.Version = desiredVersion
-		if desiredMachine.Annotations == nil {
-			desiredMachine.Annotations = map[string]string{}
-		}
-		desiredMachine.Annotations[clusterv1.UpdateInProgressAnnotation] = ""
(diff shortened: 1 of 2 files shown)


apedriza:
I'm not sure if this should be the way to handle incomplete triggers. Following CAPI proposal notes and its code they follow the same order we are currently using. What they have (see [reconciler](https://github.com/kubernetes-sigs/cluster-api/blob/main/controlplane/kubeadm/reconcilers/kubeadmcontrolplane/kubeadmcontrolplane_controller.go#L507-L517))  to cover this potential cases is that they re-trigger if they found the annotation but no hook pending. wdyt?

s3rj1k:
> I'm not sure if this should be the way to handle incomplete triggers. Following CAPI proposal notes and its code they follow the same order we are currently using. What they have (see [reconciler](https://github.com/kubernetes-sigs/cluster-api/blob/main/controlplane/kubeadm/reconcilers/kubeadmcontrolplane/kubeadmcontrolplane_controller.go#L507-L517)) to cover this potential cases is that they re-trigger if they found the annotation but no hook pending. wdyt?

something like this https://github.com/k0sproject/k0smotron/pull/1604/changes/a71c6bfaa76180f3bf826ba6c71d367866c7baa7 ?

internal/controller/controlplane/inplace.go
```diff
@@ -47,6 +49,16 @@ import (
 func (c *K0sController) reconcileInplaceK0sVersionUpdate(ctx context.Context, scope *controlplane) (ctrl.Result, error) {
 	controlplaneRequiresUpdate := scope.hasMachinesWithOnlyVersionOutdated && scope.kcp.Spec.UpdateStrategy == cpv1beta2.UpdateInPlace
 
+	// Ahead of every gate below, since a Machine whose trigger stopped short of the hook is not
+	// outdated in any way those gates look at, so they would return before reaching it.
+	triggerCompleted, err := c.completeInterruptedTriggers(ctx, scope, log.FromContext(ctx))
```
apedriza:
```suggestion
	letInterruptedToComplete, err := c.completeInterruptedTriggers(ctx, scope, log.FromContext(ctx))
```
otherwise seems that we only continue when the trigger is not completed

s3rj1k: done, amended into 2nd commit

apedriza: I will squash both as it seems one overwrites the other. Can you give the PR a descriptive name? like `Ensure trigger completed...`?

s3rj1k: done

apedriza (approved):
lgtm

PR #1610, opened 2026-09-10
Stop PowerShell file paths and content escaping into script

s3rj1k:
Paths went into a double quoted string and escapePS escaped only the quote character, so a path like
C:\k\$(ni PWNED).conf ran the subexpression on the target. They are single quoted now, which expands
nothing.

Content went into an @' here string, which is script text, so a body line holding the terminator
closed it and everything after ran as PowerShell. All content now takes the base64 and WriteAllBytes
path that append and non UTF8 bodies already used, at roughly 1.33x user data size.

### DIFF
--- a/internal/provisioner/powershell.go
+++ b/internal/provisioner/powershell.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
-	"unicode/utf8"
 )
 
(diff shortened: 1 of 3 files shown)


makhov (approved):
TestWindowsProvisioner passed on this PR: https://github.com/k0sproject/k0smotron/actions/runs/34566720137

PR #1611, opened 2026-09-10
Send Ignition file content as a data URL so YAML cannot mangle it

s3rj1k:
Content went out as a Butane inline scalar, which is YAML, so a leading newline was dropped, a body of just a newline produced an empty file, and a leading tab failed the whole render. A data URL holds any byte, and the append branch already used one.

### DIFF
--- a/internal/provisioner/ignition.go
+++ b/internal/provisioner/ignition.go
@@ -72,22 +72,23 @@ func (i *IgnitionProvisioner) ToProvisionData(input *InputProvisionData) ([]byte
 			"mode": int(mi),
 		}
 
-		// An Ignition file entry carries either contents or append, never both.
-		// Append is new, so it can use a data URL, which holds any byte.
+		// A data URL is opaque to the YAML emitter, which cannot carry a leading
+		// newline in a block scalar and emits a tab that Butane then fails to parse.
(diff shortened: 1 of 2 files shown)


internal/provisioner/ignition_test.go
```diff
@@ -485,3 +485,46 @@ func TestIgnitionRejectsUndecodableContent(t *testing.T) {
...
+			require.Equal(t, tc.content, string(bodyBytes(t, entry["contents"].(map[string]any))))
+		})
+	}
+}
+
+// TestIgnitionPreservesContentAcrossSpecVersions covers the same bodies on every
+// variant and version the CRD accepts, since the old scalar failed on all of them.
+func TestIgnitionPreservesContentAcrossSpecVersions(t *testing.T) {
+	for _, v := range []struct{ variant, version string }{
+		{"fcos", "1.0.0"}, {"fcos", "1.4.0"}, {"fcos", "1.5.0"}, {"flatcar", "1.1.0"},
```
apedriza, line 519: `openshift` `r4e` ?

internal/provisioner/ignition.go
apedriza, line 139: we can remove this opt then, right?

apedriza (approved):

PR #1612, opened 2026-09-10
Fix the file mode and the scp quoting in the job provisioner entrypoint

s3rj1k:
Two defects in the shell script the provisioning job runs, both on the same file loop.

- Default the job provisioner file mode instead of running an empty chmod
- Quote the scp arguments in the job entrypoint so a file path cannot split them


makhov:
I'm not quite following, why changes are in job_provisioner but ssh_provisioner tests?

s3rj1k:
> I'm not quite following, why changes are in job_provisioner but ssh_provisioner tests?

https://github.com/k0sproject/k0smotron/pull/1612/changes/5e701e350f3be94cb5ebcf632abcf147cf8f4409

Was just adding where other tests where before, now moved to all to dedicated file

PR #1617, opened 2026-09-11
Surface the paused condition on both control plane flavours

s3rj1k:
The condition was declared in both API versions and never set, so a paused control plane looked identical to a running one. Both controllers now call the upstream helper, which patches the condition and reports whether to stop, and the first pass of a reconcile records it and comes back.


apedriza (approved):
lgtm

PR #1618, opened 2026-09-11
Page through the node list when matching a machine to a node

s3rj1k:
The list asked for every node in the workload cluster in one request, which a large cluster cannot answer inside the timeout the client carries. The matching moved into a helper that takes the list call and follows the continue token, and it now returns on the first match rather than scanning on.


apedriza (approved):
lgtm

PR #1619, opened 2026-09-11
Fix: requeue k0smotrol cluster controller until worklaod cluster is f…

### LINKED ISSUE #1614: Cluster remains unavailable after transient LoadBalancer DNS lookup failure
### What happened?

When creating a hosted cluster using a Cluster resource with spec.service.type: LoadBalancer, k0smotron performed its initial control-plane health check before the cloud load balancer’s DNS record had propagated.

Specifically in our case, AWS ELB has been provisioned, but DNS has not yet propagated, so the healthcheck cannot resolve it. So the conditions are:

```
Available=False
ControlPlaneFunctional=False
```

We can work around this by setting an arbitrary annotation on the cluster to trigger re-reconciliation. After that reconciliation, the conditions immediately changed to:

```
Available=True
ControlPlaneFunctional=True
ControlPlaneExposed=True
KubeconfigSecretAvailable=True
```

### Involved Areas

**k0smotron-controller-manager**

We would like the controller to attempt re-reconciliation (with some backoff / retry logic), so that we don't have to manually run this step when provisioning clusters via IaC.

### Steps to reproduce

1. Create a k0smotron.io/v1beta2 Cluster using a cloud LoadBalancer Service.
2. Allow the cloud provider to assign a load-balancer hostname before its DNS record is globally resolvable
3. Observe k0smotron checking the external API endpoint during this DNS propagation window. The lookup fails with no such host.
4. Wait for the hostname to become resolvable and the API server to become reachable.
5. Observe that ControlPlaneFunctional=False and Available=False remain stale
6. Change any annotation on the Cluster resource.
7. Observe that reconciliation runs again and the cluster immediately becomes available.

### k0smotron version

2.1.1

### k0s version

NA

_No response_

apedriza:
…unctional and available

## Description

fix #1614 

## Checklist


internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -149,7 +151,7 @@ func (scope *kmcScope) reconcileResource(ctx context.Context, kmc *km.Cluster, o
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
-func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, retErr error) {
```
jnummelin, line 155: I think this has side-effects. We use `err` e.g. within the deferred status updater func. Maybe worth to check all the return places that we use the named return variables properly

internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -213,6 +215,18 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		if err != nil {
 			logger.Error(err, "Unable to update k0smotron Cluster")
 		}
+
+		// a ctrl.Result with zero values indicates that everything is fine and no immediate requeue is needed
+		// but There might be scenarios where the cluster is not fully stable yet so check the conditions and
+		// requeue if necessary.
+		if res.IsZero() && kmc.DeletionTimestamp.IsZero() {
```
jnummelin, line 221: `kmc` could be nil, since we're in deferred func?

internal/controller/k0smotron.io/k0smotroncluster_controller.go
```diff
@@ -213,6 +215,18 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			// Requeue if the control plane does not report as functional or the cluster is not available
+			// TODO: Consider adding a backoff strategy, temporary disconnections, or other transient issues.
```
jnummelin: This reads a bit weird :) I do get what you mean, someone else might not :D

jnummelin (changes requested):

internal/controller/k0smotron.io/k0smotroncluster_controller.go
apedriza: removed as the conditions checked are self-explanatory

internal/controller/k0smotron.io/k0smotroncluster_controller.go
apedriza: in this case no as the defer is registered after kmc gets populated and is not overwritten to nil at any point

internal/controller/k0smotron.io/k0smotroncluster_controller.go
apedriza: I've aggregated the patch error to the named return error, as we do in other reconcilers. Let me know if that aligns with what you have in mind 👍

internal/controller/controlplane/k0s_controlplane_controller_env_test.go
```diff
@@ -128,7 +128,9 @@ func requirePausedCondition(t *testing.T, kcp *cpv1beta2.K0sControlPlane, want m
 	t.Helper()
 
 	seen := &cpv1beta2.K0sControlPlane{}
-	require.NoError(t, testEnv.Get(ctx, util.ObjectKey(kcp), seen))
+	require.Eventually(t, func() bool {
+		return testEnv.Get(ctx, util.ObjectKey(kcp), seen) == nil
+	}, 10*time.Second, 100*time.Millisecond)
```
apedriza, line 133: Quick fix for a recently failing test. This needs to be within a waitFor as the cache might not be populated

internal/controller/k0smotron.io/k0smotroncluster_controller.go
jnummelin: right, true. I forgot that we initialize kmc with empty-value `Cluster{}` _before_ we fetch it, so even if the fetch fails, it should not have nils

internal/controller/k0smotron.io/k0smotroncluster_controller.go
jnummelin: I guess I was thinking more in terms of "if we use named return values, let's use them consistently throughout the whole func"

apedriza: I left the rest of the returns untouched, since the named vars were mainly added so the defer can respect an already scheduled requeue instead of overriding it with the 20s check, and Go already assigns them correctly anyway. I think it's safe enough to have both options, plus we skip the `res = ...; retErr = ...; return` boilerplate on every branch, wdyt?

jnummelin: Valid resoning, since this is a "small" fix it maybe doesn't warrant complete refactoring of the return values

jnummelin (approved):

PR #1620, opened 2026-09-11
Report scaling from machine totals and add the RollingOut condition

s3rj1k:
Two defects in how control plane progress is reported:

- Report scaling from the machine total rather than the up to date count
- Add the RollingOut condition so rollout progress is still reported


internal/controller/controlplane/status_test.go
apedriza, line 1: TBH I don't think we need such complexity for `setScalingConditions` and `setRollingOutCondition` tests. Personally, I like tests that can help me to understand funcs but introducing auxiliar test funcs more complex than the code to test itself feels weird. Can we just use the classic test table for both, where a given scope sets an expected condition?

s3rj1k: should be handled by ce46a8a3

Issue #1569, opened 2026-08-27
Investigate if we need CRDs in repo


makhov:
We have a lot of CRDs in the repo which are fully auto-generated. It creates noise on PRs. 
Maybe we can just remove them

apedriza:
@makhov what CRDs you referring to? The only ones I can understand that add noise to PRs are the ones related to the controllers but they are needed to build development version from source

makhov:
Yep, that ones :)
The question is do we have to keep them under version control or just always generate them as a make target's dependency

PR #1621, opened 2026-09-11
Generate config/crd as the CRDs are actually installed

s3rj1k:
The directory was a byproduct of a controller-gen call that asks for CRDs while naming only an rbac and a webhook output. It now comes from the kustomize build, so it carries the conversion webhooks and the CA injection that a raw controller-gen CRD lacks, and that call no longer writes there.


makhov:
We also have #1569 to understand should we keep them in repo at all. Maybe you can use a chance to take a look on it as well?

s3rj1k:
> We also have #1569 to understand should we keep them in repo at all. Maybe you can use a chance to take a look on it as well?

sure, I was thinking that someone external could still reference crds in expected location, so if the intention is to drop them all, we probably need to somehow announce this. If we do however want to just simplify maintenance, like make this autogenerated, we should be fine without announcing changes globally.

jnummelin:
> If we do however want to just simplify maintenance, like make this autogenerated, we should be fine without announcing changes globally.

The main intent was to reduce unnecessary churn on the PRs if we do not need those in git. We do need to publish, of course, all the CRDs and resources in releases for use cases like `clusterctl init ...` but I do not see why we'd need them in the development branches

s3rj1k:
> but I do not see why we'd need them in the development branches

drop whole `config/crd` folder?

jnummelin:
> drop whole config/crd folder?

Essentially yes, those are always generated anyways so why do we need them in the git?

The whole point is PRs like this one
image

Very painful to review 😅 

@apedriza @makhov Can you think of any reasons why we'd actually need the generated _stuff_ in the git directly?

s3rj1k:
> The whole point is PRs like this one

you don't need to review CRDs, only Makefile change

apedriza:
> you don't need to review CRDs, only Makefile change

I think the point is exactly that, not to include in the vcs things that are not going to be reviewed but adding noise. 

I will take care of docs extraction and after that let's see if it adds a new headache when releasing or hopefully gets things simpler. One of the things I would like to mantain in the dev experience. I've already memorized the make targets for dev, it would be good to keep them 🙏.

s3rj1k:
> > you don't need to review CRDs, only Makefile change
> 
> I think the point is exactly that, not to include in the vcs things that are not going to be reviewed but adding noise.
> 
> I will take care of docs extraction and after that let's see if it adds a new headache when releasing or hopefully gets things simpler. One of the things I would like to mantain in the dev experience. I've already memorized the make targets for dev, it would be good to keep them 🙏.

Need to be very careful moving installable by kubectl CRDs out of repository, currently those CRDs are already out-of-sync from code, same thing can/will happen with something that is outside of repo.

PR #1623, opened 2026-09-11
Stop regenerating a certificate authority once the control plane is up

s3rj1k:
A deleted CA secret was quietly replaced with a new one, which invalidates every node certificate and the existing kubeconfig. Generation is now gated on the initialization latch, and afterwards a missing certificate is reported by name instead, as upstream does from cluster-api v1.14.


makhov:
I think it should be coordinated with #1562. I'll try to get back to it later today

PR #1624, opened 2026-09-11
Carry the machine deletion timeouts through to the Machine

s3rj1k:
The three node timeouts were declared on the control plane template and read by nothing, so setting them was accepted and ignored. The machine template now carries them where the v1beta2 contract expects, as integer seconds, and the generated Machine gets them.


api/controlplane/v1beta2/k0s_types.go
```diff
@@ -140,6 +140,39 @@ type K0sControlPlaneMachineTemplate struct {
...
+	// spec defines the spec for Machines in a K0sControlPlane object.
+	// +optional
+	Spec K0sControlPlaneMachineTemplateSpec `json:"spec,omitempty,omitzero"`
+}
+
+// K0sControlPlaneMachineTemplateSpec defines the spec for Machines in a K0sControlPlane object.
+type K0sControlPlaneMachineTemplateSpec struct {
+	// deletion contains configuration options for Machine deletion.
+	// +optional
+	Deletion K0sControlPlaneMachineTemplateDeletionSpec `json:"deletion,omitempty,omitzero"`
```
apedriza, line 154: probably we shoudl have the same struct than https://github.com/kubernetes-sigs/cluster-api/blob/main/api/controlplane/kubeadm/v1beta2/kubeadm_control_plane_types.go#L498-L538

s3rj1k: synced what is possible without breaking API

api/controlplane/v1beta2/k0s_types.go
```diff
@@ -140,6 +140,54 @@ type K0sControlPlaneMachineTemplate struct {
 	// InfrastructureRef is a required reference to a custom resource
 	// offered by an infrastructure provider.
 	InfrastructureRef corev1.ObjectReference `json:"infrastructureRef"`
```
apedriza: shouldn't `InfrastructureRef` be inside `K0sControlPlaneMachineTemplateSpec`?

s3rj1k: https://github.com/k0sproject/k0smotron/pull/1624/changes/cc4464ef1ad23af11f47bd2c3c397aee4718526c

should be handled by this commit

apedriza (commented):
ops, I didn't realize this changes the existing spec.machineTemplate.infrastructureRef. The nested shape is the right one (it's what the CAPI v1beta2 contract expects), but we can't change a released API without giving users room to transition. WDYT about keeping both? Add `spec.machineTemplate.spec.infrastructureRef` (plus readinessGates and deletion) as proposed here, and keep the current flat `infrastructureRef` as a deprecated field with some webhook warnings when it is used

s3rj1k:
> ops, I didn't realize this changes the existing spec.machineTemplate.infrastructureRef. The nested shape is the right one (it's what the CAPI v1beta2 contract expects), but we can't change a released API without giving users room to transition. WDYT about keeping both? Add `spec.machineTemplate.spec.infrastructureRef` (plus readinessGates and deletion) as proposed here, and keep the current flat `infrastructureRef` as a deprecated field with some webhook warnings when it is used

https://github.com/k0sproject/k0smotron/pull/1624/changes/fbc709442514913b089a7c5d887fb1a8bee9ac44

something like this?

PR #1625, opened 2026-09-12
Add prm immutability if reserved and logic to restore prm reservation

GOodCoffeeLover:
## Description

1. fix status restore for PooledRemoteMachine in cases of mass status deletion like clusterctl move
2. add immutability of prm when reserved to avoid live patch of remote machine

## Checklist


internal/controller/infrastructure/remote_machine_controller.go
apedriza, line 1: I want to understand these lines. Can you describe why this changes when reserving the pool? Maybe worth moving this reservation logic into a well-tested function

GOodCoffeeLover: Yeah, sorry, this logic became a little bit too complicated.

i try to improve these logic to cover cases, when we search for pooled machine.

Previous logic was simple:
- go through all prm
- look only for machines in target pool
- if we find reserved and reserved by us - we find reserved
- if machine is unreserved - reserved it

but it  does not cover cases, when:
- prm looses its status
- prm's pool changes.

so i try to:
- restore prm status/owner: if prm is unresereved, but rm has address of this prm, assuming that prm just has lost its status, and regain ownership on this prm
- check that prm is owned by rm, even if pools differs.

GOodCoffeeLover: and it's covered by tests in adjacent file.

apedriza: > prm's pool changes.

we should reject this at admission when pool is reserved. IMO it is a valid agreement that is someone is consuming values from a PRM, modification of it during that reservation are blocked (see [this comment](https://github.com/k0sproject/k0smotron/issues/1596#issuecomment-5710531062)).

> prm looses its status

In this case we could use it to populate PRM status only if both specs matches. So we prevent spec mismatches that could modify RM values

GOodCoffeeLover: add cel and now check full spec match.

internal/controller/infrastructure/remote_machine_controller.go
```diff
@@ -363,28 +364,45 @@ func (r *RemoteMachineController) reservePooledMachineAndPopulateRemoteMachine(c
...
+			continue
 		}
+
+		// if remote machine spec matches unresureved pooled machine
+		// then assume remote manchine already has reserved pooled machine
+		// and need to reservec it again
+		if remoteMachineSpecMatchToPooled(rm, &pm) {
+			freePooledMachine = &pm
+			break
+		}
```
apedriza, line 386:
```suggestion
		if remoteMachineSpecMatchToPooled(rm, &pm) {
		    // RemoteMachine already reserved and populated
			return nil
		}
```

Nothing else to do, right?

internal/controller/infrastructure/remote_machine_controller.go
```diff
@@ -363,28 +364,45 @@ func (r *RemoteMachineController) reservePooledMachineAndPopulateRemoteMachine(c
...
 	for _, pm := range pooledMachineList.Items {
-		if pm.Spec.Pool == rm.Spec.Pool {
-			if pm.Status.Reserved && pm.Status.MachineRef.Name == rm.GetName() {
-				foundPooledMachine = &pm
-				break
-			}
+		if pm.Status.Reserved && pm.Status.MachineRef.Name == rm.GetName() {
+			foundPooledMachine = &pm
+			break
+		}
```
apedriza, line 374:
```suggestion
        if pm.Spec.Pool == rm.Spec.Pool {
              if pm.Status.Reserved && pm.Status.MachineRef.Name == rm.GetName() {
                  // RemoteMachine already reserved and populated
    			 return nil
    		  }
        }
		
```
We can keep `pm.Spec.Pool == rm.Spec.Pool` to assume RemoteMachine is provisioned. That way, we ensure the pool is not reserved for another machine with the same name (weird but possible)

Also adding an early return at this point, as nothing else is required

api/infrastructure/v1beta1/remote_machine_types.go
apedriza, line 1: We have received similar issues when updating a reserved PooledRemoteMachine, so we are including a validating webhook to avoid updates when it is reserved, as discussed (see https://github.com/k0sproject/k0smotron/pull/1597) . I think we can keep this PR more atomic by only addressing the deleted status case and removing the static validation in the annotations. Sorry for the overlap on this 🙏

PR #1629, opened 2026-09-14
Warn when a custom user data ref is set for the ignition provisioner

s3rj1k:
The field is accepted and then dropped, since an Ignition config is one JSON document and cannot be appended to as text the way cloud-init and PowerShell output can. Both webhooks now say so, as a warning rather than a rejection, because a config carried over from cloud-init is the likely cause.


apedriza:
I would use the webhook for this kind of logic instead of Kubebuilder annotations. On the other hand, I think we should just throw warnings in these cases as it is more probably the user is reusing a cloud-init config wrongly rather than writing custom user data for the ignition provisioner. In case the user faces this, it is probably just a leftover reusing configs.

s3rj1k:
> I would use the webhook for this kind of logic instead of Kubebuilder annotations. On the other hand, I think we should just throw warnings in these cases as it is more probably the user is reusing a cloud-init config wrongly rather than writing custom user data for the ignition provisioner. In case the user faces this, it is probably just a leftover reusing configs.

Personally I don't like webhook as it brings cert-manager dependency, but we can migrate to webhooks, sure.

api/controlplane/v1beta2/k0s_controlplane_webhook_files_test.go
```diff
@@ -58,3 +59,62 @@ func TestValidateK0sControlPlaneChecksFileOwners(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	warnings, err := (&K0sControlPlaneValidator{}).ValidateCreate(context.Background(), kcp)
+
+	require.NoError(t, err, "the control plane is still accepted")
+	require.Contains(t, warnings, "spec.k0sConfigSpec.provisioner.customUserDataRef is ignored by the ignition provisioner, use provisioner.ignition.additionalConfig instead")
```
apedriza:
```suggestion
	require.Contains(t, warnings, "spec.k0sConfigSpec.provisioner.customUserDataRef is ignored by the ignition provisioner, use spec.provisioner.ignition.additionalConfig instead")
```
For consistency it is better to use either absolute o rel paths instead mix them.

api/bootstrap/v1beta2/k0s_types.go
apedriza, line 1: if we omit `CustomUserDataRef` for ignition, should we avoid to reject request for invalid path within it for the ignition case?

api/controlplane/v1beta2/k0s_controlplane_webhook_files_test.go
s3rj1k: https://github.com/k0sproject/k0smotron/pull/1629/changes/cd8c9c3c20b7bfabd611a095541bd59246f71323 ?

api/bootstrap/v1beta2/k0s_types.go
s3rj1k: https://github.com/k0sproject/k0smotron/pull/1629/changes/5e64ebe1e25e7e4366cb4629f6ac2288592b39a7

more validation

PR #1631, opened 2026-09-14
Follow the upstream election order and report etcd member health

s3rj1k:
The scale down election now follows the two phase process upstream uses, an eligible subset first and then one machine out of the failure domain holding the most.

Upstream weighs every control plane component in that subset. k0s runs the API server, the scheduler and the controller manager as processes, so the etcd member is the only one with an object to read. It is also published on each Machine as EtcdMemberHealthy.


internal/controller/controlplane/scale.go
```diff
@@ -247,13 +271,20 @@ func (c *K0sController) scaleUp(ctx context.Context, scope *controlplane) error
...
+
+	if machineToDelete == nil {
+		machineToDelete = oldestInFullestFailureDomain(ctx, scope, scope.notUpToDateMachines)
+		reason = "outdated"
+	}
 	if machineToDelete == nil {
 		// If we need to scale down but there are no machines elegible for deletion, it means that all the machines are up to date but we
 		// still have more machines than desired. In this case, we can delete the oldest machine, even if it's up to date.
-		machineToDelete = scope.upToDateMachines.Oldest()
+		machineToDelete = oldestInFullestFailureDomain(ctx, scope, scope.upToDateMachines)
```
apedriza: I would encapsulate this in a `selectMachineToDelete` where it holds the details and keep `scaleDown` with less detail

s3rj1k: https://github.com/k0sproject/k0smotron/pull/1631#discussion_r4004601458

internal/controller/controlplane/scale.go
```diff
@@ -189,6 +189,30 @@ func calculateMaxSurge(scope *controlplane) int {
...
+	allMachines := collections.FromMachines(append(scope.activeMachines.UnsortedList(), scope.deletedMachines.UnsortedList()...)...)
+
+	if fd := failuredomains.PickMost(ctx, filterControlPlaneFailureDomains(*scope.cluster), allMachines, candidates); fd != "" {
+		if inDomain := candidates.Filter(collections.InFailureDomains(fd)); inDomain.Len() > 0 {
+			return inDomain.Oldest()
+		}
+	}
+
+	return candidates.Oldest()
+}
```
apedriza, line 244: comparing it with CAPI approach I think we could follow similar logic. For example, in this case we give the same priority to outdated machines and desired machines if they have the annotation for deletion, and we pick the oldest one without considering the failure domain. What I would do basically is what CAPI does:

```go
// The selection is a two phase process:
//
// In the first phase it selects a subset of machines eligible for deletion:
// - if there are outdated machines with the delete machine annotation, use them as eligible subset (priority to user requests, part 1)
// - if there are machines (also not outdated) with the delete machine annotation, use them (priority to user requests, part 2)
// - if there are outdated machines with unhealthy control plane components, use them (priority to restore control plane health)
// - if there are outdated machines  consider all the outdated machines as eligible subset (rollout)
// - otherwise consider all the machines
//
// Once the subset of machines eligible for deletion is identified, one machine is picked out of this subset by
// selecting the machine in the failure domain with most machines (including both eligible and not eligible machines).
```

can we follow the same?

s3rj1k: we can, now PR is feature-heavy

s3rj1k: please run copilot on it

internal/controller/controlplane/k0s_controlplane_controller.go
```diff
@@ -818,6 +824,16 @@ func (c *K0sController) retrieveControlPlaneState(ctx context.Context, cluster *
 		infraMachines:                      infraMachines,
 	}
 
+	// A kine cluster has no etcd members to read, so the health stays empty and every
+	// caller treats it the way it treats an unreadable one.
+	scope.etcdManaged, err = etcdManaged(kcp)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving storage.kine.datasource: %w", err)
```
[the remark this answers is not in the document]
s3rj1k: https://github.com/k0sproject/k0smotron/pull/1631/changes/910e16f2af8a9b0e977e4cfc833d8922e5790a32

api/controlplane/v1beta2/conditions_consts.go
```diff
@@ -37,6 +37,25 @@ const (
 	// remediation failed. A local string, since upstream only has a v1beta1 one.
 	K0sControlPlaneMachineRemediationFailedReason = "RemediationFailed"
 
+	// K0sControlPlaneMachineEtcdMemberHealthyCondition reports whether a control plane
+	// machine's etcd member is joined. Absent when the cluster keeps no etcd members.
```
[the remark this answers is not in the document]
s3rj1k: https://github.com/k0sproject/k0smotron/pull/1631/changes/910e16f2af8a9b0e977e4cfc833d8922e5790a32

internal/controller/controlplane/etcdhealth.go
```diff
@@ -0,0 +1,167 @@
...
+		// it stays out of the map rather than being reported either way.
+		if !ok {
+			continue
+		}
+
+		if etcdMemberUnhealthy(member) {
+			health[machine.Name] = metav1.ConditionFalse
+		} else {
+			health[machine.Name] = metav1.ConditionTrue
+		}
```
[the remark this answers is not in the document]
s3rj1k: https://github.com/k0sproject/k0smotron/pull/1631/changes/910e16f2af8a9b0e977e4cfc833d8922e5790a32

PR #1633, opened 2026-09-14
Validate a K0sControllerConfig at admission

s3rj1k:
The type had no admission webhook at all, so a config created directly rather than by the control plane controller reached the bootstrap controller unchecked. It now goes through the same file owner check the control plane already runs. The spec inlines a pointer, so Validate nil checks before reading anything, which a directly created config carrying only a version would otherwise trip. The envtest environment installs the generated configurations, which fail closed, so it has to serve the new webhook as well.


apedriza:
TBH I don't think we need admission validation as this resource is only created internally by the controlplane controller and controlplane config update should be done in the k0scontrolplane resource

s3rj1k:
> TBH I don't think we need admission validation as this resource is only created internally by the controlplane controller and controlplane config update should be done in the k0scontrolplane resource

What about k0smotron standalone mode?

apedriza:
> What about k0smotron standalone mode?

Standalone already includes its validation webhook for k0smotron Cluster resource

PR #1639, opened 2026-09-15
Create the join token once the data it goes into is ready

s3rj1k:
Both bootstrap controllers created the token secret in the workload cluster before the steps that can fail, so a control plane waiting on an address minted a fresh join credential every retry and deleted none of them.


apedriza (approved):
lgtm

PR #1641, opened 2026-09-15
Let an optional version stay unset instead of failing admission

s3rj1k:
`spec.version` is optional and documented as chosen by k0smotron when empty, but both version parses ran unconditionally. A control plane without one could not be created, and the version the controller picks could never be written back, since it snapshots the object before defaulting the field.

The first commit is a smaller thing in the same expression. The skew check compared minor segments alone, so v1.31.1 to v2.0.0 read as a large downgrade and passed. Kubernetes reaching a new major is unlikely, so it is kept separate.


apedriza:
Sorry, I'm not following. How could I switch from one major to another?

s3rj1k:
> Sorry, I'm not following. How could I switch from one major to another?

this is just a correction of code, I don't think we will ever see Kubernets 2.0

PR #1642, opened 2026-09-16
Add watches for machine and cluster for the bootstrap config controllers

apedriza:
## Description

Adding watches for machines and cluster in the bootrap controller can improve controller reaction for event directly related to the resource being reconciled

## Checklist


internal/controller/bootstrap/worker_bootstrap_controller.go
```diff
@@ -698,10 +703,75 @@ func createInstallCmd(scope *Scope) string {
...
+		client.InNamespace(c.Namespace),
+		client.MatchingLabels{
+			clusterv1.ClusterNameLabel: c.Name,
+		},
+	}
+	result := []ctrl.Request{}
+
+	machineList := &clusterv1.MachineList{}
+	if err := r.Client.List(ctx, machineList, selectors...); err != nil {
+		return nil
```
jnummelin, line 739: If we hit a transient error fetching Machine's, wouldn't returning nil mean we would actually skip reconciliation for this "event" permanently?

jnummelin (commented):
The mapper funcs are not trivial, any chance we could add some unit tests for those?

internal/controller/bootstrap/worker_bootstrap_controller.go
apedriza: Yes, we will skip the reconciliation but it is an event from a Cluster mapped to a K0sWorkerConfig resource, something we don't currently catch at all. So even in the worst case (a transient List error), we're never worse off than the current implementation, we just fall back to not having that extra watch, which is exactly today's behavior.

apedriza:
@jnummelin Added tests as suggested :)

PR #1644, opened 2026-09-16
Pin that a controller config always carries its inlined spec

s3rj1k:
The spec is inlined through a pointer, so an object carrying only a version would leave it nil and panic every promoted read the bootstrap controller performs. Nothing said why that does not happen, which is that the inlined block carries defaults the API server applies.


s3rj1k:
https://github.com/k0sproject/k0smotron/pull/1600

PR #1645, opened 2026-09-16
Report the paused condition from the bootstrap controllers

s3rj1k:
Both bootstrap controllers obeyed the paused annotation without ever saying so, which left a paused config indistinguishable from a stalled one. Switching them to the same EnsurePausedCondition helper the control plane already uses publishes the condition.


apedriza:
can you include a fix for flaky paused test here?

s3rj1k:
> can you include a fix for flaky paused test here?

https://github.com/k0sproject/k0smotron/pull/1645/changes/21670ade2202f6aeedb9a35fec382f1a47d4f2d1

seems to help

apedriza (approved):

PR #1656, opened 2026-09-19
Add test coverage for RandomString

magic-peach:
## Description

`internal/util.RandomString` had zero test coverage. It is used to generate join tokens and other random identifiers, so its length and character set guarantees are worth pinning down explicitly. Adds tests covering the requested length, the character set, and that repeated calls produce distinct output.

## Checklist

## Related issues

None, self directed test coverage improvement.


apedriza:
@magic-peach linter is failing https://github.com/k0sproject/k0smotron/actions/runs/35420305120/job/106241510681?pr=1656

PR #1657, opened 2026-09-21
Merge conditions by type instead of replacing the whole list

s3rj1k:
The v1beta2 condition arrays carry no listType marker, so server side apply treats them as atomic and a second field manager drops every condition it does not set. Upstream marks all twenty four of its own condition fields this way.


internal/controller/controlplane/conditions_listtype_env_test.go
```diff
@@ -0,0 +1,82 @@
...
+	cpv1beta2 "github.com/k0sproject/k0smotron/v2/api/controlplane/v1beta2"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// The merge behaviour lives in the CRD schema rather than in Go, so only a real API
+// server shows it. A fake client accepts either shape.
+func TestControlPlaneConditionsMergeByType(t *testing.T) {
```
apedriza: Seems out of k0smotron responsability. I would drop this test.

s3rj1k: dropped

apedriza (approved):

PR #1663, opened 2026-09-22
Pin CAPI provider versions in documentation

### LINKED ISSUE #1125: Pin CAPI provider versions in examples in docs
Currently we don't pin the versions which means we're picking up the latest. This means also often changes in CRD versions which will result in non-working examples since the APIs can change. See e.g. #1123

david6983:
## Description

Pin CAPI provider versions so new releases cannot silently introduce incompatible CRDs into the documentation examples.

The vSphere pin, v1.15.3, follows the CAPI v1.12 baseline used by our smoke tests. Its CRDs serve `VSphereCluster` and `VSphereMachineTemplate` at `v1beta1`, so the examples switch from `v1beta2` to match. OpenStack and Hetzner examples are likewise aligned with their pinned schemas. These pins provide a reproducible baseline; they are not necessarily the latest releases or a fully tested cross-provider compatibility matrix.

Before each release, maintainers should review the centralized pins in `mkdocs.yml`. When changing a pin, check CAPI compatibility, validate the affected examples, and build the docs before tagging. Existing pins can remain when appropriate. The publishing workflow supplies the k0smotron release version automatically; external provider pins are updated manually.

## Checklist

- [ ] I have added or updated tests covering the change (if applicable) — documentation-only change; no automated tests added
- [x] I have updated relevant documentation (if applicable)
- [x] I have tested the changes locally / on a test cluster — local documentation build and schema checks; cloud provisioning was not tested

## Related issues

Fixes #1125.


jnummelin:
> Before each release, maintainers should review the centralized pins in mkdocs.yml. When changing a pin, check CAPI compatibility, validate the affected examples

That's putting lot of responsibility to maintainers. Essentially that means we would need to actually test all the examples. While that'd be super nice, the maintainers don't actually have access to all provider envs. Some of the examples are community contributed and thus we do not have access to all.

Since I created the issue originally, I _think_ (IIRC) I was more thinking in the lines of: "This example has been tested with provider XYZ on version 1.2.3. For other versions your mileage might vary and you might need to adjust some things."

